PR #51, opened 2015-11-12
respond_500_to_panic

smessier:
respond 500 on recover instead of falling through to "unhandled request"
change generated middleware to log response after recover


raphael:
Thanks!

Issue #68, opened 2015-11-21
Error response clarification


bketelsen:
I love the error responses, but "kind" with an integer is confusing.  "error_number" or something similar would be more clear.

raphael:
Would "id" work?

bketelsen:
Yes that would be fine

raphael:
Thank you for the contribution!
(fixed by 154fe5e)

PR #69, opened 2015-11-22
Change Error "Kind" to Error "ID"

bketelsen:
This changes the output of custom errors to replace the "kind" with "id" for clarity.


error.go
```diff
@@ -27,12 +27,12 @@ import (
 )
 
 type (
-	// ErrorKind is an enum listing the possible types of errors.
+	// ErrorId is an enum listing the possible types of errors.
 	ErrorKind int
```
raphael: The above looks like a search and replace was a bit too eager :) Maybe the type should also be renamed to `ErrorId`?

bketelsen:
Fixes #68

Issue #50, opened 2015-11-12
HTTP response is Golang specific


jim-slattery-rs:
Tiny issue with a confusing error message.. When POSTing no data to an action that has required parameters, I get this 400 Bad Request response:
```json
[
    {
        "kind": 3, 
        "msg": "type of payload must be map[string]interface{} but got value <nil>", 
        "title": "invalid attribute type"
    }
]
```
The `msg` was a little confusing since it talks in Golang terms: `map[string]interface{}`. It really just wants me to submit some valid JSON, like at least `{}`.

raphael:
Thanks @on99, closing this.

PR #73, opened 2015-11-25
Update error message generated to be generic, #50

on99:
* `dictionary` instead of `map[string]interface{}`
* `array` instead of `[]interface{}`

References #50


raphael:
Thank you!

on99:
@raphael  The second commit resolve test error, please also merge that one otherwise travis wont pass.

PR #75, opened 2015-11-27
Explicitly invoke cellar and cleanup after test

on99:
* test cannot pass on my machine because `pwd` is not in `$PATH`
* after test, temp dir should be deleted


raphael:
LGTM, thanks!

PR #76, opened 2015-11-28
Initial implementation of client package and tool generation

raphael:


client.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Refresh makes a OAuth2 refresh access token request.
+func (s *OAuth2Signer) Refresh() error {
+	url := fmt.Sprintf(s.RefreshURLFormat, s.RefreshToken)
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := http.DefaultClient.Do(req)
```
on99: Not close may cause leak here

raphael: Oops, thank you!

PR #78, opened 2015-11-29
Fix golint issue and remove unnecessary test/files

on99:


raphael:
This is great! Do you think we could leave the generated code committed though? While it's true that it's not required it makes it easy to see what has changed since the last time it was generated. It is also a nice way to do a quick sanity check before pushing changes to the code generation algorithms. Thank you!

on99:
Yes you are right! Can I also update the generator so the permission of files it generates is 0644? It's 0755 now but we do not need source files to be excutable.

on99:
Rebased and pushed. Thank you @raphael  for review!

raphael:
LGTM, thanks!

PR #79, opened 2015-11-29
Reduce ResponseTemplate cyclomatic complexity

on99:


design/dsl/api.go
```diff
@@ -269,149 +270,64 @@ func ResponseTemplate(name string, p interface{}) {
...
+			return r
+		}
+		a.ResponseTemplates[name] = &design.ResponseTemplateDefinition{
+			Name:     name,
+			Template: t,
+		}
+	} else {
+		typ := reflect.TypeOf(p)
+		if kind := typ.Kind(); kind != reflect.Func {
+			ReportError("dsl should be a function but get %v", kind)
```
raphael: Please use:
`"dsl must be a function but got %v"`

design/dsl/api.go
```diff
@@ -269,149 +270,64 @@ func ResponseTemplate(name string, p interface{}) {
...
+			return
+		}
+
+		if num := typ.NumOut(); num > 0 {
+			ReportError("dsl should have 0 return parameter but get %d", num)
```
raphael: Please use:
`"dsl functions cannot have return parameters"` (don't think we need to specify how many were actually set)

design/dsl/api.go
```diff
@@ -269,149 +270,64 @@ func ResponseTemplate(name string, p interface{}) {
...
+			return
+		}
+
+		num := typ.NumIn()
+		val := reflect.ValueOf(p)
+		t := func(params ...string) *design.ResponseDefinition {
+			if len(params) < num {
+				ReportError("expected at least %d argument when invoking response template %s", num, name)
```
raphael: Can we have a `s

design/dsl/api.go
```diff
@@ -269,149 +270,64 @@ func ResponseTemplate(name string, p interface{}) {
...
-			t := func(params ...string) *design.ResponseDefinition {
-				if len(params) < 9 {
-					ReportError("expected nine arguments when invoking response template %s", name)
+			r := &design.ResponseDefinition{Name: name}
+
+			in := make([]reflect.Value, num)
+			for i := 0; i < num; i++ {
+				// type checking
+				if t := typ.In(i); t.Kind() != reflect.String {
+					ReportError("expect parameter type string but get %s", t)
```
raphael: Please use:
`("ResponseTemplate parameters must be strings but type of parameter at position %d is %s", i, t)`

design/dsl/api.go
on99: Yes we dont need to specify it but I just want to simulate your code.
Let me remove it if it makes more sense.

design/dsl/api.go
on99: Will do.

design/dsl/api.go
on99: Will do.

design/dsl/api.go
on99: Will do.

on99:
@raphael  updated, please check.

design/dsl/api.go
raphael: I think it makes more sense to remove it, thanks for checking!

raphael:
Thanks! please go ahead with comment 2 and I'll merge.

gogames:
@raphael  I remove the code because I think it makes more sense that we do not check the return parameters.

on99:
@raphael I remove the code because I think it makes more sense that we do not check the return parameters.

raphael:
Makes sense. LGTM.

PR #90, opened 2015-12-03
Add Metadata

bketelsen:
Added Metadata to many of the base DSL types to allow for key/value pairs to be attached to most of the nodes on the DSL tree.

UseCase: In custom generators I need the ability to annotate things that aren't strictly DSL related.

### DIFF
--- a/design/definitions.go
+++ b/design/definitions.go
@@ -69,7 +69,8 @@ type (
 		MediaTypes map[string]*MediaTypeDefinition
 		// dsl contains the DSL used to create this definition if any.
 		DSL func()
-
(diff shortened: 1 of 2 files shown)


bketelsen:
I think the lint checks are failing because goimports has a problem with goregen.  goimports removes the import for goregen for some reason.

bketelsen:
woot passing tests.  Must be a goimports bug.  Strange.

on99:
Import name `goregen` does not match the package name `regen`.
It causes `gocode`, `godef`, `goimports` error.
So make sure they are the same.

raphael:
Thank you, this is great!

PR #99, opened 2015-12-06
Handle signal gracefully

on99:
I did not make changes to cleanup logic so directory `app` `swagger` `client` are not deleted as well.

But it's reasonable as we may mistakenly delete user's code in the same directory.


raphael:
Another great PR, thank you!

raphael:
I had missed your comment somehow... I don't think we should worry about user code in `app`, `swagger` or `client` - these folders get recreated with each generation (see https://github.com/raphael/goa/blob/master/goagen/gen_app/generator.go#L52-L55).

on99:
Let me remove those directories as well.

PR #106, opened 2015-12-08
Fix make failed on fresh environment

### LINKED ISSUE #98: Missing dependency packages
There are several dependencies missing in makefile so make cannot pass, fix it by 

Add missing packages to makefile

or

`vendor` `godep`

on99:
fixes #98


raphael:
Thanks!

PR #113, opened 2015-12-09
Enable adding Swagger tags by using Metadata on API, Resource, and Action

DavyC:


goagen/gen_swagger/swagger.go
```diff
@@ -347,6 +354,24 @@ func New(api *design.APIDefinition) (*Swagger, error) {
 	return s, nil
 }
 
+func tagsFromDefinition(mdatas []design.MetadataDefinition) (tags []Tag, tagNames []string, err error) {
```
raphael: Maybe create two functions instead of one that can return both tags and names? i.e. a `tagNamesFromDefinitions` and a `tagsFromDefinition`. The former takes a slice of `design.MetadataDefinition` (like you have it now) but the latter just a single `*design.MetadataDefinition`.

goagen/gen_swagger/swagger.go
```diff
@@ -347,6 +354,24 @@ func New(api *design.APIDefinition) (*Swagger, error) {
...
+	for _, mdata := range mdatas {
+		if mtags, found := mdata["tags"]; found {
+			tagObjs := []Tag{}
```
raphael: The more "idiomatic" way of doing this is `var tagObjs []Tag`. `append` takes care of creating the slice if it's `nil`.

raphael:
This is great, thanks! Suggesting a small refactor to clean things up a little bit in the implementation but otherwise good stuff!

DavyC:
+1 make sense

raphael:
Thanks! one more thing (sorry should have caught it earlier). Would be good to use slices of pointers to structs instead of slices to structs. So `[]design.MetadataDefinition` should be `[]*design.MetadataDefinition` and `[]Tag` should be `[]*Tag`.

DavyC:
Oh yeah. I had `*Tag`, not sure why/when I changed it back. `design.MetadataDefinition` is a map, should be alright (?)

raphael:
Ah yes you're correct about `design.MetadataDefinition`. Thanks for the great contribution!

PR #133, opened 2015-12-21
update query values after setting them

BrianHicks:
Current behavior of generated code is to set query string values but never write them back to the URL. This should fix that (but do note that if I've changed the wrong file here, yell at me. I haven't tested it very well, just by sed on my current code.)


BrianHicks:
Ah, shoot... this may not work quite right. I'm not sure. Sorry if it doesn't. :\

raphael:
Thanks! I pushed a fix with PR #136

Issue #134, opened 2015-12-22
CORS for Swagger UI is broken


nennete:
Hi,

I am struggle on this. Let's see if anybody can help me.

Since I have pulled the last version, my auto-generated `swagger/swagger.go` looks like:

```
func MountController(service goa.Service) {
	service.ServeFiles("/swagger.json", "swagger/swagger.json")
}
```

This makes sense since `swagger.json` is a static file.

However, before it looked like:

```
// MountController mounts the swagger spec controller under "/swagger.json".
func MountController(service goa.Service) {
	ctrl := service.NewController("Swagger")
	service.Info("mount", "ctrl", "Swagger", "action", "Show", "route", "GET /swagger.json")
	h := ctrl.NewHTTPRouterHandle("Show", getSwagger)
	service.HTTPHandler().(*httprouter.Router).Handle("GET", "/swagger.json", h)
}

// getSwagger is the httprouter handle that returns the Swagger spec.
// func getSwagger(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
func getSwagger(ctx *goa.Context) error {
	ctx.Header().Set("Content-Type", "application/swagger+json")
	ctx.Header().Set("Cache-Control", "public, max-age=3600")
	return ctx.Respond(200, []byte(spec))
}

// Generated spec
const spec = `{...} `
```

If I understood, this old version sets a handler for the given path to serve the swagger spec.

Now, once we set up CORS using the new `goa-middleware/cors` the library parses handlers adding required headers.

However, now the controller does not tie path to handler, therefore, the request serves the file but does not change any header to allow CORS.

Any tip to solve this?

Thanks!

raphael:
Hmm that's an oversight really. The `ServeFiles` implementation uses the router directly therefore bypassing setting up middlewares. It didn't seem necessary that middleware be used to serve files directly. The problem above clearly demonstrates that that's not the case. I'll update the implementation so it properly sets up the middleware. I'm also thinking it may make sense for the `Controller` interface to expose `ServeFiles` so that specific middleware can be setup, would that help your use case?

nennete:
Yeah, that was our guess. Very appreciated if you can apply the require change so we can use the entire solution: goa + goa-middleware/cors.

Thanks a lot!

raphael:
The PR above should fix it, please close the issue if it does, thanks!

nennete:
This was fast! All good now.

Again, thanks a lot a happy holidays!!!

PR #135, opened 2015-12-22
Create a proper controller to serve files.

raphael:
So that the service middleware chain is used when handling serve file requests.


raphael:
Fix #134

PR #150, opened 2015-12-31
BREAKING: decode request.Body only with the right Content-Type header

abourget:
BREAKING CHANGE: to open the door for supporting multiple Content-Type in
incoming requests, we now only decode the req.Body as JSON when the client
specifies "application/json" in the "Content-Type" header, or when that
header is empty.

Eventually, we will implement other decoders to build up the payload.

This change allows someone to receive an incoming "application/octet-stream"
by simply reading the "ctx.Request().Body" .. or handle other types
in the controller.

[Also includes a minor doc addition.]


interlock:
:+1:

PR #151, opened 2016-01-04
BREAKING CHANGE: Versioning support

raphael:
This PR introduces versioning support as described in the [wiki](https://github.com/raphael/goa/wiki/Versioning-APIs). It's a big step towards 1.0 and there are a number a breaking changes that come with it:

* The `Service` interface `HTTPHandler()` method is replaced with `ServeMux()` which returns a `ServeMux` interface (that also implements `http.Handler`). The new goa mux support is described in the [website](http://goa.design/runtime.html). It provides the ability to specify version specific muxes.
* The `Controller` interface `NewHTTPRouterHandle` method is replaced with `HandleFunc` which returns a `HandleFunc` function which can be used to register an endpoint with a mux.
* The `Application` object does not expose the underlying `Router` anymore. Use the `ServeMux` function instead.
* The `Context` object `Get` method returns a simple `string` instead of `(string, bool)`. The string is empty if there is no URL or querystring parameter with the given name. This mirrors the behavior of `url.Values`.


raphael:
This PR introduces versioning support as described in the [wiki](https://github.com/raphael/goa/wiki/Versioning-APIs). It's a big step towards 1.0 and there are a number a breaking changes that come with it:

* The `Service` interface `HTTPHandler()` method is replaced with `ServeMux()` which returns a `ServeMux` interface (that also implements `http.Handler`). The new goa mux support is described in the [website](http://goa.design/runtime.html). It provides the ability to specify version specific muxes.
* The `Controller` interface `NewHTTPRouterHandle` method is replaced with `HandleFunc` which returns a `HandleFunc` function which can be used to register an endpoint with a mux.
* The `Application` object does not expose the underlying `Router` anymore. Use the `ServeMux` function instead.
* The `Context` object `Get` method returns a simple `string` instead of `(string, bool)`. The string is empty if there is no URL or querystring parameter with the given name. This mirrors the behavior of `url.Values`.

PR #152, opened 2016-01-04
Another pass at windows support

moorereason:
All tests pass now.  One oddity: I had to increase the timeout on the
examples test to 200ms on my windows box.

I've noticed that startup times on Windows are often slower than on Linux.  What's funny is that I can run a Linux VirtualBox VM on this same Windows machine, and the 100ms test passes every time.


raphael:
Thank you! this is great.

PR #153, opened 2016-01-05
correct version package name

bketelsen:
use the VersionPackage() function to get the package name for templates.


raphael:
Thanks!

PR #163, opened 2016-01-07
Make cancelation configurable for Graceful application

arangamani:
Graceful application currently cancels existing requests by canceling
the context. If this context is used in http requests, the requests get
canceled and thus the graceful application is not very graceful. This
change makes the cancelation configurable.


raphael:
Thank you!

PR #166, opened 2016-01-08
Decode payload when content type as parameters

ajoulie:
e.g. "application/json;charset=UTF-8"

### DIFF
--- a/service.go
+++ b/service.go
@@ -3,6 +3,7 @@ package goa
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


raphael:
Thank you! Maybe we should use `mime.ParseMediaType` instead of a regular expression? seems more robust. If this function is successful and the identifier is "application/json" then proceed with decoding the body.

ajoulie:
updated

raphael:
Thank you!

PR #167, opened 2016-01-08
log_response

smessier:
added LogResponse() to middleware

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -133,6 +133,37 @@ func LogRequest() Middleware {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


bketelsen:
awesome - would love to see a RedactorFunc() on the logger so you could redact anything possibly sensitive from the stream before it hits the logs.

raphael:
@bketelsen You can use https://godoc.org/github.com/inconshreveable/log15#FilterHandler to setup the logger to do that today.

raphael:
@smessier Thank you!

PR #169, opened 2016-01-09
Add goa client test

on99:


on99:
Invalid, let me close it.

PR #182, opened 2016-01-15
More encoding/decoding work

derekperkins:
Add XML support


encoding.go
```diff
@@ -86,21 +94,31 @@ var (
...
 	for _, contentType := range JSONContentTypes {
 		decoders[contentType] = dp
 		encoders[contentType] = ep
 	}
 
+	// Add xml support
+	xf := &xmlFactory{}
+	dp = newDecodePool(xf)
+	ep = newEncodePool(xf)
+	for _, contentType := range GobContentTypes {
```
raphael: Is this correct? shouldn't it be adding XMLContentTypes decoders and encoders here and Gob ones down below?

derekperkins: You're right. It's fixed now.

PR #186, opened 2016-01-16
Fix whitespace in dsl godoc comments

moorereason:
Action, Attribute, Type, etc have whitespace issues.  Once I got started, I
couldn't stop.  Fixes patterned after encoding/gob.  Use tabs in all of the
preformatted text blocks, even preceeding trailing line comments.  Tab width is
8.


raphael:
awesome! thank you.

PR #200, opened 2016-01-21
Don't call Validate() if there is no Validate() function on a payload

bketelsen:
If a payload has no required fields or other validation triggers a Validate() function is not generated.  The controllers don't  have this check.  This fix generates a Validate() function even if there are no validations, and simply returns a nil error.

{EDIT} Instead this change doesn't test Validate() if it shouldn't.


raphael:
LGTM

PR #207, opened 2016-01-22
Added note to refer to CollectionOf in the ArrayOf doc.

abourget:
I was looking for it.. and it's quite similar in concept to ArrayOf.

Maybe adding other distinguishing features of both in the doc would help.


raphael:
Awsome, thanks! It might also be good to specify that the crucial difference between the two is that `CollectionOf` returns a media type where `ArrayOf` returns a type. Do you think you could add a note about that?

abourget:
Yes.. will do.. wait before merging.

Le ven. 22 janv. 2016 11:51, Raphaël Simon <notifications@github.com> a
écrit :

raphael:
Thank you!

PR #209, opened 2016-01-23
Versioning overhaul

derekperkins:


service.go
```diff
@@ -276,9 +293,41 @@ func (app *Application) ServeFiles(path, filename string) error {
...
+	}
+
+	return app.newVersion(name)
+}
+
+// newVersion instantiates an version with the given name and default decoders/encoders.
+func (app *Application) newVersion(name string) Version {
+	app.versions[name] = &version{
+		name:                  name,
+		mux:                   NewMux(),
```
derekperkins: Is this how I should be instantiating a new mux?

service.go
```diff
@@ -50,12 +57,19 @@ type (
...
 		// NewController returns a controller for the resource with the given name.
 		// This method is mainly intended for use by generated code.
 		NewController(resName string) Controller
+	}
+
+	// Version is the interface for interacting with individual versions. It is embedded by
+	// application for default use with versionless apps
+	Version interface {
+		// Version returns the version string ID
+		Version() string
```
derekperkins: This is a lot of versions, but GetVersion("1.0").Name() seemed weird. Even in comments, I'm not quite sure how to refer to the stringified version. Is is a `VersionID`? `VersionName`? Just `Version`?

derekperkins:
@raphael I added some line comments on parts I'm not sure about. Another thing I'm not sure about is when to set the embedded version in app. Does `Service` need a new function `SetDefaultVersion`?

raphael:
Thank you!

PR #245, opened 2016-02-02
Add Example DSL

DavyC:
Add Example DSL to allow users define their custom example, so goa can generate swagger specs with examples. If not given, goa will automatically generate a random value for the users.

### DIFF
--- a/design/apidsl/media_type_test.go
+++ b/design/apidsl/media_type_test.go
@@ -299,29 +299,28 @@ var _ = Describe("CollectionOf", func() {
 
 var _ = Describe("Example", func() {
 	Context("defined examples in a media type", func() {
-		var mt *MediaTypeDefinition
-
 		BeforeEach(func() {
 			InitDesign()
+		})
 
-			mt = MediaType("application/vnd.example+json", func() {
+		It("produces a media type with examples", func() {
+			mt := MediaType("application/vnd.example+json", func() {
 				Attribute("test1", String, "test1 desc", func() {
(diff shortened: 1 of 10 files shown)


design/types.go
```diff
@@ -133,6 +133,8 @@ const (
 	UserTypeKind
 	// MediaTypeKind represents a media type.
 	MediaTypeKind
+	// NoneKind represents a nil
```
raphael: Is that really needed? adding a new type has a lot of consequences... all generators will have to be updated.

design/apidsl/media_type_test.go
```diff
@@ -296,3 +296,203 @@ var _ = Describe("CollectionOf", func() {
 		})
 	})
 })
+
```
raphael: Can you add a tests with cyclical dependencies? something like
```go
var mt = MediaType("vnd.application/foo", func() {
    Attributes(func() {
        Attribute("foo", "vnd.application/bar")
    })
})

var mt2 = MediaType("vnd.application/bar", func() {
    Attributes(func() {
        Attribute("bar", mt)
    })
})
```

design/types.go
DavyC: I think I am leaning toward `NoExample()` in the end. The `NoneKind` seems a bit too much for one purpose I agree.

raphael: Great, thanks!

raphael:
This is awsome, thank you!

Issue #252, opened 2016-02-03
Validation error says "..must be match..." instead of "...must match..."


konstantin-dzreev:
Example:
```
[{"id":8,"title":"value does not match validation pattern","msg":"load.Href must be match the regexp \"foobar\" but got value \"banana\""}]
```

jim-slattery-rs:
Fixed by https://github.com/goadesign/goa/pull/253

PR #253, opened 2016-02-03
#252 Validation error should say "...must match..." instead of "...must be match..."

konstantin-dzreev:
subj


raphael:
Thank you!

PR #258, opened 2016-02-05
Add more Travis CI options

moorereason:
- Use the latest releases of go 1.4 and 1.5.
- Enable OSX builds.
- Allow go tip builds to fail.


raphael:
Thanks!

Issue #260, opened 2016-02-05
Error: unknown primitive type


KerwinKoo:
```go
var DataArray = MediaType("application/Array+json", func() {
	TypeName("DataArray")
	Attributes(func() {
		Attribute("items", Any, "item list")
	})

	View("default", func() {
		Attribute("items")
	})
})
```

In new version, it creates error by using type `Any`, error information:

```
$ goagen bootstrap -d patest/design

exit status 2
panic: unknown primitive type

goroutine 1 [running]:
github.com/goadesign/goa/design.Primitive.GenerateExample(0x6, 0xc820208780, 0x0, 0x0)
	$GOPATH/go/project/src/github.com/goadesign/goa/design/types.go:277 +0x144
github.com/goadesign/goa/design.(*Primitive).GenerateExample(0xc8201731a0, 0xc820208780, 0x0, 0x0)
	<autogenerated>:161 +0xa5
github.com/goadesign/goa/design.(*AttributeDefinition).GenerateExample(0xc8200a5760, 0xc820208780, 0x0, 0x0)
	$GOPATH/go/project/src/github.com/goadesign/goa/design/definitions.go:120 +0xa7
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExample(0xc8200a5760, 0xc8200369d8, 0x1, 0x1, 0x0, 0x0, 0x0)
	$GOPATH/go/project/src/github.com/goadesign/goa/design/definitions.go:198 +0x23f
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExample(0xc8200a4dc0, 0xc8200369d8, 0x1, 0x1, 0x0, 0x0, 0xc8201cdaf0)
	$GOPATH/project/src/github.com/goadesign/goa/design/definitions.go:185 +0x927
github.com/goadesign/goa/design.(*UserTypeDefinition).Finalize(0xc82019e300)
	$GOPATH/project/src/github.com/goadesign/goa/design/types.go:513 +0x62
github.com/goadesign/goa/dslengine.finalizeSet(0xc8201795e0, 0xa, 0xa, 0x0, 0x0)
	$GOPATH/go/project/src/github.com/goadesign/goa/dslengine/runner.go:249 +0x102
github.com/goadesign/goa/design.(*APIDefinition).IterateSets(0xc8201946c0, 0xaa4980)
	$GOPATH/go/project/src/github.com/goadesign/goa/design/api.go:499 +0x33c
[...]

Issue #259, opened 2016-02-05
Error: goagen panic runtime error when set INTEGER type Minimum(0)


KerwinKoo:
I updated new commit just now, and run test code with defined an attribute inside this `MediaType` define:

```go
var StatisticData = MediaType("application/testcode+json", func() {
	Attributes(func() {
		Attribute("amount", Integer, "Item amount", func() {
			Minimum(0)
		})
	})
...

```

running `goagen bootstrap -d ...` ,  it creates runtime error:

```
$ goagen bootstrap -d patest/design
exit status 2
panic: runtime error: integer divide by zero
[signal 0x8 code=0x1 addr=0x4c8bae pc=0x4c8bae]

goroutine 1 [running]:
github.com/goadesign/goa/design.(*exampleGenerator).generateValidatedMinValueExample(0xc8201e36d0, 0x0, 0x0, 0x0)
	/CODEPATH/go/project/src/github.com/goadesign/goa/design/example.go:208 +0x8e
github.com/goadesign/goa/design.(*exampleGenerator).generate(0xc8201e36d0, 0x0, 0x0)
	CODEPATH/go/project/src/github.com/goadesign/goa/design/example.go:54 +0x4c2
github.com/goadesign/goa/design.(*AttributeDefinition).GenerateExample(0xc8201ede40, 0xc82020c4e0, 0x0, 0x0)
	CODEPATH/go/project/src/github.com/goadesign/goa/design/definitions.go:117 +0x4e
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExample(0xc8201ede40, 0xc820130960, 0x1, 0x1, 0x0, 0x0, 0x0)
	CODEPATH/go/project/src/github.com/goadesign/goa/design/definitions.go:198 +0x23f
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExample(0xc82019a2c0, 0xc820130960, 0x1, 0x1, 0x0, 0x0, 0xc8201e3af0)
	CODEPATH/go/project/src/github.com/goadesign/goa/design/definitions.go:185 +0x927
github.com/goadesign/goa/design.(*UserTypeDefinition).Finalize(0xc820145100)
	CODEPATH/go/project/src/github.com/goadesign/goa/design/types.go:513 +0x62
github.com/goadesign/goa/dslengine.finalizeSet(0xc8202025a0, 0x6, 0x6, 0x0, 0x0)
[...]

PR #263, opened 2016-02-06
Address Issue #259 and #260

DavyC:
Address issues: #259 and #260
- fix the div zero panic while generating examples
- handle `AnyKind` example
- add couple more safety checks while generating examples

### DIFF
--- a/design/example.go
+++ b/design/example.go
@@ -32,7 +32,7 @@ func (eg *exampleGenerator) generate() interface{} {
 		return eg.generateValidatedEnumExample()
 	}
 	// loop until a satisified example is generated
-	hasFormat := eg.hasFormatValidation()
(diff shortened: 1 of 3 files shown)


design/example.go
```diff
@@ -179,49 +165,115 @@ func (eg *exampleGenerator) generateFormatExample() interface{} {
...
+			return int(min) + eg.r.Int()%int(max-min)
+		}
+		return min + eg.r.Float64()*(max-min)
+	} else if min == max {
+		if eg.a.Type.Kind() == IntegerKind {
+			return int(min)
+		}
+		return min
+	} else {
+		panic("Validation: Min > Max")
```
raphael: Please do not make the code panic when user input is incorrect, panics are only ok in case of bugs. Instead since this is the finalize phase I would just ignore the condition and not generate an example.

raphael:
Thank you this is great! merging. Would be awesome if you could remove the calls to `panic` later.

PR #266, opened 2016-02-06
Use :api_version as default version param

raphael:
Also use APIVersion as context field to store version to
avoid clash with Version.


raphael:
Fix #264

PR #278, opened 2016-02-09
goagen: pass extra flags to gen plugins

moorereason:
When running a gen command, capture and pass additional, arbitrary command line flags to the
final generator.  For example, given the following command

    goagen --design=... gen --pkg-path=... -- --special=true -o test

The `--special=true -o test` flags will be passed to the final generator.

### DIFF
--- a/goagen/codegen/command.go
+++ b/goagen/codegen/command.go
@@ -27,6 +27,11 @@ var (
 
 	// CommandName is the name of the command being run.
 	CommandName string
+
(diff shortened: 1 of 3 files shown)


raphael:
Looks good! do you think you could add a test?

moorereason:
Yes, but it will probably be this evening before I can work on it (12+ hours)

raphael:
No problem, thank you!

raphael:
Thank you!

PR #282, opened 2016-02-10
checking for nil before asserting

dominicm:


raphael:
This is great! thank you.

goagen/gen_app/generator_test.go
```diff
@@ -493,7 +493,9 @@ func MountWidgetController(service goa.Service, ctrl WidgetController) {
 	mux := service.ServeMux()
 	h = func(c *goa.Context) error {
 		ctx, err := NewGetWidgetContext(c)
-		ctx.Payload = ctx.RawPayload().(Collection)
+		if rawPayload := ctx.RawPayload(); rawPayload != nil {
+			ctx.Payload = rawPayload.(Collection)
+		}
```
moorereason: Shouldn't this come after the `err != nil` if-block?  For example:

```go
ctx, err := NewGetWidgetContext(c)
if err != nil {
	return goa.NewBadRequestError(err)
}
if rawPayload := ctx.RawPayload(); rawPayload != nil {
	ctx.Payload = rawPayload.(Collection)
}
```

goagen/gen_app/writers.go
```diff
@@ -528,7 +528,9 @@ func Mount{{.Resource}}Controller(service goa.Service, ctrl {{.Resource}}Control
 {{$res := .Resource}}{{$ver := .Version}}{{range .Actions}}{{$action := .}}	h = func(c *goa.Context) error {
 		ctx, err := New{{.Context}}(c)
 {{if not $ver.IsDefault}}		ctx.APIVersion = service.Version("{{$ver.Version}}").VersionName()
-{{end}}{{if .Payload}}		ctx.Payload = ctx.RawPayload().({{gotyperef .Payload nil 1}})
+{{end}}{{if .Payload}}		if rawPayload := ctx.RawPayload(); rawPayload != nil {
+				ctx.Payload = rawPayload.({{gotyperef .Payload nil 1}})
+		}
```
moorereason: Same issue here.

goagen/gen_app/generator_test.go
raphael: Good point, thanks!

PR #283, opened 2016-02-10
ACL-214 400 errors were returned as a string containing a malformed JSON object

jim-slattery-rs:
… this was due to an unescaped `err` being added to the msg string making the JSON invalid.

E.g. here's what it was returning:
```json
"{\"kind\":\"invalid request\",\"msg\":\"invalid encoding: [{\"id\":8,\"title\":\"value does not match validation pattern\",\"msg\":\"raw.principal.href must match the regexp \\\"^/grs/(users|groups)/(\\\\\\\\d+)$\\\" but got value \\\"/grs/users/1a\\\"\"}]\"}"
```
Now it will return as a JSON object, and probably will work for XML as well:
```json
{
  "kind": "invalid request",
  "msg": "invalid encoding: [{\"id\":8,\"title\":\"value does not match validation pattern\",\"msg\":\"raw.principal.href must match the regexp \\\"^/grs/(users|groups)/(\\\\\\\\d+)$\\\" but got value \\\"/grs/users/1a\\\"\"}]"
}
```

TODO: it would be extra nice if the error details were actually JSON as well, though.. maybe as a separate "detail" field?:
```json
{
  "kind": "invalid request",
  "msg": "invalid encoding",
  "detail": [
    {
      "id":8,
      "title":"value does not match validation pattern",
      "msg":"raw.principal.href must match the regexp \"^/grs/(users|groups)/(\\\\d+)$\" but got value \"/grs/users/1a\""
    }
  ]
}
```

### DIFF
--- a/service.go
+++ b/service.go
@@ -457,7 +457,10 @@ func (ctrl *ApplicationController) HandleFunc(name string, h, d Handler) HandleF
 		handler := middleware
 		if err != nil {
 			handler = func(ctx *Context) error {
-				ctx.Respond(400, fmt.Sprintf(`{"kind":"invalid request","msg":"invalid encoding: %s"}`, err))
+				ctx.Respond(400, map[string]interface{}{
+					"kind": "invalid request",
+					"msg":  "invalid encoding: " + err.Error(),
+				})
 				return nil
 			}
 			for i := range chain {
--- a/service.go
+++ b/service.go
@@ -457,10 +457,8 @@ func (ctrl *ApplicationController) HandleFunc(name string, h, d Handler) HandleF
 		handler := middleware
 		if err != nil {
 			handler = func(ctx *Context) error {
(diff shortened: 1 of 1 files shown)


raphael:
Thank you, this makes sense. However before we go the route of marshaling data structures to return errors we would need to make `TypedError` and `MultiError` (in [error.go](https://github.com/goadesign/goa/blob/master/error.go) do the same. The idea is to keep responses consistent when an error is returned from a controller or from `HandleFunc`. I would suggest modifying this PR so the output for this is a valid JSON string. Then have a separate issue to change the way all errors get returned.

jim-slattery-rs:
Thanks, makes sense, for now I've just fixed the "msg" string to be properly formed.
First time I've used the `%q` formatter :)

raphael:
Looks great - thank you!

PR #284, opened 2016-02-10
checking err first before accessing ctx

dominicm:


goagen/gen_app/generator_test.go
```diff
@@ -437,10 +437,11 @@ func MountWidgetController(service goa.Service, ctrl WidgetController) {
 	mux := service.{{if .version}}Version("{{.version}}").ServeMux(){{else}}ServeMux(){{end}}
 	h = func(c *goa.Context) error {
 		ctx, err := NewGetWidgetContext(c)
-{{if .version}}		ctx.APIVersion = service.Version("{{.version}}").VersionName()
-{{end}}		if err != nil {
+		if err != nil {
 			return goa.NewBadRequestError(err)
 		}
+{{if .version}}		ctx.APIVersion = service.Version("{{.version}}").VersionName()
+{{end}}
```
jim-slattery-rs: looks like an extra newline got added here.. should include the `return ctrl.Get(ctx)` on this line?

moorereason: We're not very consistent with blank lines preceding an ending `return` statement.  I prefer the extra line, myself.

raphael: It's true that it's not entirely consistent but I would consider extra newlines bug (the intent was that there shouldn't be newlines - it should be the vast majority of cases). Just have to pick one style (I use newlines before return in my code too sometimes so not against it in principle), consistency seems better.

raphael:
Awesome, thank you!

PR #286, opened 2016-02-11
ACL-214 use unmarshaller if provided, so we error if required payload is not provided

jim-slattery-rs:
When making a request that requires a payload, Goa is not returning an error if the payload was simply not provided `""`. It should return an error, rather than invoke the controller action with a nil payload.

### DIFF
--- a/service.go
+++ b/service.go
@@ -449,7 +449,7 @@ func (ctrl *ApplicationController) HandleFunc(name string, h, d Handler) HandleF
 
 		// Load body if any
 		var err error
-		if r.ContentLength > 0 && d != nil {
+		if d != nil {
 			err = d(ctx)
 		}
 


raphael:
I am not sure I agree with this fix. You may have requests that can accept optional payloads. It seems to me that the right thing to do here is to extend the DSL to let the designer specify whether they payload is required then let the generated code check if needed.

jim-slattery-rs:
We did use the DSL to define required fields in the payload, but the unmarshaller is completely bypassed in this case where  `ContentLength == 0`, so the generated code never gets to check if the fields are there.

jim-slattery-rs:
I did run a test locally with an action that defines a payload, but does not mark any attributes as required.
This PR will break those cases, causing requests with empty payloads to no longer be accepted.

I'll close this PR and open a separate issue. For now, I can work around it by adding new guards in all of our controller actions to check that the payload is not nil.

PR #294, opened 2016-02-13
goagen: restore goimports-style formatting

moorereason:
After commit f85b91bb, we lost goimports-style import path sorting.  This commit
restores that behavior and simplifies the codegen.FormatCode func.

### DIFF
--- a/goagen/codegen/imports.go
+++ b/goagen/codegen/imports.go
@@ -23,7 +23,7 @@ import (
 	"golang.org/x/tools/go/ast/astutil"
 )
 
-func FormatFile(fset *token.FileSet, f *ast.File) ([]byte, error) {
+func formatFile(fset *token.FileSet, f *ast.File) ([]byte, error) {
 	sortImports(fset, f)
 	imps := astutil.Imports(fset, f)
 
--- a/goagen/codegen/imports.go
+++ b/goagen/codegen/imports.go
@@ -1,0 +1,322 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// copied from golang.org/x/tools/imports
+
+package codegen
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"go/ast"
+	"go/format"
+	"go/printer"
+	"go/token"
+	"io"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"golang.org/x/tools/go/ast/astutil"
+)
+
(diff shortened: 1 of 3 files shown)


raphael:
Thank you! looks good to me. Just wondering if there was a reason @ernesto-jimenez didn't use this approach in the first place? Also please make sure to print the content of the file if `goimports` fails like the current code is doing since the file gets deleted and there's no way to know what went wrong otherwise. This code:

```go
 	if err != nil {
 -		content, _ := ioutil.ReadFile(f.Abs())
 -		var buf bytes.Buffer
 -		scanner.PrintError(&buf, err)
 -		return fmt.Errorf("%s\n========\nContent:\n%s", buf.String(), content)
 -	}
```

Thank you again!

moorereason:
I've fixed the error message regression.  I can squash this into a single commit if you want.

ernesto-jimenez:
@raphael @moorereason, using `goimports` can cause trouble when you have more than one package with the same names and similar signatures.

The typical example is using `template` where `imports` will import `text/template` instead of `html/template` due to its heuristics.

ernesto-jimenez:
Note that while using `imports` there were several instances where `goa`'s generators were missing imports in the import specs and `goimports` was hiding it out, which can lead to problems later as per my previous comment where imports could import the wrong package.

Also, `goimports` can be quite slow sometimes.

moorereason:
So, what we really want is goimports that doesn't add imports but only removes unused imports and then formats them properly.  Do we copy a bunch of code or try to get the Go team to export some `imports` funcs?

I would really like for goa's generated code to pass golint and friends.

raphael:
That's a lot of additional code that needs to be tested and maintained just to split into stdlib vs non-stdlib imports... At the end of the day I don't know if it's worth it for generated code.... I'd be OK just using `goimports` like you had it initially. The risk as @ernesto-jimenez mentioned is that it may hide missing `import`  in generated code. I'm not too worried about getting the "wrong" import since the import shouldn't be missing in the first place and missing imports are not going to be common. The generated code always requires the same potential set of imports and that is a fairly static set.

So tl;dr we can either go back to your initial PR using goimports (and adding the printing of the source in case of failure which this latest PR didn't address) or we could decide to leave things the way they are now. Either option is OK by me - I just believe the PR in its current state is overkill.

ernesto-jimenez:
@raphael my initial PR was because I had problems with missing imports, and there were several missing imports in the import specs. Some of which where not even picked up by the generator's tests.

I'm too worried with the size of the current code too, I'm unsure what the actual behaviour actually is.

raphael:
yup like I commented above, happy to go either way.

PR #297, opened 2016-02-16
ACL-214 MediaTypesWriter will iterate views in a consistent order

jim-slattery-rs:
This will make the generated code more stable, otherwise types like these would randomly switch positions each time `app` code was generated:
- `MyAppReferenceScope`
- `MyAppReferenceScopeLink`

### DIFF
--- a/design/types.go
+++ b/design/types.go
@@ -552,6 +552,16 @@ func (m *MediaTypeDefinition) ComputeViews() map[string]*ViewDefinition {
 	return nil
 }
 
+// SortedViews returns view names in alphabetical order.
+func (m *MediaTypeDefinition) SortedViews() []string {
+	views := make([]string, 0, len(m.Views))
+	for view := range m.Views {
+		views = append(views, view)
+	}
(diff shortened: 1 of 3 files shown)


raphael:
Good catch, thank you!

Do you think you could update the PR to follow the same convention used in other places of having `IterateXXX` functions that take an iterator and iterate in order? We'd have a `func (mt *MediaTypeDefinition) IterateViews(it ViewIterator) error` function where `ViewIterator` would be `type ViewIterator func(*ViewDefinition) error`. This would live in the `design` package next to the `MediaTypeDefinition` type definition. See https://github.com/goadesign/goa/search?p=1&q=Iterate&utf8=%E2%9C%93 for a bunch of examples like this one: https://github.com/goadesign/goa/search?p=1&q=Iterate&utf8=%E2%9C%93.

jim-slattery-rs:
@raphael: All set, if you want to take another look!

raphael:
Perfect! thank you.

PR #298, opened 2016-02-17
Context v2

raphael:
This PR changes how goa handles the request context. Prior to this change goa introduced its own context data structure (`Context`). This data structure captured the request and response state and also implemented the Logger interface. This made for a ill-defined abstraction that ended up being a "kitchen sink" for various needs.

This PR removes `Context` and instead uses the standard `context.Context` interface. The interface of generated context is mostly unchanged, they wrap a `context.Context` as well as the new `ResponseData` and `RequestData` data structures which provide the same functionality `Context` used to provide (minus logging). `RequestData` encapsulate the request state including the `http.Request` object itself as well as the path and querystring parameters. `ResponseState` encapsulates the response state including the response writer, written byte count and http status code. The goa package exposes functions that can be used to retrieve these data structures from the context - this is mainly useful for middleware and generic code.

With this change goa is now also decoupled from the logger backend. Client applications are free to use whatever logger backend they want and can plug that backend via the package `Log` variable. The default backend uses the stdlib `log` package. The plan is to provide adapters for other backends in the `logging` package (TBD).


derekperkins:
:+1:

PR #300, opened 2016-02-18
Metrics

bketelsen:
This PR adds a global metrics object and instruments several key pieces of functionality in the goa package.  It does *not* add any instrumentation to the generated code.

Calls to the metrics functions will be no-ops when the metrics object isn't initialized.

### DIFF
--- a/context.go
+++ b/context.go
@@ -152,26 +152,24 @@ func (r *ResponseData) Written() bool {
 // encoders. It uses the default service encoder if no match is found.
 func (r *ResponseData) Send(ctx context.Context, code int, body interface{}) error {
 	r.WriteHeader(code)
-	go IncrCounter([]string{"goa", "response", "manual", strconv.Itoa(code)}, 1.0)
 	return RequestService(ctx).EncodeResponse(ctx, body)
 }
 
 // BadRequest sends a HTTP response with status code 400 and the given error as body.
 func (r *ResponseData) BadRequest(ctx context.Context, err *BadRequestError) error {
-	go IncrCounter([]string{"goa", "response", "400"}, 1.0)
 	return r.Send(ctx, 400, err.Error())
 }
 
 // Bug sends a HTTP response with status code 500 and the given body.
 // The body can be set using a format and substituted values a la fmt.Printf.
 func (r *ResponseData) Bug(ctx context.Context, format string, a ...interface{}) error {
-	go IncrCounter([]string{"goa", "response", "bug"}, 1.0)
 	body := fmt.Sprintf(format, a...)
 	return r.Send(ctx, 500, body)
 }
 
 // WriteHeader records the response status code and calls the underlying writer.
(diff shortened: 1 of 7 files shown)


service.go
```diff
@@ -453,6 +454,7 @@ func DefaultErrorHandler(ctx context.Context, rw http.ResponseWriter, req *http.
 	} else {
 		Log.Error(ctx, e.Error())
 	}
+	go IncrCounter([]string{"goa", "handler", "default", "error", strconv.Itoa(status)}, 1.0)
```
raphael: I would move these to were the error handler is invoked instead so that the counter gets incremented even if a user has specified a custom error handler, that is here: https://github.com/goadesign/goa/blob/master/service.go#L387

service.go
```diff
@@ -467,6 +469,7 @@ func TerseErrorHandler(ctx context.Context, rw http.ResponseWriter, req *http.Re
 	} else {
 		Log.Error(ctx, e.Error())
 	}
+	go IncrCounter([]string{"goa", "handler", "terse", "error", strconv.Itoa(status)}, 1.0)
```
raphael: See above

context.go
```diff
@@ -151,17 +152,20 @@ func (r *ResponseData) Written() bool {
 // encoders. It uses the default service encoder if no match is found.
 func (r *ResponseData) Send(ctx context.Context, code int, body interface{}) error {
 	r.WriteHeader(code)
+	go IncrCounter([]string{"goa", "response", "manual", strconv.Itoa(code)}, 1.0)
```
raphael: Would it make more sense to have these in `ReponseData.WriteHeader`? This would capture all responses.

context.go
```diff
@@ -151,17 +152,20 @@ func (r *ResponseData) Written() bool {
...
 	return RequestService(ctx).EncodeResponse(ctx, body)
 }
 
 // BadRequest sends a HTTP response with status code 400 and the given error as body.
 func (r *ResponseData) BadRequest(ctx context.Context, err *BadRequestError) error {
+	go IncrCounter([]string{"goa", "response", "400"}, 1.0)
```
raphael: Same as above

context.go
```diff
@@ -151,17 +152,20 @@ func (r *ResponseData) Written() bool {
...
 	return r.Send(ctx, 400, err.Error())
 }
 
 // Bug sends a HTTP response with status code 500 and the given body.
 // The body can be set using a format and substituted values a la fmt.Printf.
 func (r *ResponseData) Bug(ctx context.Context, format string, a ...interface{}) error {
+	go IncrCounter([]string{"goa", "response", "bug"}, 1.0)
```
raphael: Same as above above

bketelsen: Perfect .. I was wondering where a good place for this would be.

raphael:
Thank you!

Issue #307, opened 2016-02-21
Using `HashOf()` prevents generation of a swagger file.


abourget:
The `GenerateExample` code makes an example a `map[interface{}]interface{}`.. and this doesn't compile through `json.Marshal()`.

Thus, when using `HashOf`, setting an explicit `Example(map[string]interface{}{"something": 123})` in the `Attribute()` definition will bypass the creation of the generated example, and thus at least work.

Ideally, we wouldn't need to set an example, but we'd have a generated example, or have the `map[interface{}]interface{}` automatically translated to a `map[string]interface{}` in the swagger generation. It would, in any case, be invalid Swagger to have an object that doesn't have a string as a key.

raphael:
Thanks for the report! @DavyC do you think you could take a look?

raphael:
@abourget did the fix above help?

PR #310, opened 2016-02-22
Create json parsable map type (builtin key) if possible

DavyC:
Address issue #307, where `json.Marshal` cannot handle `map[interface{}]interface{}`. The change examines the key type from the `Hash` and create a `map` with builtin type if possible.


raphael:
This is great, thank you!

PR #312, opened 2016-02-24
Fix up Goify to understand initialisms

### LINKED ISSUE #295: Add common initialisms to Goify
Running `codegen.Goify("test_id", false)` returns `testId`.  It should return `testID`.

Consider adding the [common initialisms](https://github.com/golang/lint/blob/32a87160691b3c96046c0c678fe57c5bef761456/lint.go#L702) from golint to Goify's logic.

### LINKED ISSUE #306: Use idiomatic Go names for `ID`, `API` and other acronyms in generated `app` code.
Right now, `Attribute("backend_id")` produces `app` code with `BackendId` fields.

There is acronym support already in (https://github.com/goadesign/goa/blob/de9986a748f957c94c26af68218fb1f4a4c3b0b8/goagen/codegen/types.go#L249-L253), but it doesn't seem to be used in this case.

Let's also make sure we have a bunch of possible values in there, so we don't later break users' fields if we add some later.

bketelsen:
This commit fixes Goify to recognize the same initialisms that golint does.  Also adds several tests.

### DIFF
--- a/goagen/codegen/types.go
+++ b/goagen/codegen/types.go
@@ -252,56 +252,98 @@ var acros = map[string]string{
 	"api": "API",
 }
 
+var commonInitialisms = map[string]bool{
+	"API":   true,
+	"ASCII": true,
+	"CPU":   true,
+	"CSS":   true,
+	"DNS":   true,
+	"EOF":   true,
+	"GUID":  true,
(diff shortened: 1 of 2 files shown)


raphael:
Looks awesome!

PR #316, opened 2016-02-24
Fix Goify for edge (non goa) cases

bketelsen:
fixes the fixes for 295, adding more tests for positive and negative cases


raphael:
LGTM

PR #317, opened 2016-02-24
make goify great again

bketelsen:


goagen/codegen/types.go
```diff
@@ -354,6 +357,14 @@ func Goify(str string, firstUpper bool) string {
 	return res
 }
 
+// validIdentifier returns true if the rune is a letter or number
+func validIdentifier(r rune) bool {
```
raphael: This could just be `return unicode.IsLetter(r) || unicode.IsDigit(r)`

bketelsen: well that would make too much sense and save 3 lines of code.

Issue #92, opened 2015-12-03
Add tests for metadata


bketelsen:
Just in case we forget.

raphael:
Write a file `metadata_test.go` in the [apidsl](https://github.com/goadesign/goa/blob/master/design/apidsl) package that tests the [Metadata](https://github.com/goadesign/goa/blob/master/design/apidsl/metadata.go#L28) function. Ideally use [gomega](https://onsi.github.io/gomega/) to write the tests  (see any other test file for examples e.g. [response_test.go](https://github.com/goadesign/goa/blob/master/design/apidsl/response_test.go)).

nii236:
I'm going to have a look at doing this tomorrow =)

raphael:
Thank you!

PR #320, opened 2016-02-25
Feature/metadata tests

nii236:
This pull request refers to [issue 92](https://github.com/goadesign/goa/issues/92), and contains tests for `metadata.go`.

### DIFF
--- a/design/apidsl/metadata_test.go
+++ b/design/apidsl/metadata_test.go
@@ -134,6 +134,10 @@ var _ = Describe("Metadata", func() {
 			Ω(api.Metadata).To(BeNil())
 			Ω(rd.Metadata).To(BeNil())
 			Ω(mtd.Metadata).To(BeNil())
+
+			var mtdAttribute *AttributeDefinition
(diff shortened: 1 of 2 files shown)


design/apidsl/metadata_test.go
```diff
@@ -0,0 +1,144 @@
...
+				Ω(rd.Actions["Example Action"].Metadata).To(Equal(expected))
+				Ω(rd.Responses["Example Response"].Metadata).To(Equal(expected))
+				Ω(mtd.Metadata).To(Equal(expected))
+
+				var mtdAttribute *AttributeDefinition
+				mtdAttribute = mtd.Type.ToObject()["Example Attribute"]
+				Ω(mtdAttribute.Metadata).To(Equal(expected))
+			})
+		})
+		Context("with invalid metadata string", func() {
```
raphael: Is "invalid" the right term here? the metadata gets set so it's not really invalid. Maybe something like "with unicode characters" would be better?

raphael:
This is an awesome contribution, thank you @nii236 ! I left a tiny comment in there, will merge as soon as it's addressed.

PR #323, opened 2016-02-25
goagen: clean up gen_app templates

moorereason:
This commit cleans up two small issues:

- Remove unnecessary int-to-int type conversions for Integer types.
- Return error from ctx.ResponseData.Write call in context response func.


raphael:
LGTM, thank you!

Issue #314, opened 2016-02-24
Issues of the example generator


noblehng:
First, in `design/example.go`:
* I think there is a typo [here](https://github.com/goadesign/goa/blob/07928cc4306802d5efe840d64d3b6fa05a0e3e64/design/example.go#L81), `minlength` should be `maxlength`. 
* And [this](https://github.com/goadesign/goa/blob/07928cc4306802d5efe840d64d3b6fa05a0e3e64/design/example.go#L99) could make a huge slice and cause OOM error.

Second, in `design/definitions.go`:
* The MakeMap() function [here](https://github.com/goadesign/goa/blob/07928cc4306802d5efe840d64d3b6fa05a0e3e64/design/definitions.go#L157) will panic if either the key or the value is nil.
* In the same switch of the above line, `AnyKind` isn't handled. I think one could just set `a.isCustomExample = true` for `AnyKind`.

raphael:
Thank you for the detailed report! @DavyC sorry to bug again but would you have time to look into these? thank you!

DavyC:
Yap, my fault. Thanks for catching those.

DavyC:
Regarding to the OOM issue, perhaps, one solution would be to generate <10 examples in the array, in case e.g. `MinLength(1000)` or higher is given. Unfortunately, the example is no longer valid at this point... but an example with 1000 items in an array doesn't make too much sense either.

noblehng:
I got the OOM error when using a large MaxLength(), something like `MaxInt32`.

raphael:
Limiting to 10 max is probably good enough.

raphael:
Thanks @DavyC !

PR #324, opened 2016-02-26
Fixes Issue #314: handle Any while generating examples

DavyC:
Issue #314 Fixes
 * avoid OOM from creating a large slice in example because of high `MinLength`
 * add `Any` type handling in both `Array` and `Hash` `GenerateExample`
 * generate example with more precise type instead of using `interface{}` everywhere (make tests easier)

### DIFF
--- a/design/apidsl/media_type_test.go
+++ b/design/apidsl/media_type_test.go
@@ -318,13 +318,6 @@ var _ = Describe("Example", func() {
 		})
 
 		It("produces a media type with examples", func() {
-			ut := Type("example", func() {
-				Attribute("test1", Integer, func() {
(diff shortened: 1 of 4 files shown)


raphael:
Thank you for all the hard work! I was looking at it and wondering why the examples were first generated as `interface{}` then cast back to the "original" type? Is there a reason the code can't look at the type when generating so it produces values of the right (Go) type to start with?

DavyC:
You are absolutely correct. I think I went back and forth so many times and kept changing how it should be done. In the end, I confused myself... and made myself believe there is a reason I couldn't use the right Go type from the beginning :laughing:

PR #327, opened 2016-02-26
codegen: honor initialisms in lowercase Goify

### LINKED ISSUE #326: codegen: lowercase Goify should honor initialisms
`codegen.Goify` doesn't honor initialisms when `firstUpper` is `false`.  For example,

```go
_ := codegen.Goify("id", false) // yields "iD" instead of "id"
```

I'm preparing a PR for this.

moorereason:
Fixes #326


raphael:
Awesome! thanks.

PR #328, opened 2016-02-27
Generate proper media type godocs

moorereason:
This commit updates the descriptions for media type views and collections to
proper godoc formatting.

I had to import `codegen` to get access to `Goify`.

### DIFF
--- a/design/types.go
+++ b/design/types.go
@@ -625,18 +625,13 @@ func (m *MediaTypeDefinition) projectSingle(view string) (p *MediaTypeDefinition
 		val = m.Validation.Dup()
 		val.Required = required
 	}
-	description := m.Description
-	if view != "default" {
-		description += fmt.Sprintf(", %s view", view)
-	}
 	p = &MediaTypeDefinition{
 		Identifier: m.Identifier,
 		UserTypeDefinition: &UserTypeDefinition{
 			TypeName: typeName,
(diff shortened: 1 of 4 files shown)


raphael:
The change makes sense but we should not be importing the codegen package in the design. The dependency is the other way around. The design type name does not have to be a valid Go type or follow any Go convention for that matter. The code generation algorithms that generate Go code should be doing that. So maybe move that logic to codegen?

raphael:
Same comment for the updated descriptions, these looks like Go code description not design descriptions. Maybe move that too?

moorereason:
I figured you would say that about importing codegen.  :smile:  I originally tried to solve this issue just in the media type template in `gen_app`, but I didn't have access to the fields I needed.  Then I noticed that the description of the view and collections media types were being set rather crudely in the design package.

I think the better approach would be to **not** copy the Description of the main MediaType into the view and collection media types and then create a template helper function to generate the godocs for types with an empty Description.

I have a track meet to go to today, but I'll work on refactoring this PR.  Thanks for the feedback.

moorereason:
@raphael, I took another stab at this PR.  One regression is that I can't generate a "view" description for a media type because that information is not attached to the MediaTypeDefinition, or at least I couldn't find it.

raphael:
Very nice! thank you!

Issue #305, opened 2016-02-21
Metadata, tweak syntax for extensibility and simplicity


abourget:
We'd like to have something like:

    Metadata("struct:tag:json", "boo")
    Metadata("struct:tag:xml", "mama")

    Metadata("swagger:tag:backend", "Backend")
    Metadata("swagger:tag:backend:doc", "Big documentation")
    Metadata("swagger:tag:backend:url", "http://example.com")

This would introduce the convention of namespaces split by `:`.

This changes only slightly what is already there with the latest Metadata changes. It's more of a refinement than a big change.

I'll take a bit at it when I have a moment.

raphael:
Looks great, thanks!

PR #333, opened 2016-02-29
Implemented metadata with extensible model, and swagger generation.

abourget:
Follow up on https://github.com/goadesign/goa/issues/305


raphael:
Thank you!

PR #334, opened 2016-03-01
Fix the panic on example generation for HashOf

arangamani:
When using `Any` type as either a Key or a Value of a Hash, the
code generation panics with
```
panic: invalid value type
```


raphael:
I think #314 may have fixed that already? at least it's conflicting :(
Would you mind updating the PR now that #314 is merged?
Thanks!

arangamani:
Ah! I should have looked at the existing issues. With goa's development speed, I guessed that if there was a fix for it, it would already be merged to master :wink: I'll check with the latest code.

arangamani:
The latest changes in master fixes my issue. Closing...

Issue #193, opened 2016-01-21
Add version support to gen_swagger


bketelsen:

abourget:
Please :) this is a blocker for me to use versioning.

PR #335, opened 2016-03-01
Remove built-in versioning support.

raphael:
The trade off between code complexity and benefits just wasn't worth it.
Also the approach was forcing generators to implement one way of doing
versioning removing the freedom from end users to implement it the way
they needed.
There is now one design per version. All generators operate in the
context of a single version.


raphael:
Fix #192 #193 #194

PR #336, opened 2016-03-01
Added `FailOnError` function, tohandle MultiError

abourget:
, and as a way to keep a `gen/main.go` (like the output of `goagen/meta`, but inside the
project).

The goal of having a `gen/main.go` is to ease team discovery, and to
more easily work on our own generators.

The goal is to make this work: https://gist.github.com/abourget/06c803e1e9e102f7dec4


raphael:
Thank you!

Issue #274, opened 2016-02-08
Make "Consumes" and "Produces" DSL more flexible


raphael:
Currently one can specify a custom encoder/decoder in the DSL using something like this:
```go
var _ = API("myapi", func() {
    // ...
    Produces("application/vnd.foo", func() {
        Package("github.com/my/custom/encoder")
    })
    Consumes("application/vnd.bar", func() {
        Package("github.com/my/custom/decoder")
    })
})
```
The encoder/decoder package then needs to expose a single method `EncoderFactory` or `DecoderFactory`. These methods must return a data structure that implements the `goa` `Encoder` or `Decoder` interface respectively.

This works OK for basic cases but results in having to create different packages for instantiating the same encoder or decoder with different configurations. For example see https://github.com/goadesign/encoding/pull/1 - two packages that contain code that are exact copies except for the one flag used to instantiate the underlying decoder.

So the proposal is to make the DSL more flexible so that one may specify the  name of the factory function and potentially arguments as well, something like:

```go
var _ = API("myapi", func() {
    // ...
    Produces("application/vnd.foo", func() {
        Package("github.com/my/custom/encoder")
        Function("UncompressedEncoder")
        Arguments("true", "1", `"string"`, "pgk.Constant")
    })
})
```

The DSL for arguments suggested above is not that great, maybe there's something better than can be done?

derekperkins:
I agree that it isn't great, but I don't know any other better way to deal with it. My only thought looking at the proposed DSL is maybe including the actual args rather than strings, but then you start getting into import issues for `goagen`.

raphael:
One problem with actual arg is that if you wanted to refer to constant in the encoder package (which is probably a common case) then you'd have to import that package in your design. Not great.

derekperkins:
Yeah, that was my point.

derekperkins:
I don't know if it's too bad though. A pro to importing the package is that we should be able to statically verify that they are generating a function that works instead of waiting to compile afterwards.

raphael:
#340 adds the ability to set the encoder and decoder producing functions in the DSL:
```go
var _ = API("myapi", func() {
    // ...
    Produces("application/vnd.foo", func() {
        Package("github.com/my/custom/encoder")
        Function("NewUncompressedEncoder")
    })
})
```
No argument. Instead the encoding package can expose functions to tweak the encoder/decoder produced by the function. Closing for now.

PR #340, opened 2016-03-04
Encoding redux.

raphael:
Replace factory interface with function which provides for
a much more lightweight extension point.
The DSL has been extended to provide the ability for
overridding the default encoder and decoder producing functions
(NewEncoder and NewDecoder respectively).

So that designer does not have to worry about
whether or not slash prefixes or trailing slashes
are required.


raphael:
Fixes #274

PR #348, opened 2016-03-09
Switch to dimfeld/httptreemux to allow for overlapping path definitions

raphael:


bketelsen:
amazing how small that change was

PR #351, opened 2016-03-10
Security handlers, per-action middleware

abourget:
This implements `Security()` at the `API`, `Resource` and `Action` level. `OAuth2Security()`, `BasicAuthSecurity()`, `APIKeySecurity()` and `OtherSecurity()` at the API definition level.

Remaining:
* [X] Support variable assignment of `var whateverSecurity = OAuth2Security("whatever")` and reuse with `Security(whateverSecurity)`  in the Resources/Actions.
* [x] Figure out what we can do with `NoSecurity`, and perhaps set a `Kind` on the `SecurityDefinition`, and add a `Kind` that would be `NoSecurity` or something.. It would pass the `Kind` of the corresponding `SecurityMethod` (but it needs to be defined first!?)
* [x] Figure out what we'll do with `Flow` and all..
* [x] Implement DSL validation, ensuring `SecurityMethod`s exist for all `Security` invocations.
* [x] Add tests for DSL
* [x] Add example in `goa-cellar`, using /auth with BasicAuth and some endpoints with a JWT.
* [x] Swagger generation
* [x] Implement `NoSecurity()` that would cancel the resource-defined fallback.


raphael:
Thank you for doing this, good stuff! It would also be nice to update client generation so the proper signers get automatically configured and registered.

abourget:
Raphael, it fails because of cyclomatic complexity in a function I don't feel safe refactoring. Mind taking a look ?

abourget:
Added to goa-cellar: https://github.com/goadesign/goa-cellar/pull/4

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/goadesign/goa/dslengine"
+)
+
+// Security defines an authentication method required to access a goa
+// Action.  When defined on a Resource, it applies to all Actions,
+// unless overriden by individual actions.  When defined at the API
+// level, it will apply to all resources by default, following the
+// same logic.
+//
+// The method name refers to what you have passed to the different
```
raphael: I guess that will become "The method refers..." once `method` becomes `interface{}` and the code checks whether it's a `SecurityDefinition` or the name of one. At that point the code should check that the name corresponds to an existing security definition.

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+// BasicAuthSecurity defines a "basic" security method for the API.
+//
+// Example:
+//
+//     BasicAuthSecurity("password", func() {
+//         Description("Use your own password!")
+//     })
+//
+func BasicAuthSecurity(name string, dsl ...func()) {
+	if _, ok := dslengine.CurrentDefinition().(*design.APIDefinition); !ok {
```
raphael: Shouldn't this be testing whether this is a `TopLevelDefinition` ?

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+//
+// Example:
+//
+//     APIKeySecurity("jwt", func() {
+//			Description("Use your own password!")
+//    		InHeader("Authorization")
+//     })
+//
+func APIKeySecurity(name string, dsl ...func()) {
+	if _, ok := dslengine.CurrentDefinition().(*design.APIDefinition); !ok {
```
raphael: Same as BasicSecurity (should be a `TopLevelDefinition`)

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+//     OAuth2Security("googAuth", func() {
+//	    	OAuth2Flow("accessCode")
+//	    	AuthorizationURL("http://example.com/authorization")
+//	    	TokenURL("http://example.com/token")
+//	    	Scope("my_system:write", "Write to the system")
+//	    	Scope("my_system:read", "Read anything in there")
+//     })
+//
+func OAuth2Security(name string, dsl ...func()) {
+	if _, ok := dslengine.CurrentDefinition().(*design.APIDefinition); !ok {
```
raphael: Same as `BasicAuthSecurity`

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+//
+//     OtherSecurity("jwt", "apiKey", func() {
+//	    	InHeader("Authorization")
+//	    	AuthorizationURL("http://example.com/authorization")
+//	    	Scope("my_system:write", "Write to the system")
+//	    	Scope("my_system:read", "Read anything in there")
+//     })
+//
+func OtherSecurity(name string, securityType string, dsl ...func()) {
+	if _, ok := dslengine.CurrentDefinition().(*design.APIDefinition); !ok {
```
raphael: Same as `BasicAuthSecurity`

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// Scope defines an authorization scope. Used within SecurityMethod,
+// the description is required, explaining what the scope
+// means. Within a Security block, only a scope is needed.
+func Scope(name string, desc ...string) {
+	switch parent := dslengine.CurrentDefinition().(type) {
+	case *design.SecurityDefinition:
+		if len(desc) == 1 {
+			dslengine.ReportError("must not pass description to Scope()")
```
raphael: I would make the error message more generic, something like "too many arguments"

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+			return
+		}
+		parent.Scopes = append(parent.Scopes, name)
+	case *design.SecurityMethodDefinition:
+		if len(desc) == 0 {
+			dslengine.ReportError("no description provided for Scope()")
```
raphael: Descriptions are usually optional, is there a reason it has to be required here?

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+		parent.Scopes[name] = desc[0]
+	default:
+		dslengine.IncompatibleDSL()
+	}
+}
+
+// InHeader defines both `in` and `name` of a SecurityMethod.  See
+// http://swagger.io/specification/#securitySchemeObject for details.
+func InHeader(headerName string) {
```
raphael: Could we just use `Header` here? We'd have to add a case to the current `Header` function but that seems more inline with how the DSL currently works.

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+			parent.Name = headerName
+			return
+		}
+	}
+	dslengine.IncompatibleDSL()
+}
+
+// InQuery defines both `in` and `name` of a SecurityMethod. See
+// http://swagger.io/specification/#securitySchemeObject for details.
+func InQuery(parameterName string) {
```
raphael: Same as comment as `InHeader` - Just `Query` seems to flow better.

design/apidsl/security.go
```diff
@@ -0,0 +1,308 @@
...
+			parent.In = "query"
+			parent.Name = parameterName
+			return
+		}
+	}
+	dslengine.IncompatibleDSL()
+}
+
+// OAuth2Flow defines the `flow` for a SecurityMethod.  See
+// http://swagger.io/specification/#securitySchemeObject for details.
```
raphael: As a high level comment, I am not sure about referring to the swagger spec in the DSL. The DSL does not currently rely explicitly on the Swagger specification and that seems to make sense to me.  I see the goa DSL as a super set and Swagger being just one way to "project" it. With that model it seems better to me to define the DSL independently of Swagger. Maybe that doesn't change anything to the DSL and only the comments or it could actually mean making the DSL more independent / meaningful on its own which wouldn't be a bad thing.

raphael: Somewhat related but different: I am not sure about leaving "flow" as an open string. As a designer I don't know what to put there. It would be better if it was either a enum value or maybe even better if `OAuth2Flow` was broken down into specific DSL functions for each flow. Maybe we need a generic version - not sure though.

design/definitions.go
```diff
@@ -652,9 +664,12 @@ func (r *ResourceDefinition) DSL() func() {
 }
 
 // Finalize is run post DSL execution. It merges response definitions, creates implicit action
-// parameters, initializes querystring parameters and sets path parameters as non zero attributes.
+// parameters, initializes querystring parameters, sets path parameters as non zero attributes
+// and sets the fallbacks for security methods.
 func (r *ResourceDefinition) Finalize() {
 	r.IterateActions(func(a *ActionDefinition) error {
+		a.Finalize()
```
raphael: The code cannot do that (call `Finalize` on other definitions) otherwise we risk infinite loops. Instead there should be an additional "DefinitionSet" for the security definitions that is returned last to the DSL engine when iterating the design definition sets.

design/security.go
```diff
@@ -0,0 +1,76 @@
...
+	// OtherSecurityKind is for custom security types, when you want to above
+	// and beyond Swagger.
+	OtherSecurityKind
+)
+
+// SecurityDefinition defines security requirements for an Action
+type SecurityDefinition struct {
+	// NoSecurity is true when you don't want to inherit the resource
+	// (or API-level) security requirements.
+	NoSecurity bool
```
raphael: Do we need an explicit bool field for this? couldn't that just be another security kind?

design/security.go
```diff
@@ -0,0 +1,76 @@
...
+// information.
+type SecurityMethodDefinition struct {
+	Kind SecurityMethodKind
+
+	// Method is the name of the security method, referenced in
+	// Security() declarations. Ex: "googAuth", "my_big_token", "jwt".
+	Method string `json:"method"`
+
+	// Type is one of "apiKey", "oauth2" or "basic", according to the
+	// Swagger specs.
```
raphael: Same comment as before wrt referring to Swagger

security.go
```diff
@@ -0,0 +1,186 @@
...
+	Flow string
+	// TokenURL defines the OAuth2 tokenUrl.  See http://swagger.io/specification/#securitySchemeObject
+	TokenURL string
+	// AuthenticationURL defines the OAuth2 authenticationUrl.  See http://swagger.io/specification/#securitySchemeObject
+	AuthenticationURL string
+	// Scopes defines a list of scopes for the security method.
+	Scopes []string
+}
+
+// Type returns the security method type.
```
raphael: Is that really needed? Don't  see it actually being used anywhere

security.go
```diff
@@ -0,0 +1,186 @@
...
+		return
+	}
+	return scopes
+}
+
+// securityMiddleware represents a security method middleware, used
+// for deduplicating the different security method's methods!
+type securityMiddleware struct {
+	middleware Middleware
+	method     interface{}
```
raphael: Where is `method` used?

security.go
```diff
@@ -0,0 +1,186 @@
...
+	if !ok {
+		return
+	}
+	return scopes
+}
+
+// securityMiddleware represents a security method middleware, used
+// for deduplicating the different security method's methods!
+type securityMiddleware struct {
+	middleware Middleware
```
raphael: Could that be an anonymous field instead? (i.e. a securityMiddleware is a middleware rather than has a middleware)

security.go
```diff
@@ -0,0 +1,186 @@
...
+	method     interface{}
+
+	// MethodName is the name specified for this authentication method
+	// which will appear in the Swagger file.
```
raphael: Reference to swagger again, consider this in the absolute instead - in this case do we need te MethodName here? We need it in the design definitions - not sure we need it at runtime?

design/apidsl/security.go
abourget: Yes..

We want that in the first merge ? Seems like a non breaking change / feature we can add right after.

design/apidsl/security.go
abourget: Again, this would be in order to reuse the thing as a var, returned by the `*Security()`, right ?

I can make sure it's in before merge if you like. Would we keep the option of having it inside API too ?

design/apidsl/security.go
abourget: Well yes, in Swagger it becomes a `map[string]string` and it requires a description.

This invocation defines the available scopes throughout the API.. otherwise there's not much benefit to define it, other that to say what it is.

design/apidsl/security.go
abourget: Yeah, tough for me to tell as I don't use OAuth2.

I understand your comment about Swagger, and us wanting to go above the spec and I agree with that. Even with the JWT case I have at hand, which needs to use the `OtherSecurity` method to work - I'd probably prefer to have a JWTSecurity and model all the scope thing on top - but the main reason I don't just do my own middleware and stick it in with `service.Use` directly, is that I want the swagger UI and swagger tooling.. so in a sense, being very close to the spec is desirable, especially for the cases that use its constraints, in order to be able to generate a meaningful and fully working Swagger UI out of the DSL.

Now this doesn't prevent from using a DSL for the Flow values, or describing more of what we'd find in the spec directly in our comments. Both of those make sense.

I don't see much value in drifting from the precise layout of swagger either, unless we have use cases.

OAuth2Flow could be just Flow.. this is where I took the liberty to namespace according to its semantic: it makes sense only in OAuth2. Not attached to it though :-)

I see that having some more refined Flow functions could make the DSL more robust, and discoverable.. I'm not strongly opinionated about it since I don't currently use it :-)

Regarding InQuery and InHeader.. I think itnwould make sense. Isn't there a risk that Header's semantic becomes unclear ? Sometimes, I find this leads to confusion, not knowing the effect of certain DSL funcs in certain contexts, especially when some are aliases of others. I remember being confused when I read that a Member was an alias to Attribute.

(I'm on mobile and the UI refreshes when I submit a note and I lose track of where I was so I did all comments in here)

design/definitions.go
abourget: I understand.. 3 lines later we do the same with Responses so I figured the pattern was: one or the other. I don't mind using the DefinitionSet instead. Just tell me what you prefer going forward.

security.go
abourget: Ok the last few comments I need to check. Some of these are used in the Middleware implementations (like the method empty interface, used by the middleware to get the YourNameSecurity to know where to look in headers for example..)..

Regarding the "Type" comment: I'm not sure.

perhaps MethodName we don't need, would be better not to have it in runtime indeed.

About securityMiddleware being a goa.Middleware, we couldn't if we require the 'method' field, right ? Unless you meant embedding the goa.Middleware ?

Regarding the Kind instead of NoSecurity: I'd need to think that through.. there's no Kind on this object right now, not that it couldn't have one.. but something needs to be on that object, because it is the one doing the inheritance tricks.. and the only thing we need to do this is to know if you overrode Security. That being said, knowing the Kind could be an additionnal info for generators, just not one we'd use.

Let's sync up later on Slack.. I'll rework things this week.

design/apidsl/security.go
raphael: Yeah that's fine, doesn't seem like a lot of work to add it.

design/apidsl/security.go
raphael: Sounds like we are on the same page: would be good to be able to generate a swagger spec that is as close as the design as possible but we shouldn't compromise the DSL for that goal either. My initial beef was more with flow just being a string. The concept of "flow" is specific to Swagger and would not make sense to someone that is not familiar with it so it would be good to avoid it in the DSL.

design/definitions.go
raphael: Oh whoops, funny how one can be more permissive with his own code :P
If it's simple I think DefinitionSet make more sense.

security.go
raphael: I meant embedding `goa.Middleware`. OK I hadn't realized there was no Kind - leave it the way it is if that ends up being too complicated.

security.go
abourget: it's used in `OAuth2Security()` 's `Use()` method.. to pass it on to the middlewares.

raphael:
Looks great! thank you for all the work.

PR #352, opened 2016-03-11
Update error handling

raphael:
So that error responses are serialized using the default
encoder rather than hardcoded to JSON.

### DIFF
--- a/error.go
+++ b/error.go
@@ -20,36 +20,37 @@
 package goa
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"strconv"
 	"strings"
 )
 
 type (
-	// ErrorID is an enum listing the possible types of errors.
-	ErrorID int
-
-	// TypedError describes an error that can be returned in a HTTP response.
-	TypedError struct {
-		ID   ErrorID
-		Mesg string
+	// HTTPError describes an error that can be returned in a response.
+	HTTPError struct {
+		// Class of error
+		*HTTPErrorClass
+		// Err describes the specific error occurrence.
+		Err string `json:"err" xml:"err"`
 	}
 
-	// MultiError records multiple errors.
+	// HTTPErrorClass contains information sent together with the error message in responses.
+	HTTPErrorClass struct {
+		// ID identifies the class of errors for clients.
+		ID int `json:"id" xml:"id"`
+		// Title is a human friendly title for the class of errors.
+		Title string `json:"title" xml:"title"`
+		// Status is the HTTP status code used by responses that cary the error.
+		Status int `json:"status" xml:"status"`
+	}
+
+	// MultiError is an error composed of potentially multiple errors.
 	MultiError []error
-
-	// BadRequestError is the type of errors that result in a response with status code 400.
-	BadRequestError struct {
-		Actual error
-	}
 )
 
 const (
 	// ErrInvalidParamType is the error produced by the generated code when
 	// a request parameter type does not match the design.
-	ErrInvalidParamType = iota + 1
+	ErrInvalidParamType = iota
 
 	// ErrMissingParam is the error produced by the generated code when a
 	// required request parameter is missing.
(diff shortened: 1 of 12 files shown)


error.go
```diff
@@ -94,221 +95,199 @@ const (
...
-}
+	// ErrInternal is the error produced when unexpected conditions
+	// occur (i.e. bug).
+	ErrInternal
+)
 
-// Error implements error.
-func (b *BadRequestError) Error() string {
-	return b.Actual.Error()
+// CAUTION: classes must be listed in the order as the constants above.
```
abourget: are they still above ?

error.go
```diff
@@ -94,221 +95,199 @@ const (
...
+	{ErrInvalidAttributeType, "invalid attribute type", 400},
+	{ErrMissingAttribute, "missing required attribute", 400},
+	{ErrInvalidEnumValue, "invalid value", 400},
+	{ErrMissingHeader, "missing required HTTP header", 400},
+	{ErrInvalidFormat, "value does not match validation format", 400},
+	{ErrInvalidPattern, "value does not match validation pattern", 400},
+	{ErrInvalidRange, "invalid value range", 400},
+	{ErrInvalidLength, "invalid value length", 400},
+	{ErrInvalidEncoding, "invalid request encoding", 400},
+	{ErrInternal, "internal error", 500},
```
abourget: how will users define their own `HTTPError` ?

error.go
abourget: to avoid future problems, we could have a `lastClassIndex` in the `const()` clause, and in an `init`, assign each of those values by index:
```go
func init() {
  classes = make([]*HTTPErrorClass, lastClassIndex - 1)
  classes[ErrInvalidParamType] = &HTTPErrorClass{ErrInvalidParamType, "invalid parameter value", 400}
  ...
}```

File: error.go (hunk 2 above)
Comment (abourget) [inline error.go]: could we do something like:

```go
var ErrBrokenTag = NewErrorClass("broken_tag_error", 401)
```

and use it like:
```go
return ErrBrokenTag("this didn't work: %s", "hello")
```

a sort of `fmt.Errorf()` emulation, I'd like that :)

error.go
```diff
@@ -20,295 +20,233 @@
...
+	errInvalidAttributeTypeID = "goa.iat"
+	errMissingAttributeID     = "goa.mat"
+	errInvalidEnumValueID     = "goa.iev"
+	errMissingHeaderID        = "goa.mhe"
+	errInvalidFormatID        = "goa.ifo"
+	errInvalidPatternID       = "goa.ipa"
+	errInvalidRangeID         = "goa.ira"
+	errInvalidLengthID        = "goa.ile"
+	errInvalidEncodingID      = "goa.ien"
+	errInternalID             = "goa.int"
```
abourget: two things here:

1. what is the use of such constants, since they are only used below, and only once ?
2. I'd be more comfortable having generic error codes, rather than codes showing the internal system I'm using. Also, `ipt` as a human-readable machine error code is not descriptive at all. I understand it comes with another message though, but what is its use ?

error.go
```diff
@@ -20,295 +20,233 @@
...
 }
 
-// Error implements error.
-func (b *BadRequestError) Error() string {
-	return b.Actual.Error()
+// Errorf builds an HTTP error given a format and values.
+func (class *ErrorClass) Errorf(fm string, v ...interface{}) *HTTPError {
+	msg := fmt.Sprintf(fm, v...)
+	return &HTTPError{ErrorClass: class, Err: msg}
 }
```
abourget: those two (`Error` and `Errorf`) are superb ! I love it... simple and extensible..

goagen/codegen/validation.go
```diff
@@ -244,28 +244,28 @@ const (
 	enumValTmpl = `{{$depth := or (and .isPointer (add .depth 1)) .depth}}{{/*
 */}}{{if .isPointer}}{{tabs .depth}}if {{.target}} != nil {
 {{end}}{{tabs $depth}}if !({{oneof .targetVal .values}}) {
-{{tabs $depth}}	err = goa.InvalidEnumValueError(` + "`" + `{{.context}}` + "`" + `, {{.targetVal}}, {{slice .values}}, err)
+{{tabs $depth}}	err = goa.BuildError(err, goa.InvalidEnumValueError(` + "`" + `{{.context}}` + "`" + `, {{.targetVal}}, {{slice .values}}))
```
abourget: will users need to call `BuildError` too to reuse `goa.InvalidEnumValueError` ?

abourget: oh I see, it's only to stack up errors, never mind..

abourget: then would `StackErrors` would be a good name ?

error.go
abourget: why not get rid of those `goa.ipt` IDs, convert the strings we have `missing required parameter` to `missing_required_parameter`, making it clear it is an identifier rather than a freeform message.

We will have a freeform error description through `Errorf()` or `Error()` .. let that be the "title" or "description" or "error_message". What about `error` instead of `id` ? is it always clear that it is an error object that we have at hand ?

I'd imagine:

```
{"error": "missing_parameter_thing",
 "error_message": "The parameter 'superbob' was missing from 'whatever.struct'",
 "status": 400}
```

and have the status reflect that.. (or no `status` prop. in there).

However a user wants to display that error with some custom error handlers, there's no confusion as to the nature of that object :)

raphael: The idea is that "id" is used by programs while "title" helps humans. I agree that the constants are not necessary and could just be inlined but I do believe there is value in distinguishing between machine and human consumed error class descriptions (which is what `ID` and `Title` are).

error.go
raphael: Thanks! :)

goagen/codegen/validation.go
raphael: I like it - will change it.

error.go
abourget: I understand, but `title` could be the title of a blog post, and `id` the primary key of something else.. Those names are not very error-centric.  `error_code` (for machine readable) and `error_message`, with _message_ being directed towards humans is what I see usually.. could be `error_title` but.. it's not really a title either, since it can contain all the details..

Issue #354, opened 2016-03-14
Use code generation to handle CORS


raphael:
The middleware spec looks very DSLish already. Also there are issues with the router not allowing us to define routes with `*cors` wildcards where we need it.

PR #355, opened 2016-03-14
Add CORS support

raphael:
PREREQ READING: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

goa DSL now allows specifying CORS policies global to the API
or specific to resources. The DSL for defining the CORS policies
is as follows:
```go
Origin("http://swagger.goa.design", func() { // Origins can use wildcards, "*" means all origins
    Headers("X-Shared-Secret")               // Allowed headers
    Methods("GET", "POST")                   // Allowed HTTP methods
    Expose("X-Time")                         // Headers exposed to browser
    MaxAge(600)                              // Time preflight request response is cached for
    Credentials()                            // Whether CORS requests can be made with credentials
})
```
The controller actions that have associated CORS policies are
wrapped into a middleware that adds the required header for simple
CORS requests. CORS preflight request handlers are also generated
for each of the action route that is not using the OPTIONS verb.


raphael:
Fixes #354 (work in progress - DO NOT MERGE)

PR #358, opened 2016-03-16
Simplification of the config protocol for security handlers.

abourget:
WARN: this includes the DetailedError thing.. 

Updated the `middleware` for JWT too in https://github.com/goadesign/middleware/pull/26.. (didn,t update the `basicauth` yet).


raphael:
Merged with #359

PR #360, opened 2016-03-16
Change how ErrorClass works to allow for Key/Values

raphael:
The usage is now as follows:

1. Create an error class, this returns a function
2. Use the function to create errors
3. Call `Meta` on the errors to add key/value pairs

The ErrorClass function accepts a format and values a la `Printf`.
The format may be any type which makes it simple to wrap "raw"
errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ public/
 .vscode/
 *~
 *.orig
+.DS_Store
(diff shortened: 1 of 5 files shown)


abourget:
That ..  is ... awesome.
:shipit:

raphael:
Thank you for the help designing this!

PR #362, opened 2016-03-18
Generate graceful in new services (gen_main)

abourget:
Also added Timeout to ListenAndServe for graceful services.

(there's no `SetTimeout` anymore).


raphael:
Awesome! thank you.

PR #364, opened 2016-03-22
Error media

raphael:
This PR makes is possible for designs to use the new goa built-in `ErrorMedia` media type.
This media type describes a data structure compatible with the JSON serialization of the
`HTTPError` struct. 

The goa app generator uses `*goa.HTTPError` and `goa.MultiError` to define the response
methods for responses whose design use `ErrorMedia` and `CollectionOf(ErrorMedia)` 
respectively.


abourget:
Lgtm :-)

PR #367, opened 2016-03-25
Fix security empty scopes panic

megabuz:
I have a panic error when i use jwt security without Security.Scopes in handleSecurity
I can change this type assertions or add scopes in context (line 732) even it empty


raphael:
This is great, thank you for the fix!

PR #376, opened 2016-03-29
Protect against request bodies with unreasonable length

### LINKED ISSUE #332: Protect against large request bodies
The following middleware is used by other web frameworks to protect against requests with unreasonable body length:
```go
// PayloadSizeMiddleware tries to read the request body and fails if the size is too big.
func PayloadSizeMiddleware(next http.Handler) http.Handler {
 maxSize := int64(conf.GetSizeInBytes(confMaxBodySize))
 return http.HandlerFunc(func(wrt http.ResponseWriter, req *http.Request) {
  body, err := ioutil.ReadAll(http.MaxBytesReader(wrt, req.Body, maxSize))
  if err != nil {
                    ....
               }
 })
}
```
This doesn't work as is with goa because of the way encoding works: the decoders are responsible for reading the request body. The middleware could reset the body reader or there could be an option to configure the decoders to do the body length check.

raphael:
Fix #332

### DIFF
--- a/service_test.go
+++ b/service_test.go
@@ -56,6 +56,39 @@ var _ = Describe("Service", func() {
 		})
 	})
 
+	Describe("MaxRequestBodyLength", func() {
+		var oldMax int64
+		var rw *TestResponseWriter
+		var req *http.Request
+		var muxHandler goa.MuxHandler
+
+		BeforeEach(func() {
+			oldMax = goa.MaxRequestBodyLength
+			goa.MaxRequestBodyLength = 4
+			body := bytes.NewBuffer([]byte{'"', '2', '3', '4', '"'})
+			req, _ = http.NewRequest("GET", "/foo", body)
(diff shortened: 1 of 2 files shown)


service_test.go
```diff
@@ -56,6 +56,39 @@ var _ = Describe("Service", func() {
...
+		JustBeforeEach(func() {
+			muxHandler(rw, req, nil)
+		})
+
+		AfterEach(func() {
+			goa.MaxRequestBodyLength = oldMax
+		})
+
+		It("prevents reading more bytes", func() {
+			Ω(string(rw.Body)).Should(Equal(`{"code":"invalid_encoding","status":400,"detail":"http: request body too large"}` + "\n"))
```
arangamani: Wouldn't `413` be the appropriate status code in this case?

raphael:
Yes it probably would, good point! I'll change it.

PR #378, opened 2016-03-30
Change "request too large" response to use status 413.

raphael:
and better detail message.


raphael:
Address comment in #376

PR #380, opened 2016-03-31
Properly handle params coming in as slices of multiple elements.

### LINKED ISSUE #365: Use multiple parameters with ArrayOf in Params in an Action
Right now, an ArrayOf(String) in a Parameter creates a string slice in generated code, by splitting the first parameter on ",". It seems not  possible to get the other values passed in in the request with multiple Params of the same name.

Since it is easy to split on a comma in client code, given a simple string, Goa would be better off supporting the more complex case of handling multiple parameters with the same name as ArrayOf(String).

raphael:
Fix #365

### DIFF
--- a/goagen/gen_app/generator_test.go
+++ b/goagen/gen_app/generator_test.go
@@ -362,8 +362,9 @@ func NewGetWidgetContext(ctx context.Context) (*GetWidgetContext, error) {
 	var err error
 	req := goa.ContextRequest(ctx)
 	rctx := GetWidgetContext{Context: ctx, ResponseData: goa.ContextResponse(ctx), RequestData: req}
-	rawID := req.Params.Get("id")
-	if rawID != "" {
+	paramID := req.Params["id"]
+	if len(paramID) > 0 {
+		rawID := paramID[0]
 		rctx.ID = rawID
 	}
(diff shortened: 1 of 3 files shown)


abourget:
That's slightly better, I agree.. but it does violate the principle of least surprise.  I can imagine someone starting to use the multi-param feature, and gets bitten in the hoompa-doompa when he discovers that some values were split by `,` .. don't you think ?

Wouldn't we be better to have a different DSL to specify each behavior ?

abourget:
`ArrayOf` and `MultipleKeys` or `MultipleParams` or `FormValues` or `ArrayOfParams` or `ParamsOf` or something ?

abourget:
or add something to `ArrayOf` like: ```ArrayOf(Something, func() {
    MultipleParamsBehavior()
})```

(I know it's not pretty)

raphael:
I understand what you are saying and agree. On the other hand there's also an argument to be made about wanting to keep things simple and avoiding adding yet more DSL (it's always the easy solution). Using commas in query strings doesn't seem like something that would be a common thing to do. I would be more in favor of properly documenting this behavior for now and address the need for not splitting on query string values when a real use case where there's good alternative comes up.

Issue #349, opened 2016-03-09
Rework context and type code generation to enable proper "required" and "default" attribute semantic


raphael:
# Background

Today the design DSL makes it possible to define attributes as being required:
```go
var MyType = Type("my_type", func() {
    Attribute("foo")
    Required("foo")
})
```
This has two consequences:
1. The generated `Validate` method checks that the value is initialized (the string is not empty in the example above) and if not returns an error.
2. The generated struct fields do not use a pointer (in the example above the Go type has a `foo` field of type `string`.

The generated Go type for the design above is thus:

```go
type MyType struct {
    Foo string `json:"foo" xml:"foo"`
}
```

# Problem Statement

The combination of the two consequences listed above results in incorrect behavior when the attribute type is a primitive type other than string. Taking integers as example if an integer attribute in the design is defined as being required the corresponding Go struct field is of type `int`. This makes it impossible to distinguish between the case where the zero value was set in the incoming request or created when decoding. As a result the generated `Validate` code does not check that these attributes are properly initialized in the incoming request state.

To complicate things further the DSL also allows specifying default values for attributes (the implementation of which is being worked in #203). The same issue arises for non pointer basic type fields: the code setting the default values cannot tell whether the zero value was intended or a result of the decoding.

# Goals

1. The generated code must always validate that the shape of incoming requests is correct as defined in the design.
2. The final code should avoid forcing dereferencing pointers for all basic types.
3. The generated code should not copy request state after deserialization.

# Proposal

The only way to achieve goal 1 without having special logic in every decoder is to use pointers to store the basic type values. This makes it possible for the generated `Validate` and `Finalize` methods to check for fields not provided in the request state in a generic fashion.

Achieving goal 2 means that there has to be another way to get at the values - the proposal is thus to introduce functions that do the dereferencing. For attributes that are not required and don't have a default value the proposal is to generate additional `HasXXX` methods that return a boolean indicating whether the underlying value is nil or not. This is so that the signature of attribute getters always return non pointer values. Note that under this proposal fields remain accessible but renamed to `XXXField` which is always of type pointer (unless the field holds a slice or a map).

So the type shown as example in the background section would now generate:
```go
type MyType struct {
    FooField *string `json:"foo" xml:"foo"`
}

func (t *MyType) Foo() (v string) {
    if t.FooField != nil {
        v = *t.FooField
    }
    return
}
```
Optional fields would also generate:
```go
func (t *MyType) HasBar() bool {
    return t.BarField != nil
}
[...]

PR #383, opened 2016-04-01
Two structs/default value support

### LINKED ISSUE #203: Default() on a payload member is not set in the context correctly
Having a default value on a member of a payload is expected to set a default value in the context but it seems to be simply ignored. Example:
```go
Member("debug", Boolean, "Whether to enable debugging", func() {
	Default(false)
})
```
Also in this case, I would expect that the member will not be a pointer though it is not a required member since it has a default value.

arangamani:
Fixes #203 and #349 

## Overview
This PR introduces the two struct approach discussed in #349 where we create two different structs for payloads and user types. One of these structs will be private and will use pointers for all fields and the other struct will be public (the only one we currently have) and will use pointers for only fields that are not required and don't have any default values for them. Then goa will internally sets the default values on the private struct after decoding the request and then validate the private struct. After the validation succeeds, the fields in the private struct will be copied over to create the public struct and this public struct will be set in the context.

## What to expect after this change:
- Fields that have default value will not be pointers anymore
- Previously we were only checking if the field is present (required) only for string primitive. Now we will check this for all primitives.


design/types.go
```diff
@@ -288,12 +308,17 @@ func (a *Array) Kind() Kind { return ArrayKind }
 
 // Name returns the type name.
 func (a *Array) Name() string {
-	return "array"
+	return fmt.Sprintf("array")
```
arangamani: Hmmm, I'll change it back.

raphael:
Awesome work! thank you.

PR #391, opened 2016-04-07
Add View name to Media

megabuz:


raphael:
As per our discussion on slack please change the PR so that if no view is no provided then response methods are generated for all views instead of the just the default view. Other than that looks great, thank you!

raphael:
@megabuz any change you may be able to update this PR and apply the tweak above? :)

raphael:
Closing this so I can create a new PR with the fix.

PR #392, opened 2016-04-08
Change the usage for design flag of goagen

tchssk:
This is a changing for beginners of `goagen` like me.


raphael:
Thank you for the PR. Would it help if the help message said instead "design package import path" ? At the end of the day that's what it is.

tchssk:
That's better. I fixed it and pushed again.

raphael:
Awesome, thanks!

Issue #395, opened 2016-04-08
log source IP of requests in LogRequest


tve:
In https://github.com/goadesign/goa/blob/master/middleware/log_request.go#L31 add `"from", req.RemoteAddr` (dunno whether that takes an X-Forwarded-For header into account).

PR #399, opened 2016-04-09
Improve request logger middleware.

raphael:
* Add "from" key with client IP
* Only log controller and action in "started" log
* Use controller name instead of resource name


raphael:
Fix #395

Issue #404, opened 2016-04-12
Setting the example of an attribute causes panic.


ikawaha:
Following design code causes panic.

```
Attribute("testtest", HashOf(String, Any), func() {
        Example(map[string]interface{}{
                        "hello": "world",
        })
})
```

log
```
$goagen app
exit status 2
panic: unknown primitive type

goroutine 1 [running]:
panic(0x932ca0, 0xc8204e8800)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/goadesign/goa/design.Primitive.IsCompatible(0x6, 0x932ca0, 0xc8204e87d0, 0xc8204e8701)
        /home/me/lib/Go/src/github.com/goadesign/goa/design/types.go:280 +0x15b
github.com/goadesign/goa/design.(*Primitive).IsCompatible(0xc8204e8778, 0x932ca0, 0xc8204e87d0, 0x932ca0)
        <autogenerated>:179 +0xa3
github.com/goadesign/goa/design.(*Hash).IsCompatible(0xc8204e8780, 0x931720, 0xc8204ea5a0, 0xc820178a80)
        /home/me/lib/Go/src/github.com/goadesign/goa/design/types.go:495 +0x349
github.com/goadesign/goa/design.(*AttributeDefinition).SetExample(0xc8204e7c20, 0x931720, 0xc8204ea5a0, 0x30)
        /home/me/lib/Go/src/github.com/goadesign/goa/design/definitions.go:960 +0x7c
github.com/goadesign/goa/design/apidsl.Example(0x931720, 0xc8204ea5a0)
        /home/me/lib/Go/src/github.com/goadesign/goa/design/apidsl/attribute.go:249 +0x5e
```

PR #405, opened 2016-04-12
Fix lack of Any in Primitive.IsCompatible()

ikawaha:
#404
Any is primitive type.


raphael:
Great! Do you think you could rewrite the `IsCompatible` method to make it shorter? seems like the code could be written using a Type switch statement which would make it much shorter... Thank you!

raphael:
Awesome, thanks!

ikawaha:
Thank you!

PR #410, opened 2016-04-13
Client security Update

aslakknutsen:


goagen/gen_client/generator.go
```diff
@@ -664,19 +677,24 @@ func (c *Client) {{ $funcName }}(ctx context.Context, path string{{ if .Payload
...
+	*goaclient.Client{{range $security := .SecuritySchemes }}
+    Signer{{ $security.SchemeName }} goaclient.Signer{{ end }}
 }
 
 // New instantiates the client.
 func New(c *http.Client) *Client {
-	return &Client{Client: goaclient.New(c)}
+	return &Client{
+        Client: goaclient.New(c),{{range $security := .SecuritySchemes }}
+        Signer{{ $security.SchemeName }}: &{{ signerType ($security) }}{},{{ end }}
```
raphael: Please use `Goify` to sanitize the name (may not be a valid Go identifier name). Something like:

 ```go
Signer{{ goify $security.SchemeName true }}: ...
```

Also need to update where it's called (in `RegisterFlags`).

raphael:
Looks great! thank you. Left one comment about using `Goify` to sanitize the scheme name used to build a Go identifier. Once that's fixed LGTM.

aslakknutsen:
Added 'goify' to Scheme Name. Updated test case to include non go'able Scheme name to verify goifiabillity. :)

Squashed and Rebased.

PR #412, opened 2016-04-13
Add kit/log to DEPEND in Makefile

aslakknutsen:
Make test complained


raphael:
As discussed on slack the idea is to only add build and test dependencies to the Makefile (anything that `go get ./...` doesn't install).

PR #413, opened 2016-04-14
Fix: Panic on unknown type in IsCompatible

ikawaha:
I think perhaps this is better.

### DIFF
--- a/design/types.go
+++ b/design/types.go
@@ -226,6 +226,9 @@ func (p Primitive) CanHaveDefault() (ok bool) {
 
 // IsCompatible returns true if val is compatible with p.
 func (p Primitive) IsCompatible(val interface{}) bool {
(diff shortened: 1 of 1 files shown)


raphael:
Just curious: why do you think this is better? if a new type gets added then you only need to add a `case` statement and the current code just works. This change would also require changing the `if`.

ikawaha:
Following pseudo code causes panic, but p is primitive.

```
p := Number
val := []int{}
p.IsCompatible(val)
```

Panic message "unknown primitive type" indicates p is unknown, does'nt  it?
I think it appropriate to return false if val is not primitive.

This shall not apply when the usage of Primivitve.IsCompatible() is limitted.

Regards,

raphael:
Oh woops my bad, thank you.

PR #417, opened 2016-04-16
Add example to Example()

brasilikum:
Inception!


raphael:
Thank you! Do you think you could remove the blank line between the comment and the function?

raphael:
Oops one last thing I didn't catch last time: there seems to be some extra spaces before both `Attribute` function calls. Maybe tabs instead of spaces? thanks again!

PR #422, opened 2016-04-17
Add gen_test generator

aslakknutsen:
The test generator generates test helpers for the API controllers.

A test helper is generated for all combination of
Resource/Action/Route/Response/View.

The test helpers can be found in ./test/$resource.go

### DIFF
--- a/goagen/gen_test/generator.go
+++ b/goagen/gen_test/generator.go
@@ -103,7 +103,7 @@ func (g *Generator) generateResourceTest(clientPkg string, funcs template.FuncMa
 
 		for _, action := range res.Actions {
 			for _, response := range action.Responses {
-				if response.Name == design.SwitchingProtocols {
+				if response.Status == 101 { // SwitchingProtocols, Don't currently handle WebSocket endpoints
 					continue
 				}
 				for routeIndex, route := range action.Routes {
--- a/goagen/gen_test/generator.go
+++ b/goagen/gen_test/generator.go
@@ -186,7 +186,9 @@ func createTestMethod(resource *design.ResourceDefinition, action *design.Action
 		payload := ObjectType{}
 		payload.Name = "payload"
 		payload.Type = fmt.Sprintf("%s.%s", TargetPackage, codegen.Goify(action.Payload.TypeName, true))
-		payload.Pointer = "*"
+		if !action.Payload.IsPrimitive() && !action.Payload.IsArray() && !action.Payload.IsHash() {
+			payload.Pointer = "*"
+		}
 
 		validate := codegen.RecursiveChecker(action.Payload.AttributeDefinition, false, false, false, "payload", "raw", 1, true)
 		if validate != "" {
--- a/goagen/gen_test/generator.go
+++ b/goagen/gen_test/generator.go
@@ -1,0 +1,299 @@
+package gentest
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"text/template"
+
+	"github.com/goadesign/goa/design"
+	"github.com/goadesign/goa/goagen/codegen"
+	"github.com/goadesign/goa/goagen/utils"
+	"github.com/spf13/cobra"
+)
+
+// Generator is the application code generator.
+type Generator struct {
+	genfiles []string
+}
+
+// Generate is the generator entry point called by the meta generator.
+func Generate() (files []string, err error) {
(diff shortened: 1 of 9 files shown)


goagen/gen_test/generator.go
```diff
@@ -0,0 +1,299 @@
...
+func Generate() (files []string, err error) {
+	api := design.Design
+	if err != nil {
+		return nil, err
+	}
+	g := new(Generator)
+	root := &cobra.Command{
+		Use:   "goagen",
+		Short: "Test generator",
+		Long:  "controller test and package generator",
```
raphael: did you really mean for the "and" to be there?

goagen/gen_test/generator.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+		if err := file.WriteHeader("", "test", imports); err != nil {
+			return err
+		}
+
+		var methods = []TestMethod{}
+
+		for _, action := range res.Actions {
+			for _, response := range action.Responses {
+				if response.Name == design.SwitchingProtocols {
```
raphael: This should probably use the status instead of the name (names are arbitrary).

goagen/gen_test/generator.go
```diff
@@ -0,0 +1,299 @@
...
+	method.ResourceName = codegen.Goify(resource.Name, true)
+	method.Comment = fmt.Sprintf("test setup")
+	method.ControllerName = fmt.Sprintf("%s.%sController", TargetPackage, codegen.Goify(resource.Name, true))
+	method.ContextVarName = fmt.Sprintf("%sCtx", codegen.Goify(action.Name, false))
+	method.ContextType = fmt.Sprintf("%s.New%s%sContext", TargetPackage, codegen.Goify(action.Name, true), codegen.Goify(resource.Name, true))
+	method.RouteVerb = route.Verb
+	method.Status = response.Status
+	method.FullPath = goPathFormat(route.FullPath())
+
+	if view != nil && mediaType != nil {
```
raphael: can the mediaType be nil if the view isn't?

goagen/gen_test/generator.go
```diff
@@ -0,0 +1,299 @@
...
+			}
+		}
+		method.Params = params
+	}
+
+	if action.Payload != nil {
+		payload := ObjectType{}
+		payload.Name = "payload"
+		payload.Type = fmt.Sprintf("%s.%s", TargetPackage, codegen.Goify(action.Payload.TypeName, true))
+		payload.Pointer = "*"
```
raphael: A payload may not be a struct. To be fair this is not something that has been tested explicitly so it's possible other generators don't support it but it would be nice for this new one to not have the bug from the get go :)

The design supports doing:
```go
Payload(Integer)
```
for example.

goatest/testing.go
```diff
@@ -0,0 +1,34 @@
...
+// ResponseSetterFunc func
+type ResponseSetterFunc func(resp interface{})
+
+// Encode xyz
+func (r ResponseSetterFunc) Encode(v interface{}) error {
+	r(v)
+	return nil
+}
+
+// Service a
```
raphael: Looks like an incomplete comment :)

goatest/testing.go
```diff
@@ -0,0 +1,34 @@
...
+	"log"
+
+	"github.com/goadesign/goa"
+	"github.com/goadesign/goa/middleware"
+)
+
+// ResponseSetterFunc func
+type ResponseSetterFunc func(resp interface{})
+
+// Encode xyz
```
raphael: How about
```go
// Encode implements a dummy encoder that returns the value being encoded
```

goagen/gen_test/generator.go
aslakknutsen: Not based on this; https://github.com/goadesign/goa/pull/422/files/0b8212997702aac0b1a3e0e584899bcc6ea6c53b#diff-74f47f25c782559a460b41b7cc7e97a8R112

And as far as I can tell I need both to find the correct type information.. ?

goagen/gen_test/generator.go
aslakknutsen: a bit clunky perhaps, but the idea was; it generates the test helpers and the 'test' folder.

goagen/gen_test/generator.go
raphael: yes. Just pointing out that the `if` could simply be:
```go
if view != nil {
  // ...
}
```

PR #441, opened 2016-04-28
Add UUID datatype.

brycereitano:
This adds the functionality of handling UUID's as a primary key, in the routes, and payloads. This is to compliment the addition of uuid support as primary key in gorma. https://github.com/goadesign/gorma/pull/93

In addition, this fixes a small bug when removing imports. It will now properly call the right method for named and non-named imports.


bketelsen:
this looks good!

raphael:
This is great, thanks!

PR #445, opened 2016-04-29
Include vendor directory in valid pkg paths

### LINKED ISSUE #444: goagen does not use vendor directory
When vendoring a goa app the goagen tool does not pick up on the packages from the vendor directory.
It still expects package github/goadesign/goa to be in the GOPATH, while instead it is in the vendor path.

jsimonetti:
Add the vendor directory as valid path for packages to be in.
Fixes #444


jsimonetti:
Ok, obviously this needs a more idiomatic approach. I'll leave this here for further review, but this is a bit out of my league.

raphael:
Thank you, this was fixed by #480

PR #451, opened 2016-05-02
Remove `_client.go` suffix to files generated by `gen_client`.

raphael:
There is a slight risk of collision with `datatypes.go` but we'll cross
that bridge when we get there.


arangamani:
Soo I can't have a resource named `DataTypes`? 😉

raphael:
Not until you submit a PR that handles that special case :P

raphael:
https://github.com/goadesign/goa/pull/453

PR #457, opened 2016-05-03
Fix ordering of functions generated by goagen

tchssk:
Ordering of functions in `app/test` is not guaranteed to be the same from one iteration to the next.
We should use prepared keys to iterate over maps.


raphael:
Good catch, thanks! Can you update the code to use the `IterateXXX` functions instead? that's what htey are for. So `r.IterateActions` to iterate over resource actions, `a.IterateResponses` to iterate over the action responses. Thank you!

tchssk:
Oh, I failed to notice.
I fixed it to use `IterateXXX` functions.
Is it OK ?

raphael:
Yes perfect, thanks much!

PR #479, opened 2016-05-09
WIP Split ctx

aslakknutsen:
PR just to keep the conversation open.

NOTE: Tests still not updated.


aslakknutsen:
@raphael Are these https://github.com/goadesign/goa/blob/master/goagen/gen_app/writers_test.go#L839 intended to be 'copied' from goa-cellar ?

raphael:
@aslakknutsen the idea is to try and keep a common theme throughout the goa-cellar examples and tests - does not have to be an exact copy though. Copy when convenient :) There is an integration test that does a codegen and compiles the goa-cellar example on each goa commit.

aslakknutsen:
@raphael hehe right. I was just looking for the source design to do a regen based on the new changes :)

raphael:
Ah I see, yeah there is no such thing :)

raphael:
I'm closing this for now since it's getting a bit old. Thanks for the work though, should help with v2!

aslakknutsen:
Still want the change, right?

raphael:
Yes absolutely.

PR #481, opened 2016-05-09
add support for request headers in generated Swagger JSON

jim-slattery-rs:
- reusing logic between `paramsFromDefinition()` and the new `paramsFromHeaders()`
- tested various types of Header definitions and validations
- also fix a couple of "bugs":
  - "array" Parameters should use Min/MaxItems rather than Min/MaxLength
  - Minimum()/Maximum() constraints are inclusive, so ExclusiveMinimum/ExclusiveMaximum should be `false`


design/definitions.go
```diff
@@ -1413,6 +1423,20 @@ func (a *ActionDefinition) UserTypes() map[string]*UserTypeDefinition {
...
+// error.
+func (a *ActionDefinition) IterateHeaders(it HeaderIterator) error {
+	// resource-level headers
+	err := iterateHeaders(a.Parent.Headers, it)
+	if err != nil {
+		return err
+	}
+	// action-level headers
+	return iterateHeaders(a.Headers, it)
+}
```
raphael: Should the two maps be merged first? Sense like otherwise the same overridden header would be iterated twice over.

raphael:
Looks great, thank you! Just left one question/comment in one of the commits. Other than that LGTM.

design/definitions.go
jim-slattery-rs: Good point, I hadn't thought about overriding headers..

I had actually started out with merging them, but reverted because I ran into a problem with determining whether a header was `required`, since it could have come from two different places.. I'll have to give that more thought and try again.

raphael: Ideally there merge (including "requiredness" computation) should happen during the "finalize" phase so that the code doesn't need to consider the parent at generation time.

goagen/gen_swagger/swagger_test.go
```diff
@@ -402,23 +406,25 @@ var _ = Describe("New", func() {
...
+				Ω(ps[6]).Should(Equal(&genswagger.Parameter{In: "header", Name: "OptionalInt", Type: "integer", Minimum: -2, Maximum: 2}))
+				Ω(ps[7]).Should(Equal(&genswagger.Parameter{In: "header", Name: "OptionalRegex", Type: "string",
 					Pattern: `[a-z]\d+`, MinLength: 1, MaxLength: 10}))
-				Ω(ps[10]).Should(Equal(&genswagger.Parameter{In: "header", Name: "X-Account", Type: "integer", Required: true}))
+				Ω(ps[8]).Should(Equal(&genswagger.Parameter{In: "header", Name: "OptionalResourceHeaderWithEnum", Type: "string",
+					Enum: []interface{}{"a", "b"}}))
+				Ω(ps[9]).Should(Equal(&genswagger.Parameter{In: "header", Name: "OverrideOptionalHeader", Type: "string", Required: true}))
+				Ω(ps[10]).Should(Equal(&genswagger.Parameter{In: "header", Name: "OverrideRequiredHeader", Type: "string", Required: true}))
+				Ω(ps[11]).Should(Equal(&genswagger.Parameter{In: "header", Name: "X-Account", Type: "integer", Required: true}))
+				Ω(ps[12]).Should(Equal(&genswagger.Parameter{In: "header", Name: "header", Type: "string", Required: true}))
```
jim-slattery-rs: order changed since headers are now merged first, then alphabetized

raphael:
Thank you!

PR #484, opened 2016-05-10
fix nil-pointer exception in IterateHeaders()

jim-slattery-rs:
- added a test to verify


raphael:
Good catch and thanks for the test!

Issue #476, opened 2016-05-07
Unable to run goagen client if resource is named client.


illyabusigin:
Hi there,

I ran into an interesting bug where the client is unable to be generated if a resource is named client.

To reproduce:

1. Use the following DSL:
```go
package design // The convention consists of naming the design
// package "design"
import (
	. "github.com/goadesign/goa/design" // Use . imports to enable the DSL
	. "github.com/goadesign/goa/design/apidsl"
)

var _ = API("cellar", func() { // API defines the microservice endpoint and
	Title("The virtual wine cellar")    // other global properties. There should be one
	Description("A simple goa service") // and exactly one API definition appearing in
	Scheme("http")                      // the design.
	Host("localhost:8080")
})

var _ = Resource("client", func() { // Resources group related API endpoints
	BasePath("/clients")      // together. They map to REST resources for REST
	DefaultMedia(ClientMedia) // services.

	Action("show", func() { // Actions define a single API endpoint together
		Description("Get bottle by id") // with its path, parameters (both path
		Routing(GET("/:bottleID"))      // parameters and querystring values) and payload
		Params(func() {                 // (shape of the request body).
			Param("bottleID", Integer, "Bottle ID")
		})
		Response(OK)       // Responses define the shape and status code
		Response(NotFound) // of HTTP responses.
	})
})

var ClientMedia = MediaType("application/vnd.goa.example.client+json", func() {
	Description("A client")
	Attributes(func() { // Attributes define the media type shape.
		Attribute("id", Integer, "Unique client ID")
		Attribute("href", String, "API href for making requests on the client")
		Attribute("name", String, "Name of wine")
		Required("id", "href", "name")
	})
	View("default", func() { // View defines a rendering of the media type.
		Attribute("id")   // Media types may have multiple views and must
		Attribute("href") // have a "default" view.
		Attribute("name")
	})
})
```

2. Run the following goagen command:
```
goagen client -d path/to/design
```

illyabusigin:
So goa generates a client.go for the CLI which steps on any resource that’s named the same. Should `client` then be a reserved word?

raphael:
Thank you! this is now fixed.

PR #488, opened 2016-05-13
Misc fixes:

raphael:
* Protect against "client" package and local variable names clash
* Default client CLI tool version to "0" if no version in API
* Sort flags used to spawn generator tool so generated header comments are deterministic.


raphael:
Fix #476

Issue #461, opened 2016-05-04
CORS headers not added to swagger.json response


jimmidyson:
With `Origin("*", func() {...` in the API definition, CORS headers are added correctly to API responses but are not added to the /swagger.json response. This makes it trickier to use Swagger UI for API discovery & browsing, as well as dynamic client generation, etc.

raphael:
The problem is that the service `ServeFiles` method is not hooked up to the generated CORS middleware. This middleware is generated as a private function of the `app` package at the moment so you can't hook it up yourself either. Two things need to change:

1. The generated CORS middlewares should be public functions
2. The generated Swagger controller should take advantage of these functions when they exist (that is when the design specifies a CORS specification).

There should also be clear documentation around that. Thank you for bringing this up!

raphael:
Fixed by #490 - Does require changing the design, please read the comments in the PR.

PR #490, opened 2016-05-15
Add "Files" DSL for serving static assets.

raphael:
This DSL makes it possible to mount file servers onto controllers and
benefit from its CORS and/or security scheme.


raphael:
Fixes #461

raphael:
This PR adds the ability to define "file servers" on resources. A file server serves static assets from the specified file path on a given route. It can serve a single file or all files under the given file path if the route ends with a wildcard starting with `*`.  The goa cellar example takes advantage of this new DSL to serve the swagger, JSON schema, javascript and html example - see https://github.com/goadesign/goa-cellar/blob/master/design/resources.go#L7-L18. This PR removes the generation of the swagger controller - instead the design should take advantage of "Files" to specify how it should be served (as the goa-cellar example shows).

This change makes it possible to define the CORS policy and security scheme that apply to the static assets which wasn't possible before. It also helps implement a consistent strategy for serving all static assets (Swagger, JSON schema etc.) as shown in the goa-cellar example.

Note that  the PR does introduce a BREAKING change and that is that the "MountSwaggerController" is no longer generated as the Swagger dummy controller is gone as noted above in favor of the controller corresponding to the resource with the "Files" DSL. So after updating goa if you see an error about this function missing simply remove its invocation from your main (or wherever) and replace it with the "Files" DSL.

As a side note - the goa-cellar makefile shows the goagen flags used to generate the swagger, JSON schema and javascript in the location that matches the design (i.e. under the `/public` folder) https://github.com/goadesign/goa-cellar/blob/master/Makefile#L39-L44

PR #493, opened 2016-05-18
Remove need for "finalizing" service.

raphael:
Instead use closures to lazily compute middleware chains.
This makes it possible to mount controllers and middlewares in any order.


arangamani:
Yay!

Issue #431, opened 2016-04-23
wrong content type on show action


bketelsen:
calling /thing (list) gives "application/vnd.thing+json"
calling /thing/1 (show) gives "application/vnd.thing" content type

I'll look at this in a bit, just wanted to record it.

bketelsen:
same in goa-cellar:
https://github.com/goadesign/goa-cellar/blob/master/app/contexts.go

potatogopher:
@bketelsen Is the work-around implemented in goa-cellar?

raphael:
Today there's no way for goa to infer the correct media type modifier. The interface to encoder packages would have to be modified so the code generator could query them for the correct modified suffix. There is code that injects the "+json" modified but that was written prior to the introduction of generic encoders and should be removed. At this time the best option is to specify the modifier in the media type identifier in the design.

PR #495, opened 2016-05-19
Refactor service encoder and decoder.

raphael:
So that generated client code can take advantage of the objects as well.
Also make optional query string and header parameters use pointers in
client package functions.


raphael:
This PR introduces a breaking change: `Encoder` and `Decoder` are now *fields* of the `goa.Service` struct instead of methods. These fields both expose a `Register` method that does the same as what the functions were doing (and has the same signature). The generated code takes care of setting up the encoder and decoder so the change should be mostly transparent but if your code uses the `Encoder` and `Decoder` functions directly it needs to be updated to use the `Register` method instead:
```go
service.Encoder(goa.NewJSONEncoder, "application/json")
```
should now be:
```go
service.Encoder.Register(goa.NewJSONEncoder, "application/json")
```

brycereitano:
This should close #431, correct?

raphael:
I don't think so? #431 is about the server generated code inconsistently appending a "+json" modifier to the media type in responses. This PR is about encoding requests properly in the client generated code.

PR #497, opened 2016-05-20
Fix payload validations in app tests

ikawaha:
`gen_app` generates some BadRequest test, but calling a BadRequest test with invalid payload causes panic.

This change is intended to pass tests with invalid payload when expected BadRequest.


raphael:
Makes sense, thank you. @aslakknutsen may want to take a look too.

aslakknutsen:
Makes perfect sense! :)

PR #501, opened 2016-05-24
Fix potential infinite recursions during codegen.

raphael:
When a type has a field whose type is itself - which in case you are
wondering is a valid scenario :) (e.g. a "Person" type with a "friend"
attribute of type "Person").


raphael:
@arangamani fyi - can you double check the fix? and maybe add a test for it? thank you!

arangamani:
@raphael will do!

PR #505, opened 2016-05-25
Do not merge responses defined in API in all actions.

raphael:
Let actions explicitly list responses, that's what everybody expects...


xeger:
LGTM!

PR #507, opened 2016-05-27
Return err when some payloads required but no payloads specified

tchssk:
```go
Action("Update", func() {
	Payload(func() {
		Member("id")
		Required("id")
	})
})
```

The above `Action` should return `400 Bad Request` when no payloads given.
This pull request implement that.

### DIFF
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -274,6 +274,20 @@ func Params(dsl func()) {
 //	})
 //
 func Payload(p interface{}, dsls ...func()) {
+	payload(false, p, dsls...)
+}
+
+// OptionalPayload implements the action optional payload DSL. The function works identically to the
+// Payload DSL except it would set a bit in the action definition to denote that the payload is not
+// required. Examples:
+//
+//	OptionalPayload(BottlePayload)		// Request payload is described by the BottlePayload type
+//
+func OptionalPayload(p interface{}, dsls ...func()) {
+	payload(true, p, dsls...)
+}
+
+func payload(isOptional bool, p interface{}, dsls ...func()) {
 	if len(dsls) > 1 {
 		dslengine.ReportError("too many arguments given to Payload")
 		return
@@ -320,6 +334,9 @@ func Payload(p interface{}, dsls ...func()) {
 		a.Payload = &design.UserTypeDefinition{
 			AttributeDefinition: att,
 			TypeName:            fmt.Sprintf("%s%sPayload", an, rn),
+		}
+		if isOptional {
+			a.PayloadOptional = true
 		}
 	}
 }
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -71,6 +71,7 @@ func Files(path, filename string, dsls ...func()) {
 //			Required("Authorization", "X-Account")
 //		})
 //		Payload(UpdatePayload)				// Payload describes the HTTP request body (here using a type)
+//		OptionalPayload(UpdatePayload)			// You can use OptionalPayload instead of Payload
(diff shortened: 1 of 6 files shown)


raphael:
Thanks for the PR! As discussed in #286, a payload with required fields is not the same as a required payload. There may be APIs that have optional payloads but require the payload to validate when provided. We need a new DSL. I would suggest adding a `OptionalPayload` function to the `Action` DSL since it's probably the rarer case:
```go
Action("foo", func() {
    OptionalPayload(MyPayloadType)
    // ...
}
``` 
This would work identically to `Payload` except it would set a bit in the action definition to denote that the payload is not required. Code generation would use that bit to do the right thing.

tchssk:
I understood the difference between a payload with required fields and a required payload.
First I added `MissingPayloadError()` and fixed gen_app to use it.

`OptionalPayload` is a good idea.
May I try to implement that ?

raphael:
Yes please! I would start by adding a `PayloadOptional` bool field to `ActionDefinition` then have a new `OptionalPayload` DSL function that sets it to true. Then the code generator would check for this field and if false would generate code that checks that the payload is not nil. That check could be done in the generated `MountXXX` methods somewhere around https://github.com/goadesign/goa/blob/master/goagen/gen_app/writers.go#L618.

tchssk:
I took your opinion as reference and implemented `OptionalPayload` from that.
Some comments may need further improvements.

design/apidsl/action.go
```diff
@@ -274,6 +274,20 @@ func Params(dsl func()) {
 //	})
 //
 func Payload(p interface{}, dsls ...func()) {
+	payload(false, p, dsls...)
+}
+
+// OptionalPayload implements the action optional payload DSL. The function works identically to the
+// Payload DSL except it would set a bit in the action definition to denote that the payload is not
```
raphael: replace `except it would set` with `except it sets`
replace `Examples:` with `Example:`
replace 
```
// Request payload is described by the BottlePayload type
```
with
```
// Request payload is described by the BottlePayload type and is optional
```

design/apidsl/action.go
```diff
@@ -321,6 +335,9 @@ func Payload(p interface{}, dsls ...func()) {
 			AttributeDefinition: att,
 			TypeName:            fmt.Sprintf("%s%sPayload", an, rn),
 		}
+		if isOptional {
```
raphael: This could just be `a.PayloadOptional = isOptional`

design/definitions.go
```diff
@@ -236,6 +236,8 @@ type (
 		QueryParams *AttributeDefinition
 		// Payload blueprint (request body) if any
 		Payload *UserTypeDefinition
+		// PayloadOptional describes Payload is optional or not
```
raphael: Replace with `PayloadOptional is true if the request payload is optional, false otherwise.`

raphael:
Looks great, thank you! I left  a few comments in the commits. Also please add an example to the `Action` DSL function header comment that uses `OptionalPayload`, around this: https://github.com/goadesign/goa/blob/master/design/apidsl/action.go#L89
(make it clear that `Payload` and `OptionalPayload` are exclusive in the comments).
Thank you!

tchssk:
Thank you for your proofreading.
I replaced theirs and added comments to the `Action` DSL function as a new commit 796824d6242d4ad2e1c4b90d2168c3ffdcfdb65a.

If they look good, I squash the commit into 5294cc08dad21e88dc2e89dba3ab0c3d8bb93393 before merging.

raphael:
Looks great, thank you!

PR #518, opened 2016-05-31
Remove special case for JSON media type.

raphael:
When creating identifier since goa now supports arbitrary encodings.
Also properly keep media type modifiers when using `DefaultMedia`.


raphael:
Fix #431

PR #520, opened 2016-05-31
fix improperly formatted responses due to "null\n" being appended

jim-slattery-rs:
the default `NotFound` implementation can mess up the JSON response, appending `"null\n"` to an otherwise good 404 string:
So the client would receive:
```json
{"code":"not_found","status":404,"detail":"/"}
null

```
instead of:
```json
{"code":"not_found","status":404,"detail":"/"}

```

### DIFF
--- a/service.go
+++ b/service.go
@@ -113,7 +113,7 @@ func New(name string) *Service {
 		}
 		ctx := NewContext(service.Context, rw, req, params)
 		err := notFoundHandler(ctx, ContextResponse(ctx), req)
-		if err != nil {
+		if !ContextResponse(ctx).Written() {
 			service.Send(ctx, 404, err)
 		}
 	})
--- a/service.go
+++ b/service.go
@@ -113,7 +113,9 @@ func New(name string) *Service {
 		}
 		ctx := NewContext(service.Context, rw, req, params)
 		err := notFoundHandler(ctx, ContextResponse(ctx), req)
-		service.Send(ctx, 404, err)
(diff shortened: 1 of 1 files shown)


jim-slattery-rs:
Not sure if this is related to the recent change in https://github.com/goadesign/goa/pull/511

raphael:
The problem I see with the code is that it doesn't check whether a middleware has already written the response so the check
```
		if resp := ContextResponse(ctx); resp != nil && resp.Written() {
			return
		}
```
needs to be moved back inside the handler itself prior to calling `service.Send`. Can you confirm that this is indeed the problem you are seeing?

Thank you for raising the issue!

jim-slattery-rs:
@raphael Thanks, I changed it to restore the check at the end:
```go
if !ContextResponse(ctx).Written() {
	service.Send(ctx, 404, err)
}
````
And I kept the extra check at the top of the function, since that seemed like a valid optimization from https://github.com/goadesign/goa/pull/511
```go
if resp := ContextResponse(ctx); resp != nil && resp.Written() {
	return
}
```

raphael:
Makes sense, looks good. Were you able to figure out what is writing the response? is it something in your code? just want to make sure we are not masking another bug.

jim-slattery-rs:
In our case, it was the goa `ErrorHandler` already writing the 404 error response JSON.

raphael:
I see, makes sense. I guess this method should still write it if that middleware isn't used.

PR #533, opened 2016-06-07
Fix godoc typo for CONNECT func

bits01:


bketelsen:
LGTM, thanks for the PR!

PR #536, opened 2016-06-07
Allow the ability to generate no tool.

brycereitano:
I have been finding myself using quite of bit of the helpers in the client library but not the tool itself. This adds the ability to prevent the generation of the tool and cli.


raphael:
Looks good! do you think you could add a test that verifies that when the generator `notool` field is set then the tool file isn't generated?

brycereitano:
It seems tests are failing on something unrelated?

raphael:
Thanks!

Issue #544, opened 2016-06-09
CORS DSL preflight requests aren't wrapped by middleware. 


brycereitano:
Currently as it stands, the preflight requests that are generated by the CORS dsl can't be manipulated by middleware. I think this should be the case, and to prevent noise in the logs that the provided middleware ignores those requests when logging.

raphael:
PR #547 updates the generated code so that CORS preflight handlers now get properly wrapped with the corresponding controller and service middleware. I'm not sure there is much goa can or should do wrt logging. It's easy enough to wrap the log request middleware in your own that does not log if the request matches certain criteria. Doing something in goa would require a lot of additional complexity as the logger middleware gets invoked prior to the generated code. Also I think some users may actually be interested in having these logs (useful to troubleshoot CORS issues). I'm closing this for now, feel free to re-open or open another issue if you see something that can be done easily.

PR #547, opened 2016-06-11
Tweak CORS preflight handlers code generation

raphael:
So they get properly wrapped with the controller middleware.


raphael:
Fix #544

PR #549, opened 2016-06-12
Add explicit support for x-www-form-urlencoded encoding

raphael:
a.k.a. form encoding

### DIFF
--- a/encoding/form/encoding.go
+++ b/encoding/form/encoding.go
@@ -1,0 +1,23 @@
+/*
+Package form provides a "application/x-www-form-encoding" encoder and decoder.  It uses
+github.com/ajg/form for the actual implementation which can be used directly as well.  The goal of
+this package is to raise awareness of the package above and its direct compatibility with goa.
+*/
+package form
+
+import (
+	"io"
+
+	"github.com/ajg/form"
+	"github.com/goadesign/goa"
+)
+
+// NewEncoder returns a form encoder that writes to w.
+func NewEncoder(w io.Writer) goa.Encoder {
+	return form.NewEncoder(w)
+}
+
+// NewDecoder returns a form decoder that reads from r.
+func NewDecoder(r io.Reader) goa.Decoder {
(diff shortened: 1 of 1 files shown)


baseman:
Super thanks! Is there an example of this encoding for the design spec?

raphael:
There's one here: https://github.com/goadesign/examples/blob/master/form_encoding/design/design.go

baseman:
Is there an example of a single endpoint resource action. Thanks for the quick reply

raphael:
Not sure what you mean, the example above only exposes one endpoint.

baseman:
I have two actions under one resource. One accepts application/json, the other x-www-form-urlencoded

baseman:
It kosher I just needed to limit a request action payload to either json or x-www-form-urlencoded but am going to accept both (json or x-www-form-urlencoded) as this is what "Consumes" lets me do similar in the example you provided above

raphael:
Oh I understand now. Yes you need to accept both, goa will take care of picking the right decoder according to the content type header. See http://goa.design/implement/encoding/

baseman:
got it thanks again!

KerwinKoo:
@baseman  @raphael  I'm running into the same issue. Could you show me how to use `x-www-form-urlencoded` , `application/json` and `application/xml` at the same time? I insert 

```
       Consumes("application/x-www-form-urlencoded", func() {
		Package("github.com/goadesign/goa/encoding/form")
	})
	Produces("application/x-www-form-urlencoded", func() {
		Package("github.com/goadesign/goa/encoding/form")
	})
```

into `API` func, it works for `x-www-form-urlencoded` BUT `json` nor `xml`.  

I didn't get the right method from the above example in  http://goa.design/implement/encoding/,  could you show me an example to accept these content type?

Thank you !

KerwinKoo:
Sorry for bothering. I have solved this issue. I found an encoder/decoder example from `goa/encoding/doc.go`.

Issue #546, opened 2016-06-10
Error Response does not include CORS headers


brycereitano:
Successful POST
```
HTTP/1.1 200 OK
Access-Control-Allow-Credentials	true
Access-Control-Allow-Origin	*
Access-Control-Expose-Headers	Age, Cache-Control, Content-Length, Content-Type, Date, Expires, Host, Keep-Alive, Last-Modified, Location, Server, Status, Strict-Transport-Security
Access-Control-Max-Age	600
Content-Type	application/vnd.project+json; type=collection
Date	Fri, 10 Jun 2016 19:55:22 GMT
Content-Length	346
```

Failing Validation POST
```
HTTP/1.1 400 Bad Request
Content-Type: application/vnd.api.error+json
Date: Fri, 10 Jun 2016 19:55:33 GMT
Content-Length: 151
```

raphael:
I'm not able to reproduce this, can you provide a complete example? thanks!

brycereitano:
Using this design
```go
var _ = API("CORS Bug", func() {
	Origin("*", func() {
		Methods("GET", "POST", "PUT", "PATCH", "DELETE")
		Headers("Accept", "Content-Type")
		Expose("Content-Type", "Origin")
		MaxAge(600)
		Credentials()
	})
})

// User is the user resource media type
var User = MediaType("application/vnd.user+json", func() {
	Description("A User")
	Attributes(func() {
		Attribute("email", String, "email of user", func() {
			Format("email")
		})

		Required("email")
	})

	View("default", func() {
		Attribute("email")
	})
})

var _ = Resource("user", func() {
	Action("create", func() {
		Description("Create new user.")

		Routing(
			POST(""),
		)
		Payload(func() {
			Member("email")
			Required("email")
		})
		Response(Created, User)
	})
})
```

Running the main file without implementing anything.
Run first curl command with no body or valid body:
```bash
curl -v -H "Content-Type: application/json" -H "Origin: localhost" -X POST http://localhost:8080
or
curl -v -H "Content-Type: application/json" -H "Origin: localhost" -X POST -d '{email: "example@example.com"}' http://localhost:8080
```
Access Control headers are present.

Run second curl command with empty body:
```bash
curl -v -H "Content-Type: application/json" -H "Origin: localhost" -X POST -d '{}' http://localhost:8080
```
No Access Control headers are present.

raphael:
Great, thank you!

PR #550, opened 2016-06-12
Refactor how payload is unmarshaled

raphael:
Store unmarshaler error in context and handle it in generated code so
that middleware that is request specific such as CORS is properly
executed. This also simplifies the setup and makes the code easier to
evolve to the point where goa MuxHandlers do not return errors (for
compatibility with Go 1.7 HTTP handlers).


raphael:
Fix #546

PR #559, opened 2016-06-14
Add support for the generated client go code for Gopherjs 

skritchz:
Those three commits, allow building with gopherjs against the client generated go code.


raphael:
Nice! thank you. Looks like a few comments still need tweaking, from TravisCI:
```
uuid/common.go:3:1: comment on exported type UUID should be of the form "UUID ..." (with optional leading article) 
uuid/uuid.go:12:1: comment on exported function FromString should be of the form "FromString ..." uuid/uuid.go:18:1: comment on exported function NewV4 should be of the form "NewV4 ..." 
^ - Lint errors!
```

raphael:
Thank you!

PR #568, opened 2016-06-22
Fix return type in app tests

tchssk:
Test methods created by `goagen` return the result as pointer, but they should use pointer and non-pointer properly.


raphael:
Thank you!

PR #574, opened 2016-06-24
Make it possible to set service used by test helpers

raphael:
This change makes it possible to set the service used by the test helpers to run the action.
Also it removes the `WithContext` variations to instead only have one helper generated by response that always accepts a context.
The helper takes care of initializing the context if a nil value is given.
It also initialized the service to a default test service if it is nil.


raphael:
@ikawaha @aslakknutsen fyi this is a breaking change for how the tests are generated

PR #580, opened 2016-06-28
Use appropriate types for each endpoints in app tests

tchssk:
Maybe it is a regression bug of #570.


raphael:
Yes thank you, media types are objects or arrays.

PR #589, opened 2016-07-01
Remove .exe suffix from goagen output

maxandersen:
Why:

 * When generating on Windows goagen puts goagen.exe into the output
   causing different outputs on Windows vs Linux

This change addreses the need by:

 * Stripping a possible `.exe` suffix from the calculation of the
   command line. Crude, but simple


raphael:
Makes sense, thanks!

maxandersen:
just realized I hadn''t properly formatted it. push --forced an update with `go fmt` applied.

raphael:
Great, Travis would have caught it ;)

PR #590, opened 2016-07-01
Fix error merge when target metadata is nil

bits01:
Please review tests, didn't spend too much time to figure out the proper way to use Ginkgo


raphael:
Looks great, thank you!

PR #593, opened 2016-07-02
Change TBD tags to start/end block pairs

verdverm:
PR for discussion in #592 
Example: https://github.com/goadesign/examples/pull/6

change TBD to <controller>_<func> [start/end]_implement tar pairs

This enables an external tool to manage code regeneration without disruption existing modifications.

Issue #592, opened 2016-07-02
Proposal: Code regeneration while maintaining business logic


verdverm:
As a developer, I would like my workflow to use the following actions, indifferent to order.

- Edit within the design folder.
- Generate assets with Goa.
- Edit my stubs and business logic.

Proposal:

1. Use `<tag>_start` and `<tag>_end` in place of the `TBD` stubs.
1. Have the generator read the file to extract the inner-stub content
1. Add the inner-content back with the updated AST

Currently

1. The stubbed files are not regenerated
1. The stubbed files are over written

Work around:

1. Rename the current file
1. Regenerate the file
1. Copy the logic back in

(this can be automated with make,bash,sed, etc...)

raphael:
I would like to avoid opening that can of worm :) 

The scaffold files (the service `main.go`, the controller files and the client `main.go`) are generated as a way to help get started. The idea is that it's a *one-time* process. From then on these files are "yours" to maintain and update. `goagen` will *not* override these files if they exist.

This is in contrast with the files generated e.g. in the `app` package which get re-generated entirely (and overwritten) each time. The `app` package is "owned" by goa and you should not put any files in it.

This makes the design/implement cycle seamless. The only things generated by `goagen` that you can edit are the one-time generated scaffold files. Everything else is completely regenerated each time.

verdverm:
How do you feel about only changing the `TBD: implement` to:

```
    // {{ $ctrlName -}}_{{- goify .Name true }}: start_implement

    // Put your logic here

    // {{ $ctrlName -}}_{{- goify .Name true }}: start_implement
```

to facilitate an external tool or process?

raphael:
Sounds good to me.

verdverm:
I can do a PR then, I can also add my Makefile and bash script somewhere

later half sound ok? location? (docs is cool too)

raphael:
Maybe the Makefile could go to the examples repo for now? It's not the perfect place but it's better than an entirely new `tools` repo until we have more of a need for one. 

I have to admit I'm a little hesitant about adding the tags to the main just because in simple apps the generated main is sufficient. Is there a way your script could rely on the current comments?

verdverm:
How did you feel about adding a tag pair to the main file?

verdverm:
Example: https://github.com/goadesign/examples/pull/6

verdverm:
@raphael btw, are you going to be at GopherCon next week?

raphael:
[...]

PR #594, opened 2016-07-02
Tag pairs for main

verdverm:
Related to 
tag_pair issue: #592 
tag_pair PR: #593 
example: https://github.com/goadesign/examples/pull/6

This adds two pairs of tags to main for pre/post blocks.


raphael:
Thank you but as I mentioned in #592 I don't think adding these to `main` adds a lot of value. To me the trade-off is not there.

raphael:
Closing this as I don't believe all the extra comments add value (the main doesn't change save for the mounting of controllers which is easily done manually).

verdverm:
Where do you generally initial databases, read config, and such?

raphael:
What I mean is that the file `main.go` won't change much even if you change the design so it's not worth doing all that work just to get the couple of lines added by the code generator (to mount a new controller - that would be the only change).

Again to be clear: the files `main.go` and all the controller files (files generated at the top level) *are not meant to be regenerated*. The idea is to generate them once to help you get started but then these are *your* code. This is in contrast with e.g. the `app` package that gets entirely re-generated each time.

Issue #598, opened 2016-07-05
Lowercase c:\ should be uppercase in goagen for Windows


JKhawaja:
When running e.g.  `goagen bootstrap -d myapi\design` if `$GOPATH` is currently set at e.g. `C:\Projects\` on Windows box, goagen tool will return error e.g.`Go source file "c:\Projects\myapi\goagen828027691\design" not in Go workspace, adjust GOPATH C:\Projects\`. 

Changing the Windows `$GOPATH` variable to a lowercase `c:\Projects\` fixes the issue.

raphael:
Thanks for the report! this should be fixed in master. Would you mind verifying?

JKhawaja:
I still seem to be getting the same error. I cloned the master branch and ran the makefile  (which I believe generated me a new goagen binary) and uppercased the $GOPATH variable but still recieved same error as before.

raphael:
Can you try running `go get -u github.com/goadesign/goa/...` and see if you still get the error?

JKhawaja:
It seems to have "worked". But, now when running `goagen bootstrap -d cellar/design` it is returning an error: 
```
exit status 1
missing API definition, make sure design is properly initialized
```

JKhawaja:
I seem to also be getting a different error on a personal project when running `goagen bootstrap -d myproject/design`: 
```
exit status 1
c:\$GOPATH\src\myproject\app does not contain a Go package
```

raphael:
These issues seem different so I'm going to close this one. It looks like these could be vendoring issues? at least you'd get the same errors if things were not vendored properly.

JKhawaja:
Even if the second error is (possibly but unlikely) related to vendoring, the first error: https://github.com/goadesign/goa/issues/598#issuecomment-231181696
only started occurring after the updated merge. Also, the second error is only run on a repo with a design package only so I am not sure how it could be looking for some other Go package.

raphael:
OK I'm reopening the issue, will test on Windows.

raphael:
That second error is probably due to dirty files that stayed around from previous failed generation attempts. It's the Go compiler complaining that the `app`  directory is empty or does not contain Go files. Can you start again from fresh?

JKhawaja:
Tried everything fresh (including a new `go get -u github.com/goadesign/goa/...`) and still getting the same errors. :/

raphael:
[...]

PR #599, opened 2016-07-06
Fix how goagen detects that a path is under GOPATH

raphael:
So that it also works on Windows when the casing of e.g. the drive letter differs


raphael:
Fix #598

PR #605, opened 2016-07-07
Add imports in app tests

tchssk:
`goagen` writes some fixed packages as import paths in app tests.
It should also include packages for types that used in `Params` and `QueryParams`.


raphael:
Thank you, I guess I had forgotten to add the time package? The idea is to add all the *potential* packages - the code generator then takes care of removing packages that are not actually being used. Do you think you could change your PR to simply add the missing potential packages to the list? Thank you!

tchssk:
As you say, time package was missing and I misread the idea.
Excuse me, I replaced the commit :sweat_smile:

raphael:
No apologies needed, thank you for fixing all my bugs :)

PR #610, opened 2016-07-11
Add missing sql.Scanner/driver.Value interfaces to uuid.UUID

### LINKED ISSUE #609: Goa uuid.UUID fail combined with Gorma, missing sql.Scanner and driver.Value interface
After Goa 'reimplemented' UUID over using satori/go.uuid, all interfaces the satori v implemented got lost.

The Postgres SQL Driver can't store/read the Goa UUID without the Value/Scanner interfaces.

aslakknutsen:
Copy implementation from satori/go.uuid/uuid.UUID

fixes #609

A bit crude fix, but works.. Better suggestion?


raphael:
Thank you! I think we need to revisit how we approach the problem. Including all that code doesn't feel very satisfactory :). The basic problem @skritchz is trying to fix is to remove the use of `uuid.UUID` when the client code is written in gopherJS because the package uses constructs that don't work there. On the server side we should be able to just use the uuid package. So my recommendation would be to change the code generators to use the satori package directly on the server side.

aslakknutsen:
Yea, the Goa uuid.UUID was introduced to remove the usage of the satori package. 

Any tips on how to use the satori package and still make goperJS happy?

raphael:
What I'm suggesting above is to use the satori package server side (`app` package) and only use the goa uuid package client side. GopherJS is only interested in client side and the postgress stuff is only relevant to server side. So the change would be to update the `genapp` generator to import the satori package instead of the goa UUID package. This should make what you need work and shouldn't affect GopherJS.

goagen/gen_client/generator.go
```diff
@@ -269,7 +269,7 @@ func (g *Generator) generateClientResources(pkgDir, clientPkg string, funcs temp
 		codegen.SimpleImport("io"),
 		codegen.SimpleImport("net/http"),
 		codegen.SimpleImport("time"),
-		codegen.NewImport("uuid", "github.com/goadesign/goa/uuid"),
+		codegen.NewImport("uuid", "github.com/satori/go.uuid"),
```
raphael: This should still use the goa `uuid` package since it's client side.

goagen/gen_client/generator.go
```diff
@@ -371,7 +371,7 @@ func (g *Generator) generateResourceClient(pkgDir string, res *design.ResourceDe
 		codegen.SimpleImport("time"),
 		codegen.SimpleImport("golang.org/x/net/context"),
 		codegen.SimpleImport("golang.org/x/net/websocket"),
-		codegen.NewImport("uuid", "github.com/goadesign/goa/uuid"),
+		codegen.NewImport("uuid", "github.com/satori/go.uuid"),
```
raphael: same as above.

validation.go
```diff
@@ -9,7 +9,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/goadesign/goa/uuid"
+	"github.com/satori/go.uuid"
```
raphael: Since this is used by the client as well it should also use the goa `uuid` package.

aslakknutsen:
Reverted change of using satori/go.uuid in Client code.

goagen/gen_client/cli_generator.go
```diff
@@ -693,4 +732,99 @@ func hasFlag(name string) bool {
 	}
 	return false
 }
-`
+
+func jsonVal(val string) *interface{} {
+	var t interface{}
+	err := json.Unmarshal([]byte(val), t)
+	if err != nil {
+		panic(err)
```
raphael: we shouldn't panic if these fail. It means a bad input from the user not a bug.
Instead these should return an error that bubbles all the way up (understood this is going to trigger more changes).

aslakknutsen: So, bubble it all the way up to Run() ? Convert each array/type individually in Run, for so to log and return err out of run?

raphael: Yeah bubble up all the way to `Run` so that the failure is consistent with e.g. a malformed payload JSON.

goagen/gen_client/generator.go
```diff
@@ -674,8 +702,10 @@ func toString(name, target string, att *design.AttributeDefinition) string {
 			return fmt.Sprintf("%s := strconv.FormatBool(%s)", target, name)
 		case design.NumberKind:
 			return fmt.Sprintf("%s := strconv.FormatFloat(%s, 'f', -1, 64)", target, name)
-		case design.StringKind, design.DateTimeKind, design.UUIDKind:
+		case design.StringKind:
 			return fmt.Sprintf("%s := %s", target, name)
+		case design.DateTimeKind, design.UUIDKind:
+			return fmt.Sprintf("%s := %s.String()", target, strings.Replace(name, "*", "", -1)) // remove pointer if present
```
raphael: can there be a pointer? time.Time and uuid.UUID are plain values...

goagen/gen_client/cli_generator.go
```diff
@@ -693,4 +793,119 @@ func hasFlag(name string) bool {
 	}
 	return false
 }
-`
+
+func jsonVal(val string) (*interface{}, error) {
+	var t interface{}
+	err := json.Unmarshal([]byte(val), t)
```
raphael: shouldn't that be `err := json.Unmarshal([]byte(val). &t)`?

goagen/gen_client/cli_generator.go
```diff
@@ -693,4 +793,119 @@ func hasFlag(name string) bool {
...
+		val, err := jsonVal(id)
+		if err != nil {
+			return nil, err
+		}
+		vals = append(vals, val)
+	}
+	return vals, nil
+}
+
+func timeVal(val string) (*time.Time, error) {
```
raphael: I don't think this should be returning a pointer

goagen/gen_client/cli_generator.go
```diff
@@ -693,4 +793,119 @@ func hasFlag(name string) bool {
...
+		val, err := timeVal(id)
+		if err != nil {
+			return nil, err
+		}
+		vals = append(vals, *val)
+	}
+	return vals, nil
+}
+
+func uuidVal(val string) (*uuid.UUID, error) {
```
raphael: Same comment as `timeVal` - I don't think this should return a pointer.

goagen/gen_client/generator.go
aslakknutsen: the 'name' here is the ValueName from paramData. ValueName will have a pointer prefix to VarName if Param is Required.

raphael:
Awesome PR, thank you!

aslakknutsen:
You're welcome! :)

Issue #555, opened 2016-06-13
Help for -o flag should not list the value of current directory


bits01:
`-o, --out string      output directory (default "/Users/x/go/src/github.com/goadesign/goa")`

It implies that is the default value no matter what directory the tool is run from. Instead it's more helpful to show it as:

`-o, --out string      output directory (default ".")`

bits01:
Fixed by PR #611

PR #611, opened 2016-07-12
Issue #555 fix help for -o flag

bits01:
Fix for issue #555


raphael:
Thank you! doesn't this also change the value that the generators get from an absolute path to `.` ? the generators should be getting an absolute path so if that's the case something should convert the `.` before the value is actually used.

bits01:
It probably does. But does it matter? I can make the change if it matters.

bits01:
Take 2 ... attempt to convert to absolute path. Please review. Thank you.

raphael:
Looks great, thank you!

Issue #516, opened 2016-05-31
Generated swagger docs/example not honoring Links() in MediaType


jim-slattery-rs:
With a MediaType using `Links()`, e.g.:
```go
var SomeType = MediaType("application/vnd.sometype", func() {
	Attributes(func() {
		Attribute("subject", SubjectRefMediaType)
		Required("subject")
	})

	Links(func() {
		Link("subject")
	})

	View("default", func() {
		Attribute("links")
	})
})
```

The generated swagger docs seem to be ignoring the Link, and indicating the structure looks like this example:
```json
{
	"subject": {
		"href": "/something"
	}
}
```

whereas the actual accepted structure must look like this:
```go
{
	"links": {
		"subject": {
			"href": "/something",
		}
	}
}
```

bits01:
Same is true for the generated client side data types, they are missing the Links fields.
This can be seen in the current goa-cellar example:

Server side Bottle:
```
type Bottle struct {
	// Account that owns bottle
	Account *Account `json:"account,omitempty" xml:"account,omitempty" form:"account,omitempty"`
	// API href of bottle
	Href string `json:"href" xml:"href" form:"href"`
	// ID of bottle
	ID int `json:"id" xml:"id" form:"id"`
	// Links to related resources
	Links *BottleLinks `json:"links,omitempty" xml:"links,omitempty" form:"links,omitempty"`
	Name  string       `json:"name" xml:"name" form:"name"`
	// Rating of bottle between 1 and 5
	Rating   *int   `json:"rating,omitempty" xml:"rating,omitempty" form:"rating,omitempty"`
	Varietal string `json:"varietal" xml:"varietal" form:"varietal"`
	Vineyard string `json:"vineyard" xml:"vineyard" form:"vineyard"`
	Vintage  int    `json:"vintage" xml:"vintage" form:"vintage"`
}
```

Client side Bottle:
```
type Bottle struct {
	// Account that owns bottle
	Account *Account `json:"account,omitempty" xml:"account,omitempty" form:"account,omitempty"`
	Color   string   `json:"color" xml:"color" form:"color"`
	Country *string  `json:"country,omitempty" xml:"country,omitempty" form:"country,omitempty"`
	// Date of creation
	CreatedAt time.Time `json:"created_at" xml:"created_at" form:"created_at"`
	// API href of bottle
	Href string `json:"href" xml:"href" form:"href"`
	// ID of bottle
	ID   int    `json:"id" xml:"id" form:"id"`
[...]

PR #616, opened 2016-07-15
Fix swagger generation of media type links.

raphael:
This PR fixes the Swagger generation so that the media types get properly projected prior to being rendered in the schema. Currently the code uses the "default" view. It may make sense to introduce metadata to control that behavior in the future.


raphael:
Fix #516

PR #620, opened 2016-07-18
Make sure projected media types define a default view

raphael:
So that code generators and finalizers (e.g. example finalizer) can rely on it.


raphael:
Fix #516

PR #621, opened 2016-07-18
request_id_length_limit

smessier:
added a default length limit of 128 for incoming X-Request-Id header values for safety
the issue is that request IDs will appear frequently in logs and the server could be impacted if a bad actor sends very long request IDs

### DIFF
--- a/middleware/request_id.go
+++ b/middleware/request_id.go
@@ -13,8 +13,13 @@ import (
 	"golang.org/x/net/context"
 )
 
-// RequestIDHeader is the name of the header used to transmit the request ID.
-const RequestIDHeader = "X-Request-Id"
+const (
+	// RequestIDHeader is the name of the header used to transmit the request ID.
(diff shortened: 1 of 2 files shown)


middleware/request_id.go
```diff
@@ -13,8 +13,13 @@ import (
 	"golang.org/x/net/context"
 )
 
-// RequestIDHeader is the name of the header used to transmit the request ID.
-const RequestIDHeader = "X-Request-Id"
+const (
+	// RequestIDHeader is the name of the header used to transmit the request ID.
+	RequestIDHeader = "X-Request-Id"
+
+	// DefaultRequestIDLengthLimit is the default maximum length for header value.
```
raphael: Seems like you're missing a "the" here - maybe specify that this only applies to the request ID header too (should be obvious given the context but I guess it's better to be explicit): "for the request ID header value"

raphael:
Thanks, looks great! just a tiny comment inline.

PR #623, opened 2016-07-19
Fix a comment in app tests

tchssk:
`gen_app` generates following comment.

```go
// CreateBottleCreated Create runs the method Create of the given controller with the given parameters and payload.
```

Maybe second word `Create` is unneeded.

```go
// CreateBottleCreated runs the method Create of the given controller with the given parameters and payload.
```

https://github.com/goadesign/goa-cellar/blob/f923c55476595a95fa4f97b03eb33b084fd9e58a/app/test/bottle_testing.go#L18


raphael:
Thanks!

PR #626, opened 2016-07-20
Added missing import for step 3 of example

chrispassas:
The strconv import did not exist. Adding it to the readme to make the walkthrough more clear.


raphael:
Thank you!

PR #627, opened 2016-07-20
Make error classes return Go's error interface

raphael:
instead of a goa specific struct which introduces package coupling.
Errors created via error classes implement the goa ServiceError interface
which provides methods for retrieving the response status and unique
error instance ID. Middlewares can check whether an error exposes these
functions to determine whether the error corresponds to a error created
via error classes or whether the error is a raw Go error.


raphael:
This PR includes *BREAKING CHANGES*. Specifically:

* The `goa.Error` struct is gone, replaced with `error`.
* The media type identifier for error responses has changed from `vnd.api.error+json` to `vnd.goa.error`
* The semantic of the error generating functions created via ErrorClass has changed, see below.

While the signature of functions returned by calling `ErrorClass` functions has not changed the semantic of the parameters is different. Scan your source for the use of `NewErrorClass` and make sure to update all calls to the returned function.

Before error class functions took a format and values as argument, for example `goa.ErrBadRequest("Something bad happened: %s", err)`.

After error class functions take key pair values as optional additional arguments, for example:
`goa.ErrBadRequest("Something bad happened", "error", err)`.

error.go
```diff
@@ -63,149 +71,180 @@ var (
 )
 
 type (
-	// Error contains the details of a error response.
-	Error struct {
+	// ErrorClass is an error generating function.
+	// It accepts a message and optional key value pairs and produces errors that implement
+	// ServiceError.
+	// If the message is a string or a Stringer then the string value is used.
```
MaerF0x0: `or a Stringer then` --> `or a fmt.Stringer` for clarity ?

error.go
```diff
@@ -63,149 +71,180 @@ var (
...
 		// Code identifies the class of errors.
 		Code string `json:"code" xml:"code" form:"code"`
 		// Status is the HTTP status code used by responses that cary the error.
 		Status int `json:"status" xml:"status" form:"status"`
 		// Detail describes the specific error occurrence.
 		Detail string `json:"detail" xml:"detail" form:"detail"`
-		// MetaValues contains additional key/value pairs useful to clients.
-		MetaValues map[string]interface{} `json:"meta,omitempty" xml:"meta,omitempty" form:"meta,omitempty"`
+		// Meta contains additional key/value pairs useful to clients.
+		Meta []map[string]interface{} `json:"meta,omitempty" xml:"meta,omitempty" form:"meta,omitempty"`
```
MaerF0x0: this design choice surprises me, why make it a slice of kvs instead of a single map?

raphael: So that the order is kept when writing the response.

PR #635, opened 2016-07-24
BREAKING - Remove BaseParams , use Params instead

raphael:
This commit simplifies the DSL by getting rid of BaseParams. BaseParams is now replaced with Params. Both BaseParams and Params used to exist for resources which was confusing. The initial intent for BaseParams was to provide a way to define the BasePath parameters but that can be accomplished by Params as well. One advantage is that Traits can simply use Params now to define both base path params and e.g. global query string params.

There can now also be multiple Params declarations and when that's the case the final list of params is computed by merging the declarations (last one wins if there are name conflicts). This makes it possible to define params in traits and others directly for example.


raphael:
This PR contains **BREAKING CHANGES**. It removes the `BaseParams` DSL entirely in favor of `Params`. The semantic is the same though so a simple search and replace of `BaseParams` with `Params` in the DSL should suffice.

Issue #615, opened 2016-07-14
Generated client side data types missing Links attribute


bits01:
Splitting this from #516 as it seems to be a different issue.

Best example to see this in is `goa-cellar`: client side `Bottle` is missing the `Links *BottleLinks` attribute.

From a conversation with Raphael on Slack, so we don't lose the info:

this is where the types are generated: https://github.com/goadesign/goa/blob/master/goagen/gen_client/generator.go#L280-L289
so I think what we want to do is check whether the type is a MT and not just a UT
if it's a MT then we need to call `Project` on each view it defines
and generate the corresponding structs
Each struct name would have the view name appended except if the view name is "default"

raphael:
This is fixed in master.

bits01:
Thank for the fix, I will try it out. This is huge help.

PR #638, opened 2016-07-24
genclient: generate all projected media types

raphael:
Update client generator to generate all projected media types. The media types and user types are now generated in `media_types.go` and `user_types.go` respectively.


raphael:
Fix #615

PR #641, opened 2016-07-26
Copy original error ID in the error handler if not verbose

bits01:
Copy original error ID in the error handler if not verbose. This ensures that the logged error ID matches what the client receives.


raphael:
Thanks!

Issue #622, opened 2016-07-19
Generated media_types not using specified View


Tri125:
Tested on c2765b82364bf0bf7482dfb18a185e78e45255bd

If say a mediaType named Book links to another media named CoverArt and inside the View definition of Book you specify a different view than the default view (ie: a View named "Link") for the CoverArt attribute, the generated mediaType code for the Book type will have an attribute of type *Coverart where type *CovertartLink is expected.

Minimal api definition to reproduce:

```
package design

import (
	. "github.com/goadesign/goa/design"
	. "github.com/goadesign/goa/design/apidsl"
)

var _ = API("api", func() {
	BasePath("/")
	Host("localhost:8080")
})

var Book = MediaType("application/vnd.book+json", func() {
	Attributes(func() {
		Attribute("ISBN", String)
		Attribute("href", String)
		Attribute("name", String)
		Attribute("cover", CoverArt)

		Required("ISBN", "href", "name", "cover")

		Links(func() {
			Link("cover")
		})
	})

	View("default", func() {
		Attribute("ISBN")
		Attribute("href")
		Attribute("name")
		Attribute("cover", func() {
			View("link")
		})
	})
})

var CoverArt = MediaType("application/vnd.coverArt+json", func() {
	Attributes(func() {
		Attribute("id", String)
		Attribute("href", String)
		Attribute("artist", String)

		Required("id", "href", "artist")
	})

	View("default", func() {
		Attribute("id")
		Attribute("href")
		Attribute("artist")
	})

	View("link", func() {
		Attribute("artist")
		Attribute("href")
	})
})

var _ = Resource("book", func() {

	DefaultMedia(Book)
	BasePath("/books")

	Action("show", func() {
		Routing(GET("/:bookISBN"))
		Params(func() {
			Param("bookISBN", String)
		})
		Response(OK)
		Response(NotFound)

	})
})



```

raphael:
The issue was that the `View` function was only supported on media type attributes and not view attributes. PR #642 adds the ability to specify views on view attribute.

raphael:
Can you verify and close the ticket if it's indeed fixed? thanks!

Tri125:
The issue has been fixed. Thank you.

Tri125:
I guess I closed too soon. I don't generally check the swagger, but there is a little issue with the $ref:  field name.

From the minimal api definition that I've included, there will be an error in swagger.yaml:

```
"Reference could not be resolved: #/definitions/CoverartLinkLink"

[...]

PR #642, opened 2016-07-26
Use view attribute views

raphael:
for example:

```go
View("default", func() {
    Attribute("related_mt", func() {
        View("tiny")
    })
})
```

Prior to this change only views specified on media type attributes were
taken into account.


raphael:
Fix #622

PR #643, opened 2016-07-26
Sort default attribute tags in alphabetical ordert

tchssk:
`gen_app` generates struct tags for attributes in alpahbetical order when some `Metadata()` are specified, but it generates that in following order by default.

1. json
2. xml
3. form

Maybe they should be the same order.


raphael:
Makes a lot of sense, thanks!

tchssk:
Thanks for your speedy merging!

Issue #644, opened 2016-07-26
panic: assignment to entry in nil map


MaerF0x0:
```
$ goagen bootstrap  -d github.com/leftronic/api_v/design

exit status 2
panic: assignment to entry in nil map

goroutine 1 [running]:
panic(0x417160, 0xc8201a55f0)
        /usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:481 +0x3e6
github.com/goadesign/goa/design.(*ActionDefinition).mergeResponses(0xc8201bc140)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:1496 +0x15a
github.com/goadesign/goa/design.(*ActionDefinition).Finalize(0xc8201bc140)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:1429 +0xaa
github.com/goadesign/goa/design.(*ResourceDefinition).Finalize.func2(0xc8201bc140, 0x0, 0x0)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:817 +0x2d
github.com/goadesign/goa/design.(*ResourceDefinition).IterateActions(0xc820178c30, 0x690ae8, 0x0, 0x0)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:673 +0x265
github.com/goadesign/goa/design.(*ResourceDefinition).Finalize(0xc820178c30)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:819 +0x47
github.com/goadesign/goa/dslengine.finalizeSet(0xc8201d0ac0, 0x2, 0x2, 0x0, 0x0)
        /Users/mike.graf/go/src/github.com/goadesign/goa/dslengine/runner.go:328 +0x102
github.com/goadesign/goa/design.(*APIDefinition).IterateSets(0xc820146b00, 0x690b18)
        /Users/mike.graf/go/src/github.com/goadesign/goa/design/definitions.go:489 +0x546
github.com/goadesign/goa/dslengine.Run(0x0, 0x0)
        /Users/mike.graf/go/src/github.com/goadesign/goa/dslengine/runner.go:114 +0x45c
main.main()
        /Users/mike.graf/go/src/github.com/leftronic/api_v3/goagen159760158/main.go:26 +0x7b
```

LMK what else would be of use.

**Update**:  I have distilled it down to adding calls to "Response" in my Resource definition. 

```go
package design

import (
[...]

PR #645, opened 2016-07-26
Fix potential panic when finalizing action definitions

raphael:
If the action did not define responses but the parent resource did.


raphael:
Fix #644

PR #647, opened 2016-07-26
Request ID is set in ctx and in req header

bits01:
* Introduce utility funcs for getting/setting request ID in Context
* Set X-Request-Id header on the request from the context (at some point this should be done in a client middleware)


bits01:
Please review new version. Request header set iff present in the context. Note that a newly generated req ID will also create a new context with it, I assume there are no objections to that part, it seems to be the correct thing to do. Thank you.

raphael:
Looks great, thank you!

Issue #646, opened 2016-07-26
client: Doer.Do() should take a Context


bits01:
If the user provides a custom Doer implementation, there's no way to get to the Context.

This will be a minor breaking change and there should be an adapter function that can take an http.Client and return a new Doer so that existing code can be easily fixed.

bits01:
I can work on a PR if the suggestion is acceptable.

raphael:
Makes sense, one thing to consider is that with Go 1.7 the context is inside the request object already. For now we can add it as an extra argument but once 1.7 is out we may want to have a specific implementation for it.

bits01:
How will the Client work with Go 1.7? The current signature of the `Do` method is:
`func (c *Client) Do(ctx context.Context, req *http.Request) (*http.Response, error)`
In Go 1.7 req will have a Context but also one will be passed to the Do method. They may or may not be the same one, so we may have competing contexts in the Do method. I think this should all be revisited for Go 1.7, not sure what the plan is for goa, perhaps have a goa 1.7 release to match Go 1.7 :-)

raphael:
For now what makes sense is to keep using the context argument and basically ignore the one in the request. But yes at some point it would be good to merge the two.

bits01:
Agreed. Proposal:
- for pre-go1.7: let's use a Context on Doer.Do() too (with a helper func to ease the breaking change)
- for go1.7+: remove the Context from both Client.Do() and Doer.Do() in favor of the one in the req

bits01:
PR #648

PR #648, opened 2016-07-27
Add Context param to Doer.Do

bits01:
This PR is for #646

BREAKING CHANGE - should be announced on #goa
- client.Doer.Do() takes a Context parameter
- existing callers that pass nil work as before
- existing callers that pass an http.Client instance need to wrap it using client.HTTPClientDoer
- existing callers that pass a custom Doer need to update the Do method in their impls to take the Context parameter


raphael:
Thank you!

PR #650, opened 2016-07-27
Change projected media types so they define the "default" view.

raphael:
That's instead of the view used to project them so that projecting a
projected media type is idempotent (type name remains the same).


raphael:
Fix #622

PR #653, opened 2016-07-28
Fix path prefix check.

egonelbre:
Windows paths are case-insensitive.

goagen bootstrap failed because GOPATH was using different casing.

### DIFF
--- a/goagen/codegen/workspace.go
+++ b/goagen/codegen/workspace.go
@@ -310,7 +310,7 @@ func PackagePath(path string) (string, error) {
 		if gp, err := filepath.Abs(gopath); err == nil {
 			gopath = gp
 		}
-		if strings.HasPrefix(absPath, gopath) {
+		if filepath.HasPrefix(absPath, gopath) {
 			base := filepath.FromSlash(gopath + "/src")
 			rel, err := filepath.Rel(base, absPath)
 			return filepath.ToSlash(rel), err


goagen/codegen/workspace.go
```diff
@@ -310,7 +310,7 @@ func PackagePath(path string) (string, error) {
 		if gp, err := filepath.Abs(gopath); err == nil {
 			gopath = gp
 		}
-		if strings.HasPrefix(absPath, gopath) {
+		if filepath.HasPrefix(absPath, gopath) {
```
raphael: I don't think this does what you think. See the [implementation](https://golang.org/src/path/filepath/path_unix.go?s=606:643#L13)
Also the [comment](https://golang.org/pkg/path/filepath/#HasPrefix) of the method indicates it should not be used.

raphael:
Thanks! but the fix doesn't look right. See my comments inline.

goagen/codegen/workspace.go
raphael: Hmm I was looking at the wrong implementation, the [Windows version](https://golang.org/src/path/filepath/path_windows.go) does do the right thing. Still the comment indicates this should not be used. Not sure what the alternative is though.

raphael: This is what I've done in other places for this:
https://github.com/egonelbre/goa/blob/44b6500703786f02032da6c00b5725aa1a9cb0e1/goagen/codegen/workspace.go#L102-L103
Maybe there's a better way...

egonelbre: From blame where the note was introduced:

    1. It does not take path boundaries into account.
    2. It assumes that Windows==case-insensitive file system
    and non-Windows==case-sensitive file system, neither of
    which is always true.
    3. Comparing ToLower against ToLower is not a correct
    implementation of a case-insensitive string comparison.
    4. If it returns true on Windows you still don't know how long
    the matching prefix is in bytes, so you can't compute what
    the suffix is.

For `Abs` you have similarly that: "The absolute path name for a given file is not guaranteed to be unique." And, I don't know any Windows API calls that could be used instead either...

From http://www.gamedev.net/topic/628734-cross-platform-filepath-naming-conventions/, 

> Essentially, path comparisons will always be somewhat fragile.

There might be a better way of opening the folders at the same root when the casing doesn't match check whether they are the same file... however I'm not sure whether it will work on networked devices...

Basically, `filepath.HasPrefix` is the devil I know... and I'm not quite sure what the "better way" would be, but it sounds like a very deep rabbit-hole.

raphael: OK thank you for the explanation!

Issue #651, opened 2016-07-28
add declared headers to goa.RequestData.Params


tve:
Q: `RequestData.Params` contains params from the path and query string but not headers. Why? If I define `Headers(func(), Header("My-Header", String))` it acts like just another param in that my action's context contains a `MyHeader` field. I believe that if I in addition define a query string param called my_header then it shares the same context field. So why are Header params not in `RequestData.Params`?

Raphael Simon: let me throw back the question at you: why do you expect headers to be in the Params? I think most people would not expect it.

Thorsten von Eicken:I'm not saying all headers just the ones I've defined to be params to my actions.
For debugging I use a middleware that prints the params to the action about to be executed and I want to also print the headers consumed by the action as params, but I can't easily (or at all?) get to this list.

Another situation: suppose I have a Api-Version header and an api_version query string param. They both end up in ctx.ApiVersion, right? Which one wins if both are specified? And what does ctx.Params contain in that case?

Raphael Simon:
you can differentiate them by setting a metadata "struct:field:name" on one of them so the computed name doesn't clash. Otherwise you get one. path param / query string wins over header if both are present and not disambiguated via metadata.

Raphael Simon:
I don't see a lot of drawback to doing it, my only concern is that everybody is going to be surprised by it (until they think it through and realize it makes sense)

yes that could be a v2 thing :slightly_smiling_face:

ok do you mind opening an issue? it's nice because it's a non breaking change (if not renaming). There will have to be a nice comment for the `Params` field explaining that it also contains the values of headers defined in the design.

raphael:
#654 and #655 implement this. The request data object `Params` field now includes the header values for the header parameters defined in the design keyed by (param) name.

PR #654, opened 2016-07-28
Set header values defined in design in request data Params map.

raphael:
This makes it possible to iterate over all values defined in the design as being "parameters"
to the action including path parameters, query string parameters and now headers.
This PR also removes the previous behavior or splitting query params on commas to build slices.
Instead params must be passed in as actual arrays i.e. foo=bar&foo=baz. This makes it
possible to properly handle query strings such as foo=bar&foo=baz,abc which now result in the
Foo field being initialized with []string{"bar","baz,abc"} instead of
[]string{"bar","baz","abc"} as was done previously.


raphael:
Fix #651

Issue #659, opened 2016-07-29
gen_swagger omits parameter validations of the zero value


ikawaha:
gen_swagger omits parameter validations of the zero value. 
Following parameter validations are omitted in swagger.(json|yaml):

* Minimum(0)
* Maximum(0)
* MinLength(0)
* MaxLength(0)
* MinItems(0)
* MaxItems(0)

## example design

bottleID: Minimum(0), Maximum(0), Default(0)

```
package design // The convention consists of naming the design
// package "design"
import (
        . "github.com/goadesign/goa/design" // Use . imports to enable the DSL
        . "github.com/goadesign/goa/design/apidsl"
)

var _ = API("cellar", func() { // API defines the microservice endpoint and
        Title("The virtual wine cellar") // other global properties. There should be one
        Version("")
        Description("A simple goa service") // and exactly one API definition appearing in
        Scheme("http")                      // the design.
        Host("localhost:8080")
})

var _ = Resource("bottle", func() { // Resources group related API endpoints
        BasePath("/bottles")      // together. They map to REST resources for REST
        DefaultMedia(BottleMedia) // services.

        Action("show", func() { // Actions define a single API endpoint together
                Description("Get bottle by id") // with its path, parameters (both path
                Routing(GET("/:bottleID"))      // parameters and querystring values) and payload
                Params(func() {                 // (shape of the request body).
                        Param("bottleID", Integer, "Bottle ID", func() {
                                Minimum(0) // !!!!
                                Maximum(0) // !!!!
                                Default(0)
                        })
                })
                Response(OK)       // Responses define the shape and status code
                Response(NotFound) // of HTTP responses.
        })
})

// BottleMedia defines the media type used to render bottles.
var BottleMedia = MediaType("application/vnd.goa.example.bottle+json", func() {
        Description("A bottle of wine")
        Attributes(func() { // Attributes define the media type shape.
                Attribute("id", Integer, "Unique bottle ID")
                Attribute("href", String, "API href for making requests on the bottle")
                Attribute("name", String, "Name of wine")
                Required("id", "href", "name")
        })
        View("default", func() { // View defines a rendering of the media type.
                Attribute("id")   // Media types may have multiple views and must
                Attribute("href") // have a "default" view.
                Attribute("name")
        })
})
```

## gen_swagger

result: 
```
paths:
  /bottles/{bottleID}:
    get:
      description: Get bottle by id
      operationId: bottle#show
      parameters:
      - default: 0
        description: Bottle ID
[...]

PR #660, opened 2016-07-29
Fix the handling of the zero value validations

ikawaha:
see #659


raphael:
Thank you! do you think you could add a test that would have failed before the fix?

raphael:
Perfect, thank you!

ikawaha:
Thanks for merging!

Issue #557, opened 2016-06-14
SupportedValidationFormats: support IP validation (as in IPv4 or IPv6)


bits01:
Currently an attribute can accept only one of `Format("ipv4")` or `Format("ipv6")`. There is no way to specify both IPv4 and IPv6 as accepted formats. An `ip` format that would allow either could work.

As a more general solution, the JSON schema `oneOf` could be useful, perhaps as a `OneOf` DSL, though I'm not sure how hard it would be to implement.

bits01:
PR #664

raphael:
Thanks for the PR!

PR #664, opened 2016-07-30
Add format IP validation (either IPv4 or IPv6)

bits01:
PR for #557
Add format IP validation (either IPv4 or IPv6)


bits01:
Ready for review. Could not find any references in goa.design to IP validation.

raphael:
This is great, thank you!

PR #670, opened 2016-08-01
Add (Un)Marshal Text/Binary methods to support json parsing

aslakknutsen:
This is required by Client to parse Payload with UUID in it.

### DIFF
--- a/encoding/json/encoding_suite_test.go
+++ b/encoding/json/encoding_suite_test.go
@@ -1,0 +1,13 @@
+package json_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestMiddleware(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Json Encoding Suite")
+}
--- a/encoding/json/encoding_suite_test.go
+++ b/encoding/json/encoding_suite_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 )
 
-func TestMiddleware(t *testing.T) {
+func TestJsonEncoding(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Json Encoding Suite")
 }
--- a/encoding/json/encoding_test.go
+++ b/encoding/json/encoding_test.go
@@ -1,0 +1,48 @@
+package json_test
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"github.com/goadesign/goa/uuid"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("JsonEncoding", func() {
+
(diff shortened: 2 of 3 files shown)


aslakknutsen:
No idea how this effect the GopherJS client.

encoding/json/encoding_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package json_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestMiddleware(t *testing.T) {
```
raphael: I guess this should be `TestEncoding` :)

encoding/json/encoding_test.go
```diff
@@ -0,0 +1,48 @@
...
+			encoder.Encode(data)
+			encoded = b.String()
+
+			Ω(encoded).Should(ContainSubstring(data.ID.String()))
+			Ω(encoded).Should(ContainSubstring(data.Name))
+		})
+
+		It("decode", func() {
+			var b bytes.Buffer
+			b.WriteString(encoded)
```
raphael: you need to initialize encoded somewhere...

encoding/json/encoding_suite_test.go
aslakknutsen: *cough* yes... :)

uuid/uuid.go
```diff
@@ -18,11 +18,37 @@ func NewV4() UUID {
...
+}
+
+// MarshalBinary Wrapper over the real MarshalBinary method
+func (u UUID) MarshalBinary() ([]byte, error) {
+	return uuid.UUID(u).MarshalBinary()
+}
+
+// UnmarshalBinary Wrapper over the real UnmarshalBinary method
+func (u *UUID) UnmarshalBinary(data []byte) error {
+	t := uuid.UUID{}
```
raphael: It seems strange that the target is a pointer to UUID but that the code then uses a non-pointer `uuid.UUID` to perform the call, is that intended?

aslakknutsen: It's trying to use the satori/go.uuid to do the actual work, then 'copy' that work onto the target.. ? 

Not sure I'm following the question...

raphael: I would have expected the functions to be
```go
func (u UUID) UnmarshalBinary(...
```
(No pointer on the target)

aslakknutsen: Without it u[i] = b doesn't do anything from the caller perspective?

raphael: Then how does `t.UnmarshalBinary` work? shouldn't it be `t := &uuid.UUID{}`

aslakknutsen: t := uuid.UUID{} is the satori/go.uuid. It too use a pointer target on UnmarshalBinary

raphael: You're right, should have read the satori implementation more carefully, thank you!

PR #671, opened 2016-08-01
WIP: Allow Payload to be read from cli pipe

aslakknutsen:
usage:
cat payload | ./types-cli create prism

Maybe not the correct location to add the support, but I couldn't find anything in Cobra about direct support.

In PreRunE we check if Payload != "" and if it is, see if we can read it in from a NamedPipe. If read we set cmd.Payload to the read value and continue like nothing happened :)


raphael:
The strategy looks good to me! We just need to make sure it doesn't have averse effects on Windows. Also needs tests and docs obviously. Thank you!

raphael:
@aslakknutsen ping :) are you still looking into this?

aslakknutsen:
Well....

I was looking for someone with a windows box to verify if this worked in dos as well... and then...  got distracted. :)

*squirrel*

raphael:
Going to close this for now.

Issue #672, opened 2016-08-01
NoExample on API causes panic with ArrayOf


bits01:
Design file:

[design.txt](https://github.com/goadesign/goa/files/394582/design.txt)

raphael:
This is fixed, thanks!

PR #673, opened 2016-08-02
Improve example generation.

raphael:
Properly handle recursive cycle and unify generation of random values
with finalization of attributes.


raphael:
Fix #672

PR #675, opened 2016-08-02
Fix gen_swagger not to omit payload validations of the zero value

ikawaha:
Sorry, Issue #659 was incomplete.
The same thing happens in the payload.


raphael:
Thank you, good catch!

PR #678, opened 2016-08-03
Add (Un)Marshal Text/Binary methods to support json parsing (#670)

KerwinKoo:
* Add (Un)Marshal Text/Binary methods to support json parsing

* Add tests and fix impl ;)

* Add local test setup. Add round trip test


raphael:
Whoops. thanks!

Issue #677, opened 2016-08-03
Duplicate types declared when type is prefixed with action name


faxal:
I wanted to use separate payload types for update and create actions, so I named them accordingly.

Following fails to build with duplicate type declaration errors

```
package design

import (
	. "github.com/goadesign/goa/design"
	. "github.com/goadesign/goa/design/apidsl"
)

var CreateTPayload = Type("CreateTPayload", func() {
	Description("A type")
	Attribute("id", String, "id")
})

var T = MediaType("application/vnd.t+json", func() {
	TypeName("T")
	Reference(CreateTPayload)
	Attributes(func() {
		Attribute("id")
	})

	View("default", func() {
		Attribute("id", String)
	})
})

var _ = Resource("t", func() {
	BasePath("/ts")
	DefaultMedia(T)

	Action("create", func() {
		Description("Create new T")
		Routing(
			POST(""),
		)

		Payload(CreateTPayload)
		Response(Created)
	})

})
```

raphael:
Fixed by #681 thanks!

PR #681, opened 2016-08-03
Fix potential duplicate generation of payload types.

raphael:
This commit does two things:

1. It makes sure that action definitions that refer to user types to
   define the payload only duplicates the definition if needed (i.e. if
   the payload definition overrides some of the type attributes).

2. It makes sure that a payload type is only generated if there is not
   a user type with the same type name.


raphael:
Fix #677

Issue #676, opened 2016-08-02
Generated client incorrectly encodes arrays in query string


tve:
The generated client produces an array encoding for the query string using ',' instead of a repetition of the param. E.g., `ids=1,2,3` instead of `ids=1&ids=2&ids=3`.

Also, the test whether a query string param is necessary should use `if len(array) > 0` instead of `if array != nil`.

tve:
Looks like the test may not matter as much, I think you need to produce:
```
if len(databags) > 0 {
  for _, e := range databags {
    values.Add("databags", e)
  }
}
```

raphael:
This should be fixed, thank you!

PR #684, opened 2016-08-04
Properly handle array params

raphael:
Add each param to the url query string values instead of building comma separated string.


raphael:
Fix #676

Issue #668, opened 2016-08-01
Panics when using a MediaType definition inside a Type definition


faxal:
```
exit status 2
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0x8deba0, 0xc82000e110)
	/home/fazal/Packages/go15/src/runtime/panic.go:464 +0x3e6
github.com/goadesign/goa/design.(*MediaTypeDefinition).DefaultView(0xc8201c2680, 0x0, 0x0)
	/home/fazal/.go/src/github.com/goadesign/goa/design/types.go:773 +0x27c
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExampleRecursive(0xc8201d6b80, 0xc8201d84a0, 0xc8201aaf50, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0)
	/home/fazal/.go/src/github.com/goadesign/goa/design/definitions.go:1025 +0x5f6
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExampleRecursive(0xc8201d6900, 0xc8201d84a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/fazal/.go/src/github.com/goadesign/goa/design/definitions.go:1047 +0x948
github.com/goadesign/goa/design.(*AttributeDefinition).finalizeExample(0xc8201d6900, 0x0, 0x0)
	/home/fazal/.go/src/github.com/goadesign/goa/design/definitions.go:981 +0x58
github.com/goadesign/goa/design.(*UserTypeDefinition).Finalize(0xc8200bff40)
	/home/fazal/.go/src/github.com/goadesign/goa/design/types.go:670 +0x51
github.com/goadesign/goa/design.(*MediaTypeDefinition).Finalize(0xc8201c2600)
	/home/fazal/.go/src/github.com/goadesign/goa/design/types.go:712 +0x55
github.com/goadesign/goa/dslengine.finalizeSet(0xc8201d8440, 0x2, 0x2, 0x0, 0x0)
	/home/fazal/.go/src/github.com/goadesign/goa/dslengine/runner.go:328 +0x102
github.com/goadesign/goa/design.(*APIDefinition).IterateSets(0xc820196420, 0xa87220)
	/home/fazal/.go/src/github.com/goadesign/goa/design/definitions.go:474 +0x290
github.com/goadesign/goa/dslengine.Run(0x0, 0x0)
[...]

PR #690, opened 2016-08-05
Do not copy attribute type in DupAtt

raphael:
This is so that if the type is a media type or user type then it's OK
if the DSL hasn't run yet. The DSL will run later initializing the
referred type. If we dupped then the dupped type DSL would never get
run and the media type would not have its views initialized for example.


raphael:
Fix #668

Issue #695, opened 2016-08-05
Make ErrorMediaIdentifier a variable


raphael:
Rather than a constant so that it's possible to override its value.

PR #697, opened 2016-08-05
Make ErrorMediaIdentifier a variable instead of a constant

raphael:
So that it's possible to override it, e.g. for clients that can only handle
Content-Type headers of the form 'application/json'


raphael:
Fix #695

Issue #665, opened 2016-07-31
Examples for links do not respect MT link view


bits01:
See `goa-cellar` generated swagger files. The `Account` MT `link` view includes only `id` and `href` but the generated examples show a full view for Account links.

This causes a problem when MTs refer to other MTs in deep recursive structures. The generated swagger files can be really large and the swagger-ui has a hard time parsing them.

Not sure whether this is an issue with just the links or in general with any resource that specifies a default view other than `default` given the recent change in #661

raphael:
Fixed in both master and v1, thanks!

PR #702, opened 2016-08-07
Fix generation of examples for media types

raphael:
Using the non default view. This PR cleans up how projected media types are handled:
A projected media type identifier now includes the view so that it's possible to infer
whether a media type has already been projected. Projected media types also only
define the "default" view.


raphael:
Fix #665

PR #704, opened 2016-08-07
Fix ContentType isn't applied in MediaType (#694)

### LINKED ISSUE #694: Specify a custom Content-Type for Goa errors and MediaTypes, Types
Goa should provide a way to specify custom Content-Type instead of forcing RFC 6848. There are applications which don't respect that (i.e., swagger-codegen is producing Java client which fails to decode response).

Currently there is no way to do that.

dobegor:
Fixes #694.
`ContentType` wasn't applied, instead, the `MediaType` identifier was used.


raphael:
Do you think you could add a test?

raphael:
Thanks! I'll add a test in a separate PR.

dobegor:
Thanks! Haven't got enough skill to get into Goa's testing suite enough, though.

Issue #712, opened 2016-08-08
Test generator generating non-buildable tests when Integer type is required


rootpd:
Consider following simplified `MediaType` definition:

```go
var Foo = MediaType("application/vnd.foo+json", func() {
	Attributes(func() {
		Attribute("bar", Integer)
		Required("bar")
	})
})

var Baz = MediaType("application/vnd.baz+json", func() {
	Attributes(func() {
		Attribute("bar", String)
		Required("bar")
	})
})
```

Generated app snippet adds *required* validation only for the String type:

```go
// Foo description (default view)
//
// Identifier: application/vnd.foo+json
type Foo struct {
	Bar int `form:"bar" json:"bar" xml:"bar"`
}

// Baz description (default view)
//
// Identifier: application/vnd.baz+json
type Baz struct {
	Bar string `form:"bar" json:"bar" xml:"bar"`
}

// Validate validates the Baz media type instance.
func (mt *Baz) Validate() (err error) {
	if mt.Bar == "" {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "bar"))
	}
	return
}
```

However generated test snippet is trying to call `Validate` function in both cases. Attaching snippet generated for the `Baz:Integer`.

```go
// ...
// Validate response
if err != nil {
    t.Fatalf("controller returned %s, logs:\n%s", err, logBuf.String())
}
if rw.Code != 200 {
    t.Errorf("invalid response status code: got %+v, expected 200", rw.Code)
}
var mt *app.Foo
if resp != nil {
    var ok bool
    mt, ok = resp.(*app.Foo)
    if !ok {
        t.Errorf("invalid response media: got %+v, expected instance of app.Foo", resp)
    }
    err = mt.Validate() // <-- THIS ONE DOESN'T EXIST
    if err != nil {
        t.Errorf("invalid response media type: %s", err)
    }
}
```

The reason appears to be that *app* generator [is calling](https://github.com/goadesign/goa/blob/master/goagen/gen_app/writers.go#L763) `recursiveValidate` with parameter `private` set to `false`, but the *test* generator [is using](https://github.com/goadesign/goa/blob/master/goagen/gen_app/test_generator.go#L165) `true` for the same param when it's generating tests for media types.

This boils down to [following template](https://github.com/goadesign/goa/blob/master/goagen/codegen/validation.go#L344), which basically generates the validation only for the `String` type if `private` parameter is set to `true`.

[...]

PR #713, opened 2016-08-08
Fix spurious call to Validate() in test helpers

raphael:
The generated test helper code would sometimes call Validate()
on media types that did not define it because the test was being
made against the private MT data structure instead of the public one.


raphael:
Fix #712

PR #717, opened 2016-08-09
Generate cors headers without wildcards

### LINKED ISSUE #711: Cors headers with wildcards are invalid in browser
I have this bit on the design:

```go
Origin("*.arduino.cc", func() {
	Methods("GET", "PUT", "POST", "DELETE")
	Headers("Authorization", "Origin", "X-Requested-With", "Content-Type", "Accept")
	Credentials()
})
```

That results in the creation of methods such as:

```go
func handlePublicOrigin(h goa.Handler) goa.Handler {
	return func(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
		origin := req.Header.Get("Origin")
		if origin == "" {
			// Not a CORS request
			return h(ctx, rw, req)
		}
		if cors.MatchOrigin(origin, "*.arduino.cc") {
			ctx = goa.WithLogContext(ctx, "origin", origin)
			rw.Header().Set("Access-Control-Allow-Origin", "*.arduino.cc")
			rw.Header().Set("Vary", "Origin")
			rw.Header().Set("Access-Control-Allow-Credentials", "true")
			if acrm := req.Header.Get("Access-Control-Request-Method"); acrm != "" {
				// We are handling a preflight request
				rw.Header().Set("Access-Control-Allow-Methods", "GET, PUT, POST, DELETE")
				rw.Header().Set("Access-Control-Allow-Headers", "Authorization, Origin, X-Requested-With, Content-Type, Accept")
			}
			return h(ctx, rw, req)
		}

		return h(ctx, rw, req)
	}
}
```

Unfortunately this line:

```go
rw.Header().Set("Access-Control-Allow-Origin", "*.arduino.cc")
```

Causes the browser to refuse the request because

```
XMLHttpRequest cannot load http://localhost:9000/builder/v1/swagger.json. The 'Access-Control-Allow-Origin' header contains the invalid value '*.arduino.cc'. Origin 'http://webide.arduino.cc' is therefore not allowed access.
```

Should in case of matched origin set an header equal to the origin?

matteosuppo:
The header Access-Control-Allow-Origin doesn't accept
wildcards such as *.example.com. Therefore after the
origin has been checked as valid, the header should
contain the origin without wildcards.

For example if we got an origin such as test.example.com
and we accept origins matching *.example.com, the resulting
header will be: Access-Control-Allow-Origin: test.example.com

Fix #711 

Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>

### DIFF
--- a/goagen/gen_app/writers.go
+++ b/goagen/gen_app/writers.go
@@ -703,7 +703,7 @@ func handle{{ .Resource }}Origin(h goa.Handler) goa.Handler {
 		}
 {{ range $policy := .Origins }}		if cors.MatchOrigin(origin, {{ printf "%q" $policy.Origin }}) {
 			ctx = goa.WithLogContext(ctx, "origin", origin)
-			rw.Header().Set("Access-Control-Allow-Origin", "{{ $policy.Origin }}")
+			rw.Header().Set("Access-Control-Allow-Origin", origin)
 {{ if not (eq $policy.Origin "*") }}			rw.Header().Set("Vary", "Origin")
 {{ end }}{{ if $policy.Exposed }}			rw.Header().Set("Access-Control-Expose-Headers", "{{ join $policy.Exposed ", " }}")
 {{ end }}{{ if gt $policy.MaxAge 0 }}			rw.Header().Set("Access-Control-Max-Age", "{{ $policy.MaxAge }}")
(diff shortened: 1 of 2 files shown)


raphael:
This is great! thank you.

PR #723, opened 2016-08-10
Allow cors spec to be defined as regexps

### LINKED ISSUE #718: Multiple CORS Origins or Origin Regex
At the moment we can use wildcards only once, to define patterns such as `*.arduino.cc`.

But suppose I wanted to define a patterns such as `*.arduino.cc:*`, or multiple acceptable origins.

Is there a way to use multiple Origins?
If not, would it be a good idea to use a regular expression in the function Origin of the dsl?

matteosuppo:
If the string passed into Origin() begins and ends with a "/"
it's considered a regular expression.

For example "/[api|swagger].goa.design/" can match both
api.goa.design and swagger.goa.design

Fix #718

Signed-off-by: Matteo Suppo <matteo.suppo@gmail.com>

### DIFF
--- a/cors/cors.go
+++ b/cors/cors.go
@@ -26,7 +26,6 @@ const OriginKey key = "origin"
 // - a plain string identifying an origin. eg http://swagger.goa.design
 // - a plain string containing a wildcard. eg *.goa.design
 // - the special string * that matches every host
-// - A regular expression wrapped in //. eg /[api|swagger].goa.design/
 func MatchOrigin(origin, spec string) bool {
 	if spec == "*" {
 		return true
@@ -50,6 +49,13 @@ func MatchOrigin(origin, spec string) bool {
 		return false
 	}
 	return true
+}
+
+// MatchOriginRegexp returns true if the given Origin header value matches the
+// origin specification.
+// Spec must be a valid regex
+func MatchOriginRegexp(origin string, spec *regexp.Regexp) bool {
+	return spec.Match([]byte(origin))
 }
 
 // HandlePreflight returns a simple 200 response. The middleware takes care of handling CORS.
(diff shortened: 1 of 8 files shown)


raphael:
Thank you, this is a great PR! I'm wondering though if there's a way we could avoid having to recompile the regexp on every request. Maybe we could introduce a `MatchOriginRegexp` function in the `cors` package that accepts a regular expression for the `origin` parameter. The generated code would compile the regexp once when mounting the action.

matteosuppo:
Sounds reasonable

matteosuppo:
It could generate something like:

```go
func handleBoardsOrigin(h goa.Handler) goa.Handler {
    spec := regexp.MustCompile("[api|swagger].goa.design")
	return func(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
		origin := req.Header.Get("Origin")
		if origin == "" {
			// Not a CORS request
			return h(ctx, rw, req)
		}
		if cors.MatchOriginRegexp(origin, spec) {
			ctx = goa.WithLogContext(ctx, "origin", origin)
			rw.Header().Set("Access-Control-Allow-Origin", origin)
			rw.Header().Set("Access-Control-Allow-Credentials", "true")
			if acrm := req.Header.Get("Access-Control-Request-Method"); acrm != "" {
				// We are handling a preflight request
				rw.Header().Set("Access-Control-Allow-Methods", "GET, PUT, POST, DELETE")
				rw.Header().Set("Access-Control-Allow-Headers", "Authorization, Origin, X-Requested-With, Content-Type, Accept")
			}
			return h(ctx, rw, req)
		}

		return h(ctx, rw, req)
	}
}

matteosuppo:
Tomorrow I'll fix the tests

raphael:
Great, thanks. Generated code snippet looks good.

raphael:
Bonus points for adding a test for the generated code as well :)

raphael:
Awesome PR, thank you!

PR #736, opened 2016-08-22
goa.adapter#logit(): remove unneeded check which would make an unsafe concurrent modification

jim-slattery-rs:
- logit() doesn't need to modify `a.keyvals` because it was already fixed up in New()

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -76,9 +76,6 @@ func (a *adapter) logit(msg string, keyvals []interface{}, iserror bool) {
 		keyvals = append(keyvals, ErrMissingLogValue)
 	}
 	m := (len(a.keyvals) + 1) / 2
-	if len(a.keyvals)%2 != 0 {
-		a.keyvals = append(a.keyvals, ErrMissingLogValue)
-	}
 	n += m
 	var fm bytes.Buffer
 	lvl := "INFO"


raphael:
Thank you! could I convince you to add a test that would have failed prior to this fix?

logging.go
```diff
@@ -76,9 +76,6 @@ func (a *adapter) logit(msg string, keyvals []interface{}, iserror bool) {
 		keyvals = append(keyvals, ErrMissingLogValue)
 	}
 	m := (len(a.keyvals) + 1) / 2
-	if len(a.keyvals)%2 != 0 {
-		a.keyvals = append(a.keyvals, ErrMissingLogValue)
```
jim-slattery-rs: I couldn't think of a test to add -- this line is basically unreachable code, as far as I can tell.

`a.keyvals` will always have an even length, thanks to the check in New(): https://github.com/jim-slattery-rs/goa/blob/72a07361f9b5491636eeacaae3b3145b6fe63e96/logging.go#L62-L64

Therefore, we fortunately don't have to check it again here, since it would anyways be unsafe to mutate `a.keyvals`.

raphael:
oh I misread the fix... looks great, thanks!

PR #737, opened 2016-08-23
Fix gen_clinet to suppress blank lines

tchssk:
This is a trivial change.

## Before

```go
if bar != nil { 
 
    header.Set("foo", *bar) 
}
```

## After

```go
if bar != nil { 
    header.Set("foo", *bar) 
}
```


raphael:
Nice, thank you!

Issue #739, opened 2016-08-27
The length validations count the length of the string in bytes.


ikawaha:
The length validations such as MaxLength() / MinLength() count the length of the string in bytes.

For example MaxLength(2), `”ab”` is valid, 
but  `”😀😀”` is not valid, so len(`”😀😀”`) = 8.

Should the length validations count the length of the string in unicode characters?

raphael:
Thank you for the fix!

PR #740, opened 2016-08-27
Fix length validations to count the length of the string using by utf8.RuneCountInString().

ikawaha:
see #739


raphael:
Makes sense, thank you! Do you think you could add a test as well? Thanks!

ikawaha:
Thanks! I will try to add some test.

raphael:
Great, thank you!

Issue #742, opened 2016-08-29
goagen: Generated code doesn't compile


PSG-Luna:
I have the following resource using ErrorMedia:

```
    Action("getSendMovementStreamError", func() {
        Routing(GET("/movementStream/waitOnError/:tokenID"))
        Description("Get error information about why a stream failed. This will block until the send stream is closed.")
        Params(func() {
            Param("tokenID", UUID, "The streaming token UUID")
        })
        Response(OK, ErrorMedia)
    })
```

The following code is then generated:

```
// GetSendMovementStreamError runs the getSendMovementStreamError action.
func (c *PlayerController) GetSendMovementStreamError(ctx *app.GetSendMovementStreamErrorPlayerContext) error {
    // PlayerController_GetSendMovementStreamError: start_implement

    // Put your logic here

    // PlayerController_GetSendMovementStreamError: end_implement
    res := app.error{}
    return ctx.OK(res)
}
```

This code doesn't compile due to attempting to use an unexported variable.

PR #746, opened 2016-08-31
Correctly handle OK error response in genmain

raphael:
Handle special case where OK resposne returns ErrorMedia


raphael:
Fix #742

PR #751, opened 2016-08-31
Updated jwt middleware to allow a user to edit the list of accepted k…

jaredbischof:
…eys on demand. Also, allowing a user to identify which key to validate a JWT against by providing the jwtkeyname field in the request header.

### DIFF
--- a/middleware/security/jwt/jwt.go
+++ b/middleware/security/jwt/jwt.go
@@ -2,15 +2,152 @@ package jwt
 
 import (
 	"crypto/rsa"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
 	"strings"
+	"sync"
 
 	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/goadesign/goa"
 	"golang.org/x/net/context"
 )
+
+// ErrInvalidKey is returned when a key is not of type string, []string,
+// *rsa.PublicKey or []*rsa.PublicKey.
+var ErrInvalidKey = errors.New("invalid parameter, the only keys accepted " +
+	"are *rsa.publicKey, []*rsa.PublicKey (for RSA-based algorithms) or a " +
+	"signing secret string, []string (for HS algorithms)")
+
+// ErrKeyDoesNotExist is returned when a key cannot be found by the provided
+// key name.
+var ErrKeyDoesNotExist = errors.New("key does not exist")
+
+// KeyResolver is a struct that is passed into the New() function, which allows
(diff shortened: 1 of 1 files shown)


middleware/security/jwt/jwt.go
```diff
@@ -2,16 +2,153 @@ package jwt
...
+	case *rsa.PublicKey:
+		kr.keyMap[name] = append(kr.keyMap[name], keys)
+	case []*rsa.PublicKey:
+		for _, key := range keys {
+			kr.keyMap[name] = append(kr.keyMap[name], key)
+		}
+	case string:
+		kr.keyMap[name] = append(kr.keyMap[name], keys)
+	case []string:
+		for _, key := range keys {
```
raphael: Couldn't that just be
```
kr.keyMap[name] = append(kr.keyMap[name], keys...)
```?

File: middleware/security/jwt/jwt.go [hunk 2]
```diff
@@ -2,16 +2,153 @@ package jwt
...
+	}
+	return
+}
+
+// GetAllKeys returns a list of all the keys stored in the resolver.
+func (kr *KeyResolver) GetAllKeys() []interface{} {
+	kr.Lock()
+	defer kr.Unlock()
+	var keys []interface{}
+	for name := range kr.keyMap {
```
raphael: This loop could simply be:
```go
for _, ks := range kr.keyMap {
    keys = append(keys, ks...)
}
```

middleware/security/jwt/jwt.go
raphael: Ah no that doesn't work because the types don't match...

middleware/security/jwt/jwt.go
raphael: Same as above, my suggestion doesn't work because the slice elements have different types...

middleware/security/jwt/jwt.go
```diff
@@ -91,18 +208,62 @@ func New(validationKeys interface{}, validationFunc goa.Middleware, scheme *goa.
 
 			incomingToken := strings.Split(val, " ")[1]
 
+			jwtKeyName := req.Header.Get("jwtkeyname")
```
raphael: Would it make sense to make the name of the header configurable? Maybe it's a field on the resolver? `KeyNameHeader` or something like that?

raphael:
Looks great overall! thank you! do you think you could also add a few tests?

jaredbischof:
Will update this soon, just haven't had time recently.

PR #756, opened 2016-09-02
goagen/gen_app: Extra parameter in generated Href functions.

### LINKED ISSUE #755: There is always "param" in generated code even if one is not specified for Hrefs
```
var _ = Resource("currentUser", func() {
	BasePath("")
	Action("show", func() {
		Routing(GET("/profile"))
		Response(OK)
	})
})
```
Produces the following:
```
// CurrentUserHref returns the resource href.
func CurrentUserHref() string {
	return fmt.Sprintf("/profile", param)
}
```

Which fails to build

faxal:
Fixes #755


raphael:
Thank you! I hadn't seen this PR before I submitted my own fix in #760 .... I prefer my fix because it has a test and does not generate a call to `fmt.Sprintf` if not needed so closing this PR. This was the correct change though so thank you!

PR #771, opened 2016-09-07
Modify metrics so that names are normalized to legal chars

theshadow:
* Add function for normalizing metric keys
* Add tests for metrics functions.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -38,12 +38,12 @@ type Metrics interface {
 
 // used for dealing with race conditions.
 type metricDiscarder struct{}
-func (md *metricDiscarder) SetGauge(key []string, val float32) {}
-func (md *metricDiscarder) EmitKey(key []string, val float32) {}
-func (md *metricDiscarder) IncrCounter(key []string, val float32) {}
-func (md *metricDiscarder) AddSample(key []string, val float32) {}
-func (md *metricDiscarder) MeasureSince(key []string, start time.Time) {}
 
+func (md *metricDiscarder) SetGauge(key []string, val float32)         {}
+func (md *metricDiscarder) EmitKey(key []string, val float32)          {}
+func (md *metricDiscarder) IncrCounter(key []string, val float32)      {}
+func (md *metricDiscarder) AddSample(key []string, val float32)        {}
+func (md *metricDiscarder) MeasureSince(key []string, start time.Time) {}
 
 // NewMetrics initializes goa's metrics instance with the supplied
 // configuration and metrics sink
--- a/metrics.go
+++ b/metrics.go
@@ -42,7 +42,7 @@ func NewNoOpSink() metrics.MetricSink {
 	return &NoOpSink{}
 }
 
-// NoopSink default NOOP metrics recorder
+// NoOpSink default NOOP metrics recorder
 type NoOpSink struct{}
 
 // SetGauge method
--- a/metrics.go
+++ b/metrics.go
@@ -3,13 +3,26 @@
 package goa
 
 import (
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/armon/go-metrics"
 )
 
-// metriks is the local instance of metrics.Metrics
-var metriks *metrics.Metrics
+const (
+	allMatcher      string = "*/*"
+	allReplacement  string = "all"
+	normalizedToken string = "_"
+)
+
+var (
+	// metriks is the local instance of metrics.Metrics
+	metriks *metrics.Metrics
+
+	// used for normalizing names by matching '*' and '/' so they can be replaced.
+	invalidCharactersRE = regexp.MustCompile(`[\*/]`)
+)
 
 // NewMetrics initializes goa's metrics instance with the supplied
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -63,6 +96,21 @@ func MeasureSince(key []string, start time.Time) {
...
 	if metriks != nil {
+		normalizeKeys(key)
 		metriks.SetGauge(key, val)
 	}
 }
+
+// This function is used to make metric names safe for all metric services. Specifically, prometheus does
+// not support * or / in metric names.
+func normalizeKeys(key []string) {
+	if key != nil {
```
raphael: this check is not needed (you can `range` over nil)

metrics.go
```diff
@@ -3,27 +3,57 @@
...
+	allReplacement  string = "all"
+	normalizedToken string = "_"
+)
+
+var (
+	// interface for metrics.Metrics
+	metriks Metrics
+
+	// used for normalizing names by matching '*' and '/' so they can be replaced.
+	invalidCharactersRE = regexp.MustCompile(`[\*/]`)
```
raphael: should the regexp be the same as the one in the prometheus package? the test should be whether the character does *not* match the regexp (white list instead of black list).

metrics.go
```diff
@@ -3,27 +3,57 @@
...
 
-// NewMetrics initializes goa's metrics instance with the supplied
+// Deprecated - NewMetrics initializes goa's metrics instance with the supplied
 // configuration and metrics sink
 func NewMetrics(conf *metrics.Config, sink metrics.MetricSink) (err error) {
 	metriks, err = metrics.NewGlobal(conf, sink)
 	return
 }
 
+// SetMetrics initializes goa's metrics instance with the supplied
```
raphael: ... with the supplied metrics adapter interface.

metrics.go
```diff
@@ -63,6 +96,21 @@ func MeasureSince(key []string, start time.Time) {
...
+// This function is used to make metric names safe for all metric services. Specifically, prometheus does
+// not support * or / in metric names.
+func normalizeKeys(key []string) {
+	if key != nil {
+		for i, k := range(key) {
+			// first replace */* with all
+			k = strings.Replace(k, allMatcher, allReplacement, -1)
+
+			// now replace all other invalid characters with a safe one.
+			key[i] = invalidCharactersRE.ReplaceAllString(k, normalizedToken)
```
raphael: See comment on the regexp, this should be a test to see whether the character is in a set of white listed characters.

metrics_test.go
```diff
@@ -0,0 +1,120 @@
+package goa_test
```
raphael: Tests look great!

metrics.go
```diff
@@ -49,44 +58,55 @@ func (md *metricDiscarder) MeasureSince(key []string, start time.Time) {}
...
+	m, err := metrics.NewGlobal(conf, sink)
+	SetMetrics(m)
+
+	return nil
 }
 
 // SetMetrics initializes goa's metrics instance with the supplied metrics adapter interface.
 func SetMetrics(m Metrics) {
-	metriks = m
+	mu.Lock()
```
raphael: The mutex is not needed anymore with `atomic.Value`

metrics.go
```diff
@@ -3,47 +3,119 @@
...
+	allMatcher      string = "*/*"
+	allReplacement  string = "all"
+	normalizedToken string = "_"
+)
+
+var (
+	// metriks atomic value storage
+	metriks atomic.Value
+
+	// mu mutex for metriks
```
raphael: Please remove the mutex, it is not needed anymore (see the [first example](https://golang.org/pkg/sync/atomic/#Value) of sync.Value).

metrics.go
```diff
@@ -3,47 +3,119 @@
...
+	mu sync.Mutex
+
+	// used for normalizing names by matching '*' and '/' so they can be replaced.
+	invalidCharactersRE = regexp.MustCompile(`[\*/]`)
+
+	// Taken from https://github.com/prometheus/client_golang/blob/66058aac3a83021948e5fb12f1f408ff556b9037/prometheus/desc.go
+	metricsNameRE = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_:]*$`)
```
raphael: The regexp above doesn't seem to be used anywhere. It seems to be the better one though since it is more precise. Could you adjust the code to use this regexp instead of `invalidCharactersRE` ? It's the reverse condition though so the code needs to be adapted as well...

theshadow: I'm confused, it is actually used in normalize. How were you expecting the normalize function to operate? Currently this is it's flow.

```
normalizeKeys(keys):
    for each key in keys:
         if not validCharsRegex.Match(key):
               key = invalidCharsRegex.Replace(key, "_")
```

raphael: oh whoops missed that. It seems the test for invalid characters is not sufficient though? there are many other characters than * and / that don't pass the metricsNameRE regexp.

raphael:
Thank you for all the work!

PR #775, opened 2016-09-13
Fix to contain only valid tags in gen_swagger

tchssk:
This is an API definition.

```go
var _ = apidsl.API("An API", func() {
    apidsl.Metadata("swagger:tag:Backend")
    apidsl.Metadata("struct:tag:json", "myName,omitempty")
    apidsl.Metadata("swagger:gat:dnekcaB")
```

`gen_swagger` generates following swagger.yaml (and json).

```yaml
tags:
  name: Backend
  name: json
  name: dnekcaB
```

But valid tag is only top so it should be outputs only that.

```yaml
tags:
  name: Backend
```

### DIFF
--- a/goagen/gen_swagger/swagger.go
+++ b/goagen/gen_swagger/swagger.go
@@ -448,7 +448,12 @@ func scopesMapList(scopes map[string]string) string {
 }
 
 func tagsFromDefinition(mdata dslengine.MetadataDefinition) (tags []*Tag) {
-	for key, value := range mdata {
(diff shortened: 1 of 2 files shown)


raphael:
whoops. thank you! Do you think you could add a test that would fail without the fix?

tchssk:
I added a test and fixed `tagsFromDefinition()` to use prepared keys to iterate over maps.

raphael:
This is great, thank you!

PR #779, opened 2016-09-16
Support swagger extensions

tchssk:
This pull request extends `apidsl.Metadata()` to support [Swagger Extensions](https://github.com/OAI/OpenAPI-Specification/blob/master/guidelines/EXTENSIONS.md).
Now it includes only core implementations so I will add tests and comments for godoc.

https://gophers.slack.com/archives/goa/p1473747445000640

### DIFF
--- a/goagen/gen_swagger/swagger.go
+++ b/goagen/gen_swagger/swagger.go
@@ -290,6 +290,7 @@ type (
 		Extensions map[string]string `json:"-"`
 	}
 
+	// These types are used in marshalJSON() to avoid recursive call of json.Marshal().
 	_Info               Info
 	_Path               Path
 	_Operation          Operation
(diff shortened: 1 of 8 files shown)


goagen/gen_swagger/swagger.go
```diff
@@ -274,9 +286,70 @@ type (
 		Description string `json:"description,omitempty"`
 		// ExternalDocs is additional external documentation for this tag.
 		ExternalDocs *ExternalDocs `json:"externalDocs,omitempty"`
+		// Extensions defines the swagger extensions.
+		Extensions map[string]string `json:"-"`
 	}
+
+	_Info               Info
```
raphael: Can you leave a comment explaining why the `_` types are needed?

goagen/gen_swagger/swagger.go
```diff
@@ -274,9 +286,70 @@ type (
...
+		asserted[k] = v
+	}
+	merged, err := json.Marshal(asserted)
+	if err != nil {
+		return nil, err
+	}
+	return merged, nil
+}
+
+func (i Info) MarshalJSON() ([]byte, error) {
```
raphael: Please add a comment for these

tchssk:
Oops, I added the comments but travis tells me another error...

raphael:
Pro tip: run `make` before pushing, that's what Travis does :)

tchssk:
Thanks. I forget to `make` sometimes😓

tchssk:
Thank you for your waiting. It is ready.

raphael:
Do you think you could also send a PR against the v1 branch? this is a backwards compatible change and it would be great if people on the v1 branch could benefit from it too :)

tchssk:
It's OK. But this PR is based on #775 so I will send that to v1 first.

raphael:
Perfect, thanks again!

PR #782, opened 2016-09-19
Fix to contain only valid tags in gen_swagger

tchssk:
This is a backport of #775 to [v1](https://github.com/goadesign/goa/releases/tag/v1.0.0).


raphael:
Thank you!

PR #788, opened 2016-09-22
Add ability to set JSON encoded  Swagger extension values

raphael:


raphael:
@tchssk fyi, you can now encode values in JSON in the Metadata field for Swagger extensions.

tchssk:
Great! Thank you for telling me. @raphael

tchssk:
I read the document of Swagger Extensions again.
It includes follwing line.

>Extension properties are always prefixed by "x-" and can have any valid JSON format value.

If it means 

>Extension properties are always prefixed by "x-" and can have **only** any valid JSON format value.

maybe we should have picked up only valid values as JSON and simply stored them as `string` like below.

```go
func extensionsFromDefinition(mdata dslengine.MetadataDefinition) map[string]string {
    extensions := make(map[string]string)
    for key, value := range mdata {
        chunks := strings.Split(key, ":")
        if len(chunks) != 3 {
            continue
        }
        if chunks[0] != "swagger" || chunks[1] != "extension" {
            continue
        }
        if strings.HasPrefix(chunks[2], "x-") != true {
            continue
        }
        // Validate value.
        var unmarshaled interface{}
        if err := json.Unmarshal(value[0], &unmarshaled); err != nil {
            // Skip if invalid.
            continue
        }
        // Set if valid.
        extensions[chunks[2]] = value[0]
    }
    if len(extensions) == 0 {
        return nil
    }
    return extensions
}
```

How do you think? @raphael

raphael:
Hmm yes that's a good point. It's going to be less practical for strings though as you'd have to double escape them. I could see people getting bitten by this. Maybe we could keep doing what the algorithm does today and if json unmarshal fails then fall back to using the value directly. It's not great from a performance point of view but this is code generation so it's fine :)

tchssk:
I almost understood.
Thank you for your reply!

raphael:
Yeah my response wasn't super clear now that I re-read it :)

What I mean to say is that like you suggest we should remove the `json:` prefix and always do `json.Unmarshal` on the value. However if that fails I think we shouldn't fail entirely and instead just use the value as is. So I can still write:

```
Metadata("swagger:extension:x-name", "string value")
```

I don't have to do:

```
Metadata("swagger:extension:x-name", `"string value"`)
```

But that would work too.

tchssk:
Oh I understand.

And I was planned to undo each `Extensions` to `map[string]string` but they may have to keep being currently type. A value may have to be stored as `interface{]` when it valid and as `string` when it invalid.

raphael:
cool! yeah I think the `Extensions` fields have to be `map[string]interface{}`

tchssk:
OK. I'll send a new pull request.

PR #790, opened 2016-09-24
Try to use swagger extensions as JSON by default

tchssk:
This is related to #788.


raphael:
Looks great, thank you!

tchssk:
I made it! #791

PR #796, opened 2016-10-05
Always use goas uuid package

### LINKED ISSUE #795: Inconsistent uuid package used in generation
Most places in goa use `github.com/goadesign/goa/uuid` but the application code generator uses `github.com/satori/go.uuid`. This creates incompatible types when using Gorma to generate models. 

I propose we fix `goagen` to always use the goas uuid package.

stengaard:
This avoids generating incompatible types.

Fixes #795

### DIFF
--- a/design/random.go
+++ b/design/random.go
@@ -6,8 +6,8 @@ import (
 	"math/rand"
 	"time"
 
+	"github.com/goadesign/goa/uuid"
 	"github.com/manveru/faker"
-	"github.com/satori/go.uuid"
 )
 
 // RandomGenerator generates consistent random values of different types given a seed.
--- a/design/types.go
+++ b/design/types.go
@@ -18,7 +18,7 @@ import (
 	"time"
 
 	"github.com/goadesign/goa/dslengine"
-	"github.com/satori/go.uuid"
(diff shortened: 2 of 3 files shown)


raphael:
Wrt the failed test, please update the content of `.travis.yml` with the latest from master - this removes the `1.5.3` build which is broken (and adds `1.7.1`). Thanks for looking into this!

raphael:
@stengaard ping :) are you still looking into this?

stengaard:
No, sorry. Let's close this - I won't have time in the foreseeable future. :(

schmorrison:
I am curious about this because I can't seem to get gorma to use a `gorma.UUID` in a field without seeing this on build:
`../api/models/bidmodel_helper.go:50: cannot use &m.UUID (type *"github.com/goadesign/goa/uuid".UUID) as type *"github.com/satori/go.uuid".UUID in assignment`

Is changing the uuid package in the generated files from satori/go.uuid to goa/uuid a reasonable fix in the mean time?

schmorrison

raphael:
I don't have enough information to make that determination :) Which generated file(s) import statement would need to change?

PR #806, opened 2016-10-07
Added the ability to dynamically add/remove keys from the jwt securit…

jaredbischof:
…y middleware after creation.

### DIFF
--- a/middleware/security/jwt/jwt.go
+++ b/middleware/security/jwt/jwt.go
@@ -2,15 +2,156 @@ package jwt
 
 import (
 	"crypto/rsa"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
 	"strings"
+	"sync"
 
 	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/goadesign/goa"
 	"golang.org/x/net/context"
 )
+
+// ErrInvalidKey is returned when a key is not of type string, []string,
+// *rsa.PublicKey or []*rsa.PublicKey.
+var ErrInvalidKey = errors.New("invalid parameter, the only keys accepted " +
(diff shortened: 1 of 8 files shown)


middleware/security/jwt/jwt.go
```diff
@@ -2,16 +2,157 @@ package jwt
...
+}
+
+// AddKeys can be used to add keys to the resolver which will be referenced
+// by the provided name. Acceptable types for keys include string, []string,
+// *rsa.PublicKey or []*rsa.PublicKey. Multiple keys are allowed for a single
+// key name to allow for key rotation.
+func (kr *KeyResolver) AddKeys(name string, keys interface{}) error {
+	kr.Lock()
+	defer kr.Unlock()
+	if _, ok := kr.keyMap[name]; !ok {
```
raphael: This initialization is not needed, the first argument of `append` can be `nil`.

middleware/security/jwt/jwt.go
```diff
@@ -91,18 +212,41 @@ func New(validationKeys interface{}, validationFunc goa.Middleware, scheme *goa.
 
 			incomingToken := strings.Split(val, " ")[1]
 
+			var jwtKeyName string
+			if resolver.jwtKeyNameField != "" {
+				jwtKeyName = req.Header.Get(resolver.jwtKeyNameField)
+			}
+
+			// Make a copy of the current keys in the KeyResolver's key map
+			resolver.Lock()
```
raphael: It would probably make sense to use a `RWLock` instead of just a lock since here we only need a read lock. This seems like a valuable change because this code runs for each request and using a simple lock forces synchronization which could impact performance and is not needed.

jsimonetti:
This can probably be closed as it is merged in #818

raphael:
Yes, thank you!

PR #817, opened 2016-10-11
Change UseTrait APIDSL function to be variadic v1

jsimonetti:
Signed-off-by: Jeroen Simonetti jeroen@simonetti.nl


jsimonetti:
eh, oops :P

PR #818, opened 2016-10-11
Added the ability to dynamically add/remove keys from the jwt securit…

jaredbischof:
…y middleware after creation.


raphael:
Awesome, thank you!

PR #823, opened 2016-10-13
Do not add a list of security scopes in the description swagger, if it is empty

EvgenyGavrilov:


raphael:
Thank you! Do you think you could make a PR against the `v1` branch as well?

PR #825, opened 2016-10-14
Fix a bug that the payload is always required 'true' in swagger documents.

ikawaha:
https://github.com/goadesign/goa/blob/master/goagen/gen_swagger/swagger.go#L882
```go
	if action.Payload != nil {
		payloadSchema := genschema.TypeSchema(api, action.Payload)
		pp := &Parameter{
			Name:        "payload",
			In:          "body",
			Description: action.Payload.Description,
			Required:    true,            // ←★ always true even if optional payloads be used.
			Schema:      payloadSchema,
		}
		params = append(params, pp)
	}
```


tchssk:
Oh I forgot to fix `gen_swagger` in #507😅

raphael:
Looks great, thank you! Do you think you could also make a PR to the `v1` branch? thanks!

PR #826, opened 2016-10-14
Fix a bug that the payload is always required 'true' in swagger documents.

ikawaha:
This is a backport of #825 to v1.


raphael:
Thank you!

PR #832, opened 2016-10-16
Refactor JWT key resolver

raphael:
To introduce new "simple" resolver.


raphael:
@jaredbischof FYI this maintains the same interface but cleans up things a little bit and adds a new "simple" resolver so that the simple case stays simple.

jaredbischof:
@raphael Thanks for notifying me.

PR #837, opened 2016-10-24
Add unix domain socket listener to service.ListenAndServeUnix

hatajoe:
Hi.

I very often use unix domain socket listener with HTTP proxy server. (e.g, Nginx)
In goa, it looks no way to use that.
This p-r add function `service.ListenAndServeUnix` that support unix domain socket listener.
Can you review this?

Thank you.

### DIFF
--- a/service.go
+++ b/service.go
@@ -157,9 +157,9 @@ func (service *Service) ListenAndServeTLS(addr, certFile, keyFile string) error 
 	return http.ListenAndServeTLS(addr, certFile, keyFile, service.Mux)
 }
 
-// Serve starts specified listener on the given host.
-func (service *Service) Serve(listener net.Listener) error {
-	if err := http.Serve(listener, service.Mux); err != nil {
+// Serve accepts incoming HTTP connections on the listener l, invoking the service mux handler for each.
+func (service *Service) Serve(l net.Listener) error {
+	if err := http.Serve(l, service.Mux); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


raphael:
Thanks for the PR! goa has to compile / run on Windows and in App Engine (package syscall cannot be used there). Also it seems a bit strange to handle signals only when listening on a Unix socket. Given all this and how it's simple to do this in your own code I'm not sure this is worth adding to goa. Is there a reason you cannot have that code outside? Maybe this should be an example instead.

hatajoe:
Sorry, I was not able to consider multiple environments. 
And certainly seems little bit strange.

I will take you another approach that inject listener to function like [net.http.Serve](https://golang.org/pkg/net/http/#Serve). 
This approach seems simple than before.
Would you like about this? (I will push soon)

raphael:
This approach seems much better indeed, thank you! Just one nitpick: do you think you could change the comment of the method to something like: ```
Serve accepts incoming HTTP connections on the listener l, invoking the service mux handler for each.
```

hatajoe:
I fixed it.
My English is not good enough ;)
Thank you for your review.

raphael:
This is great thank you! Do you think you could also make a PR against the `v1` branch?

hatajoe:
I made a p-r https://github.com/goadesign/goa/pull/838 .
but seems something I was wrong?

PR #858, opened 2016-11-10
swagger ui fails to send a proper request when a paramter is Arrayof(...)

fede-bitlogic:
…nd multiple parameters, instead of a single parameter separated by comma


raphael:
Great, thank you!

Issue #854, opened 2016-11-07
Ran into issue with using type name with ArrayOf for recursive data types


illyabusigin:
Hi there!

## Issue
I was trying to setup a recursive Type value like so:
```go
// StructureNodePayload represnts a node in the structure hierarchy user type
var StructureNodePayload = Type("StructureNodePayload", func() {
	Attribute("id", String, "The identifier for the node. This should correspond to structure identifier.", func() {
		Example("fQLPQBLFjGKmmmw8sWmGU")
	})

	Attribute("children", ArrayOf("StructureNodePayload"), "Child nodes associated with the structure, if any.")
	Required("id")
})
```

I attempted to generate the code using the following `goagen` command: `goagen app -d test_goa_issue/design`

The above code causes a goroutine stack limit panic:
```
exit status 2
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x416d14, 0xe)
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:566 +0x95
runtime.newstack()
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/stack.go:1061 +0x416
runtime.morestack()
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/asm_amd64.s:366 +0x7f

goroutine 1 [running]:
reflect.(*rtype).MethodByName(0x39b580, 0x42af68, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/usr/local/Cellar/go/1.7.3/libexec/src/reflect/type.go:858 fp=0xc440500278 sp=0xc440500270
reflect.Value.MethodByName(0x39b580, 0xc42693e300, 0x15, 0x42af68, 0x5, 0x15, 0x0, 0x0)
	/usr/local/Cellar/go/1.7.3/libexec/src/reflect/value.go:1142 +0x78 fp=0xc440500348 sp=0xc440500278
text/template.(*state).evalField(0xc440501490, 0x39b580, 0xc42693e300, 0x15, 0x42af68, 0x5, 0x6059e0, 0xc4201708a0, 0xc440500828, 0x1, ...)
	/usr/local/Cellar/go/1.7.3/libexec/src/text/template/exec.go:548 +0x153 fp=0xc440500630 sp=0xc440500348
[...]

PR #860, opened 2016-11-12
Properly handle recursive data types

raphael:
When generating code for validating and finalizing user types.


raphael:
Fix #854

PR #880, opened 2016-11-18
Fix uninitialized variable causing `goagen client --notool` crash

Syeberman:


goagen/gen_client/generator_test.go
```diff
@@ -90,6 +90,17 @@ var _ = Describe("Generate", func() {
...
+		Context("with --notool", func() {
+			BeforeEach(func() {
+				os.Args = append(os.Args, "--notool")
+			})
+
+			It("should not return an error", func() {
+				Ω(genErr).Should(BeNil())
+				Ω(files).Should(HaveLen(5)) // 9, minus 4 entries for tool paths
+			})
+		})
```
Syeberman: If you run commit 8e2e4d5 on its own you'll see this test crash.

goagen/gen_client/generator.go
```diff
@@ -149,14 +149,16 @@ func (g *Generator) Generate() (_ []string, err error) {
...
 		}
+		arrayToStringTmpl = template.Must(template.New("client").Funcs(funcs).Parse(arrayToStringT))
+	}
+
+	if !g.NoTool {
+		var cliPkg string
 		cliPkg, err = codegen.PackagePath(cliDir)
 		if err != nil {
 			return
 		}
```
Syeberman: GitHub's diff doesn't make this clear, but all I've done is moved the above `cliPkg` code down to inside the `if !g.NoTool` block.  `cliDir` is only set when `g.NoTool` is false, so using `--notool` would always result in a panic.

raphael:
Thank you for the fix and clear PR! Do you think you could backport the fix to the `v1` branch? Thank you!

Syeberman:
Check out https://github.com/goadesign/goa/pull/882 for the v1 backport

PR #882, opened 2016-11-18
Cherry-pick 3f168ff onto v1: Fix uninitialized variable causing `goag…

Syeberman:
…en client --notool` crash


raphael:
Thank you!

PR #897, opened 2016-11-21
fix unlinked reference

azihsoyn:
maybe ref #878


azihsoyn:
@raphael Please review this PR.
Is this fit your intent?

Thanks.

raphael:
Whoops, thank you. I'm wondering if that always produces valid JSON schema identifiers though.... I'll merge since things are clearly broken right now but we should make sure that doesn't end up producing invalid code.

raphael:
Could you back port that fix to the v1 branch please?

azihsoyn:
Thanks merging!

I pushed v1 PR #902

Regards.

PR #900, opened 2016-11-21
Fixup for 2e83b0dfdd45e1e54615606924dca565b547b08b

kwk:
We are using the v1 branch and lately there seems to be a change about the `contentType` parameter. I've commented on the recent commit [here](https://github.com/goadesign/goa/commit/2e83b0dfdd45e1e54615606924dca565b547b08b#commitcomment-19905560). 

This change removes the parameter `contentType` parameter from a function call under the same conditions as introduced in 2e83b0dfdd45e1e54615606924dca565b547b08b.

Also, I've added a workaround to remove the `header declared and not used error` which was caused because no content type is being anymore.

### DIFF
--- a/goagen/gen_client/generator.go
+++ b/goagen/gen_client/generator.go
@@ -989,8 +989,7 @@ func (c *Client) {{ $funcName }}(ctx context.Context, path string{{ if .Params }
 {{ end }}	if err != nil {
 		return nil, err
 	}
-{{ if or .Headers .HasPayload }}	header := req.Header
-		_ = header // Avoid "header declared and not used" error
+{{ if or .Headers (and .HasPayload .HasMultiContent) }}	header := req.Header
 {{ if .HasPayload }}{{ if .HasMultiContent }}	if contentType != "*/*" {
 		header.Set("Content-Type", contentType)
 	}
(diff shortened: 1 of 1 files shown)


goagen/gen_client/generator.go
```diff
@@ -990,6 +990,7 @@ func (c *Client) {{ $funcName }}(ctx context.Context, path string{{ if .Params }
 		return nil, err
 	}
 {{ if or .Headers .HasPayload }}	header := req.Header
+		_ = header // Avoid "header declared and not used" error
 {{ if .HasPayload }}{{ if .HasMultiContent }}	if contentType != "*/*" {
```
raphael: It seems moving the check for `HasMultiContent` might be better, something like:
```go
{{ if or .Headers (and .HasPayload .HasMultiContent) }}	header := req.Header
{{ if and .HasPayload .HasMultiContent }}	if contentType != "*/*" {
```
The idea being that we only to set headers if there's a payload and a `contentType` parameter or if there are headers to be set. The code above also combines the two `if` on line `994` into one which means one of the `end` below (on line `997` must be removed). Thank you!

raphael:
Thank you! I added a comment in the commit, do you think you could make that change? and once you've made it could you also make a PR against branch `v1`. Thank you again, much appreciated!

goagen/gen_client/generator.go
kwk: @raphael Fixed in 762a4962cd8e27e788396e6b8e1f3054f8894198

raphael:
Thank you!

kwk:
@raphael 

> [...] and once you've made it could you also make a PR against branch v1. 

Here's the PR for v1: https://github.com/goadesign/goa/pull/903

PR #901, opened 2016-11-21
Fix unlinked reference

azihsoyn:
maybe ref #878


azihsoyn:
Sorry

PR #902, opened 2016-11-21
fix unlinked reference (v1)

azihsoyn:
maybe ref #878 

via https://github.com/goadesign/goa/pull/897#issuecomment-261984051


raphael:
Thank you!

raphael:
Thank you!

PR #903, opened 2016-11-21
Backport for #900

kwk:
* Fixup for 2e83b0dfdd45e1e54615606924dca565b547b08b

We are using the v1 branch and lately there seems to be a change about the `contentType` parameter. I've commented on the recent commit [here](https://github.com/goadesign/goa/commit/2e83b0dfdd45e1e54615606924dca565b547b08b#commitcomment-19905560). 

This change removes the parameter `contentType` parameter from a function call under the same conditions as introduced in 2e83b0dfdd45e1e54615606924dca565b547b08b.

Also, I've added a workaround to remove the `header declared and not used error` which was caused because no content type is being anymore.

* Create header if .HasPayload .HasMultiContent 

See https://github.com/goadesign/goa/pull/900#pullrequestreview-9501415 for change request.


raphael:
Thank you!

PR #907, opened 2016-11-22
Revert to using type names for Swagger schema IDs

raphael:
Since the names are different for media type projected to different
views anyway.


raphael:
@kwk fyi I'm reverting the change that switched to media type identifiers for creating JSON schema IDs. Using names is safer as the reference path cannot get confused by the `/` in the identifier.

PR #909, opened 2016-11-22
Fix for #908

### LINKED ISSUE #908: In generated code contentType is passed to a generated function that doesn't accept a contentType
Hi,

in [`goagen/gen_client/cli_generator.go`](https://github.com/goadesign/goa/blob/b0fd166d97a02d5e7659a876491fc010fd46f785/goagen/gen_client/cli_generator.go#L799) there's a definition of this template variable:

```go
const commandsTmpl = `
{{ $cmdName := goify (printf "%s%sCommand" .Action.Name (title .Resource.Name)) true }}// Run makes the HTTP request corresponding to the {{ $cmdName }} command.
func (cmd *{{ $cmdName }}) Run(c *{{ .Package }}.Client, args []string) error {
	var path string
	if len(args) > 0 {
		path = args[0]
	} else {
{{ $default := defaultPath .Action }}{{ if $default }}	path = "{{ $default }}"
{{ else }}{{ $pparams := defaultRouteParams .Action }}	path = fmt.Sprintf({{ printf "%q" (defaultRouteTemplate .Action) }}, {{ joinFieldNames $pparams }})
{{ end }}	}
{{ if .Action.Payload }}var payload {{ gotyperefext .Action.Payload 2 .Package }}
	if cmd.Payload != "" {
		err := json.Unmarshal([]byte(cmd.Payload), &payload)
		if err != nil {
{{ if eq .Action.Payload.Type.Kind 4 }}	payload = cmd.Payload
{{ else }}			return fmt.Errorf("failed to deserialize payload: %s", err)
{{ end }}		}
	}
{{ end }}	logger := goa.NewLogger(log.New(os.Stderr, "", log.LstdFlags))
	ctx := goa.WithLogger(context.Background(), logger){{ $specialTypeResult := handleSpecialTypes .Action.QueryParams .Action.Headers }}{{ $specialTypeResult.Output }}
	resp, err := c.{{ goify (printf "%s%s" .Action.Name (title .Resource.Name)) true }}(ctx, path{{ if .Action.Payload }}, {{/*
	*/}}{{ if or .Action.Payload.Type.IsObject .Action.Payload.IsPrimitive }}&{{ end }}payload{{ else }}{{ end }}{{/*
	*/}}{{ $params := joinNames true .Action.QueryParams .Action.Headers }}{{ if $params }}, {{ format $params $specialTypeResult.Temps }}{{ end }}{{/*
	*/}}{{ if .Action.Payload }}, cmd.ContentType{{ end }})
	if err != nil {
		goa.LogError(ctx, "failed", "err", err)
		return err
	}
	goaclient.HandleResponse(c.Client, resp, cmd.PrettyPrint)
	return nil
}
`
```

I'm especially interested in this line:

```
	*/}}{{ if .Action.Payload }}, cmd.ContentType{{ end }})
```

The template generates this code in our application:

**tool/cli/commands.go**
```go
// Run makes the HTTP request corresponding to the CreateTrackerCommand command.
func (cmd *CreateTrackerCommand) Run(c *client.Client, args []string) error {
	var path string
	if len(args) > 0 {
		path = args[0]
	} else {
		path = "/api/trackers"
	}
	var payload client.CreateTrackerAlternatePayload
	if cmd.Payload != "" {
		err := json.Unmarshal([]byte(cmd.Payload), &payload)
		if err != nil {
			return fmt.Errorf("failed to deserialize payload: %s", err)
		}
	}
	logger := goa.NewLogger(log.New(os.Stderr, "", log.LstdFlags))
	ctx := goa.WithLogger(context.Background(), logger)
	resp, err := c.CreateTracker(ctx, path, &payload, cmd.ContentType)
	if err != nil {
		goa.LogError(ctx, "failed", "err", err)
		return err
	}

	goaclient.HandleResponse(c.Client, resp, cmd.PrettyPrint)
	return nil
}
```

The following line causes a compilation error because the `contentType` isn't expected by the `c.CreateTracker()` function:

```go
	resp, err := c.CreateTracker(ctx, path, &payload, cmd.ContentType)
```

The implementation (also generated) of `CreateTracker` looks like the following code. Notice that is has no `contentType` parameter

**client/tracker.go**
```go
// Add new tracker configuration.
func (c *Client) CreateTracker(ctx context.Context, path string, payload *CreateTrackerAlternatePayload) (*http.Response, error) {
	req, err := c.NewCreateTrackerRequest(ctx, path, payload)
	if err != nil {
		return nil, err
	}
	return c.Client.Do(ctx, req)
}
```

The design for this resource looks like this (shortened):

```go
var _ = a.Resource("tracker", func() {
	a.BasePath("/trackers")
//...
	a.Action("create", func() {
		a.Routing(
			a.POST(""),
		)
		a.Description("Add new tracker configuration.")
		a.Payload(CreateTrackerAlternatePayload)
		a.Response(d.Created, "/trackers/.*", func() {
			a.Media(Tracker)
		})
		a.Response(d.BadRequest, func() {
			a.Media(d.ErrorMedia)
		})
		a.Response(d.InternalServerError)
		a.Response(d.NotFound)
	})
//...
})
```
No, I wonder if this has something to do with these changes because they deal with contentType:

4efae3cfe69f54efd4f8cb6bc2999a7e7ff8a6b2
dd32788884dd83b460868092b726421c13b95efd
c72c2cdb2afe2d6a256de22f10d9de5cd563efd2

Any help on this issue is highly appreciated.

kwk:
In issue #908 I noticed that the CLI command still wants to pass in a `contentType` into a generated function that no longer accepts a `contentType` by default, only if the API has more than one type defined using `Consumes()`. I've applied the same logic from 4efae3cfe69f54efd4f8cb6bc2999a7e7ff8a6b2 to the `cli_generator.go` file.

Fixes #908

### DIFF
--- a/goagen/gen_client/cli_generator.go
+++ b/goagen/gen_client/cli_generator.go
@@ -74,6 +74,7 @@ func (g *Generator) generateMain(mainFile string, clientPkg, cliPkg string, func
 		HasBasicAuthSigners bool
 		HasAPIKeySigners    bool
 		HasTokenSigners     bool
+		HasMultiContent     bool
 	}{
 		API:                 g.API,
 		Version:             version,
@@ -82,6 +83,7 @@ func (g *Generator) generateMain(mainFile string, clientPkg, cliPkg string, func
 		HasBasicAuthSigners: hasBasicAuthSigners,
 		HasAPIKeySigners:    hasAPIKeySigners,
 		HasTokenSigners:     hasTokenSigners,
+		HasMultiContent:     len(g.API.Consumes) > 1,
 	}
 	if err := file.ExecuteTemplate("main", mainTmpl, funcs, data); err != nil {
 		return err
@@ -819,7 +821,7 @@ func (cmd *{{ $cmdName }}) Run(c *{{ .Package }}.Client, args []string) error {
 	resp, err := c.{{ goify (printf "%s%s" .Action.Name (title .Resource.Name)) true }}(ctx, path{{ if .Action.Payload }}, {{/*
 	*/}}{{ if or .Action.Payload.Type.IsObject .Action.Payload.IsPrimitive }}&{{ end }}payload{{ else }}{{ end }}{{/*
 	*/}}{{ $params := joinNames true .Action.QueryParams .Action.Headers }}{{ if $params }}, {{ format $params $specialTypeResult.Temps }}{{ end }}{{/*
-	*/}}{{ if .Action.Payload }}, cmd.ContentType{{ end }})
+	*/}}{{ if .Action.Payload }}{{ if .HasMultiContent }}, cmd.ContentType{{ end }}{{ end }})
 	if err != nil {
 		goa.LogError(ctx, "failed", "err", err)
 		return err


raphael:
ugh, thank you! Looks like the integration test failed though.

raphael:
Sorry for all the trouble with that change... PR #913 and #914 should fix the issue you reported in #908

kwk:
@raphael Thank you for fixing it. I wasn't able to reproduce the error that travis had.

PR #922, opened 2016-11-24
Avoid generating call to Validate when attribute type does not define the method

### LINKED ISSUE #918: User types without required field don't have a Validate() method
Now, while this may sound like a normal thing to do, the generated code on the other hand will try to call a `Validate` method on the user type struct.

Consider this media and user type without any required attribute:

```go
// JSONAPIErrors is an array of JSONAPI error objects
var JSONAPIErrors = a.MediaType("application/vnd.jsonapierrors+json", func() {
	a.UseTrait("jsonapi-media-type")
	a.TypeName("JSONAPIErrors")
	a.Description(``)
	a.Attributes(func() {
		a.Attribute("errors", a.ArrayOf(JSONAPIError))
		a.Required("errors")
	})
	a.View("default", func() {
		a.Attribute("errors")
		a.Required("errors")
	})
})

var JSONAPIError = a.Type("JSONAPIError", func() {
	a.Description(`Error objects provide additional information about problems encountered while
performing an operation. Error objects MUST be returned as an array keyed by errors in the
top level of a JSON API document.

See. also http://jsonapi.org/format/#error-objects.`)

	a.Attribute("id", d.String, "a unique identifier for this particular occurrence of the problem.")
	a.Attribute("links", a.HashOf(d.String, JSONAPILink), `a links object containing the following members:
* about: a link that leads to further details about this particular occurrence of the problem.`)
	a.Attribute("status", d.String, "the HTTP status code applicable to this problem, expressed as a string value.")
	a.Attribute("code", d.String, "an application-specific error code, expressed as a string value.")
	a.Attribute("title", d.String, `a short, human-readable summary of the problem that SHOULD NOT
change from occurrence to occurrence of the problem, except for purposes of localization.`)
	a.Attribute("detail", d.String, `a human-readable explanation specific to this occurrence of the problem.
Like title, this field’s value can be localized.`)
	a.Attribute("source", a.HashOf(d.String, d.Any), `an object containing references to the source of the error,
optionally including any of the following members

* pointer: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. "/data" for a primary data object,
           or "/data/attributes/title" for a specific attribute].
* parameter: a string indicating which URI query parameter caused the error.`)
	a.Attribute("meta", a.HashOf(d.String, d.Any), "a meta object containing non-standard meta-information about the error")
})
```

A generated `Validate` method on the media type currently looks like this:

```go
// Validate validates the JSONAPIErrors media type instance.
func (mt *JSONAPIErrors) Validate() (err error) {
	if mt.Errors == nil {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "errors"))
	}

	for _, e := range mt.Errors {
		if err2 := e.Validate(); err2 != nil {
			err = goa.MergeErrors(err, err2)
		}
	}
	return
}
```

As you can see, this line tries to call `Validate` on each of the embedded error objects: 

```go
if err2 := e.Validate(); err2 != nil {
```

It results in this error:

```
app/media_types.go:106: e.Validate undefined (type *JSONAPIError has no field or method Validate)
```

One could fix the caller or simply always generate a stub validation method like the following whenever there are no required fields:

```go
func (mt *JSONAPIError) Validate() (err error) {
    return nil
}
```

What do you think?

raphael:
Fix #918

### DIFF
--- a/goagen/codegen/validation.go
+++ b/goagen/codegen/validation.go
@@ -139,15 +139,7 @@ func (v *Validator) recurse(att *design.AttributeDefinition, nonzero, required, 
 			buf.WriteString(validation)
 			first = false
 		}
-		var val string
-		if _, ok := a.ElemType.Type.(*design.UserTypeDefinition); ok {
-			val = RunTemplate(v.userValT, map[string]interface{}{
-				"depth":  depth + 1,
-				"target": "e",
-			})
-		} else {
-			val = v.Code(a.ElemType, true, false, false, "e", context+"[*]", depth+1, false)
-		}
+		val := v.Code(a.ElemType, true, false, false, "e", context+"[*]", depth+1, false)
 		if val != "" {
 			data := map[string]interface{}{
 				"elemType":   a.ElemType,
@@ -323,10 +315,15 @@ func validationsCode(validation *dslengine.ValidationDefinition, data map[string
 		}
 	}
 	if required := validation.Required; len(required) > 0 {
-		data["required"] = required
-		if val := RunTemplate(requiredValT, data); val != "" {
-			res = append(res, val)
+		var val string
+		for i, r := range required {
+			if i > 0 {
+				val += "\n"
+			}
+			data["required"] = r
+			val += RunTemplate(requiredValT, data)
 		}
+		res = append(res, val)
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


rootpd:
Hi @raphael. This PR broke generation of string validators for nested types - at least in the scenario I encountered. Consider following design:

```go
        Attribute("uuid", UUID)
	Attribute("series", ArrayOf(Serie))
	Required("uuid", "series")
})

var Serie = Type("Serie", func() {
	Attribute("id", Integer)
	Attribute("label", String)
	Attribute("color", String)
	Required("id", "label", "color")
})
```

This generates validation block which is correct for fields of type `Foo`, but causes compile error for fields of type `Serie` - `e.Label` and `e.Color` are pointers here.

```go
// Validate validates the lineChart type instance.
func (ut *foo) Validate() (err error) {
	if ut.UUID == nil {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "uuid"))
	}
	if ut.Series == nil {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "series"))
	}
	for _, e := range ut.Series {
		if e.Label == "" {
			err = goa.MergeErrors(err, goa.MissingAttributeError(`response.series[*]`, "label"))
		}
		if e.Color == "" {
			err = goa.MergeErrors(err, goa.MissingAttributeError(`response.series[*]`, "color"))
		}
	}
	return
}
```

The problem here is that validator for `ut.Series` fields is generated by calling `validator.Code` with last parameter `private` set to `false` opposing to expected `true` for the user type. See your [yesterday's commit](https://github.com/goadesign/goa/commit/3af8d22e2e6106d05928aef9b4d52a0efa8d8e81#diff-afe105acadb6b8af2545c4fd90eb2f81R142).

It looks like originally in my case it was executed via the IF branch, not the ELSE and this is the result it used to generate:

```go
// Validate validates the lineChart type instance.
func (ut *foo) Validate() (err error) {
	if ut.UUID == nil {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "uuid"))
	}
	if ut.Series == nil {
		err = goa.MergeErrors(err, goa.MissingAttributeError(`response`, "series"))
	}
	for _, e := range ut.Series {
		if err2 := e.Validate(); err2 != nil {
			err = goa.MergeErrors(err, err2)
		}
	}
	return
}
```

Thanks for checking.

rootpd:
On the other hand, when I used the original IF branch with the latest code, goagen generated `e.Validate()` call also for the field which didn't have it defined.

Use following design to check, whether also that scenario is being treated correctly:

```go
        Attribute("uuid", UUID)
        Attribute("y_axes", ArrayOf(YAxis))
	Attribute("series", ArrayOf(Serie))
	Required("uuid", "series", "y_axes")
})

var Serie = Type("Serie", func() {
	Attribute("id", Integer)
	Attribute("label", String)
	Attribute("color", String)
	Required("id", "label", "color")
})

var YAxis = Type("YAxis", func() {
	Attribute("id", Integer)
	Attribute("title", String)
	Required("id")
})
```

raphael:
whoops, thank you. My fix wasn't quite right. The generated code should still call `Validate` but only if it needs to. #925 and #926 should have the correct fix. Thank you!

PR #942, opened 2016-11-30
Pass extra arguments to generator

### LINKED ISSUE #934: goagen gen does not allow custom flags
For example: 

jsimonetti:
This change will allow arguments to goagen which are added after the '--' argument to be passed allong to the generator.

Only full flags (such as --flag=value) are passed, because these flags have not been parsed yet. Short flags (such as -f value) are not passed.
A boolean flag should be added in the form of --flag=true.

The generate command becomes `goagen gen -d <design> --pkg-path=<generator> -- --custom-arg=value`

Fixes #934

### DIFF
--- a/goagen/meta/generator.go
+++ b/goagen/meta/generator.go
@@ -33,6 +33,11 @@ type Generator struct {
 	// generator on the command line.
 	Flags map[string]string
 
+	// CustomFlags is a list of arguments given to the final generator that
+	// where after the '--' argument and thus where not pardes by goagen.
+	// They will be forwarded to the final generator without modification.
+	CustomFlags []string
+
 	// OutDir is the final output directory.
 	OutDir string
 
@@ -44,7 +49,7 @@ type Generator struct {
 
 // NewGenerator returns a meta generator that can run an actual Generator
 // given its factory method and command line flags.
-func NewGenerator(genfunc string, imports []*codegen.ImportSpec, flags map[string]string) (*Generator, error) {
+func NewGenerator(genfunc string, imports []*codegen.ImportSpec, flags map[string]string, customflags []string) (*Generator, error) {
 	var (
 		outDir, designPkgPath string
 		debug                 bool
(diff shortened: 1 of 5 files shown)


jsimonetti:
For clarification:

The generate command becomes `goagen gen -d <design> --pkg-path=<generator> -- --custom-arg=value`

raphael:
Thank you for the PR! It would be nice to avoid the custom flag parsing... Could we instead take what comes after the `--` and plop it verbatim at the end of the command line we build for the custom generator? this way custom commands, shorthand flags etc. would just work.

jsimonetti:
Absolutely. I Just wanted to use the existing Flags map in the meta.Generator struct to keep the changes limited to main.
We can add a `CustomFlags []string` property to meta.Generator, which can hold everything after --

raphael:
Great! Why a slice though? Shouldn't it just be a string? (Maybe I should look at what cobra does to understand better...)

jsimonetti:
Basically a slice only because this is the type cobra passes along.
It makes sence in a way because `os.Args` is a slice too. So if we would want to do extra parsing in `meta.Generator` in the future we could without extra splitting. If you prefer a string thats fine by me too.
I suppose it depends on if you want the `strings.Join` to happen in main or somewhere in `meta.Generator`, either wat, I have no oppinion.

raphael:
Got it, a slice makes sense then, thanks for the clarification!

jsimonetti:
If the final result of this is ok, I will squash everything together in one commit and resubmit the PR, just to make everything more readable.

goagen/meta/generator.go
```diff
@@ -33,6 +33,11 @@ type Generator struct {
 	// generator on the command line.
 	Flags map[string]string
 
+	// CustomFlags is a list of arguments given to the final generator that
+	// where after the '--' argument and thus where not pardes by goagen.
```
raphael: There are a few typos in that comment, maybe replace with:
```
// CustomFlags is the list of arguments that appear after the -- separator.
// These arguments are appended verbatim to the final generator command line.
```

jsimonetti:
New PR947

PR #946, opened 2016-12-02
Make sure to escape path parameters when building request URL

### LINKED ISSUE #944: CLI should perform URL encoding
For example, I attempted to to pass a parameter such as 
`--ip_address 192.168.1.1/24`, e.g. an IP address in CIDR notation.
The framework of course generates a status 404 not found since it sees the slash as an additional path component. The slash should be encoded.

raphael:
in the CLI code. Query string parameters are already escaped properly.

Fix #944


mark-lb:
@raphael but shouldn't this be matched by unescaping params on the server side?
Thanks

raphael:
Hmm I'm not sure the server side should unescape all path parameters always. This is a high price to pay performance wise for what is really a corner case at the end. It's OK for the CLI to always do it but it should be the responsibility of the controller implementation to unescape when it knows it's needed.

PR #947, opened 2016-12-02
Pass arguments to final generator

### LINKED ISSUE #934: goagen gen does not allow custom flags

jsimonetti:
This change will allow arguments to goagen which are add after the '--' argument to be passed allong to the generator.

The generate command becomes goagen gen -d <design> --pkg-path=<generator> -- --custom-arg=value

Fixes #934

### DIFF
--- a/goagen/meta/generator_test.go
+++ b/goagen/meta/generator_test.go
@@ -24,6 +24,7 @@ var _ = Describe("Run", func() {
 	var outputDir string
 	var designPkgPath, setDesignPkgPath string
 	var designPackageSource string
+	var customFlags []string
 
 	var m *meta.Generator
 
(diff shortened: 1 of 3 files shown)


goagen/meta/generator_test.go
```diff
@@ -270,4 +288,18 @@ func Generate() ([]string, error) {
...
+import "fmt"
+import "os"
+
+func Generate() ([]string, error) {
+	for _, arg := range os.Args {
+		if arg == "{{.}}" {
+			return nil, nil
+		}
+	}
+	return nil, fmt.Errorf("no flag --custom=arg found")
```
raphael: `--custom=arg` should probably be `{{.}}`

raphael:
Thank you, the PR looks great. Left two nitpick comments, I'll merge once you address these.

raphael:
Thank you! Do you think you could also make a PR against the v1 branch?

PR #948, opened 2016-12-03
Pass arguments to final generator

### LINKED ISSUE #934: goagen gen does not allow custom flags

jsimonetti:
This change will allow arguments to goagen which are add after the '--' argument to be passed allong to the generator.

The generate command becomes goagen gen -d --pkg-path= -- --custom-arg=value

Fixes #934


raphael:
Thank you!

PR #953, opened 2016-12-07
Suppress diffs of generated codes on GitHub

tchssk:
This pull request makes `goagen` compatible to diff suppression of [Linguist](https://github.com/github/linguist).

https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L44-L81
https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L258-L263

Generated `app` codes by goagen will be suppressed in diffs on GitHub.


raphael:
This is great! thank you. Do you think you could backport this to the `v1` branch as well?

raphael:
ugh I broke this with #1160 - will fix.

tchssk:
Oh thank you.

PR #962, opened 2016-12-13
Fix the generation of length validation for strings that contain multi-byte UTF-8 characters.

ikawaha:
Checking the difference of v1.1.0, I noticed that #740 is not backporting to the v1 branch.

### DIFF
--- a/goagen/codegen/validation.go
+++ b/goagen/codegen/validation.go
@@ -260,6 +260,7 @@ func ValidationChecker(att *design.AttributeDefinition, nonzero, required, hasDe
 		"context":   context,
 		"target":    target,
 		"targetVal": t,
+		"string":    att.Type.Name() == "string",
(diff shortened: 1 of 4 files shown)


goagen/codegen/validation.go
```diff
@@ -260,6 +260,7 @@ func ValidationChecker(att *design.AttributeDefinition, nonzero, required, hasDe
 		"context":   context,
 		"target":    target,
 		"targetVal": t,
+		"string":    att.Type.Name() == "string",
```
raphael: Small thing: do you think you could use `att.Type.Kind() == design.StringKind` instead ? Thank you!

raphael:
Thank you! I left a small comment in the code.

ikawaha:
Thank you for your comment !

PR #963, opened 2016-12-13
Fix string type checking.

ikawaha:
see https://github.com/goadesign/goa/pull/962#discussion_r92101541


raphael:
Awesome, thank you!

Issue #964, opened 2016-12-13
There are unmerged features/fixes that were announced in v1.1.0.


ikawaha:
I checked features/fixes that were [announced in v1.1.0](https://goa.design/blog/003-v1_1_0/).
The following PRs are probably not backported to the v1 branch.

* [ ] [Added the ability to dynamically add/remove keys](https://github.com/goadesign/goa/pull/818) 
* [ ] [Refactor JWT key resolver](https://github.com/goadesign/goa/pull/832)
* [ ] [Fix setting a default value to datetime](https://github.com/goadesign/goa/pull/758)
* [ ] [Add an interface instead of using testing.T](https://github.com/goadesign/goa/pull/733)
* [ ] [address issue #738 by creating a title for the test ](https://github.com/goadesign/goa/pull/780)
* [ ] [removed unneeded check which would make an unsafe concurrent modification](https://github.com/goadesign/goa/pull/736)
* [ ] [Modify metrics so that names are normalized to legal chars](https://github.com/goadesign/goa/pull/771)

Check List
---
### Programmatic Code Generation
* [x] https://github.com/goadesign/goa/pull/686
### Swagger Support Improvements
* [x] https://github.com/goadesign/goa/pull/779
* [x] https://github.com/goadesign/goa/pull/788
* [x] https://github.com/goadesign/goa/pull/790
* [x] https://github.com/goadesign/goa/pull/863
* [x] https://github.com/goadesign/goa/pull/858
* [x] https://github.com/goadesign/goa/pull/936
### Query String Array Parameters
* [x] https://github.com/goadesign/goa/pull/684
### Security Improvements
* [ ] https://github.com/goadesign/goa/pull/818 Added the ability to dynamically add/remove keys from the jwt security middleware after creation.
* [ ] https://github.com/goadesign/goa/pull/832 Refactor JWT key resolver
* [x] https://github.com/goadesign/goa/pull/823 
### CORS Support Improvements
* [x] https://github.com/goadesign/goa/pull/723
* [x] https://github.com/goadesign/goa/pull/717
* [x] https://github.com/goadesign/goa/pull/763
### Support for Modifying the goa Error Response Content-Type
* [x] https://github.com/goadesign/goa/pull/697
### Support for UUID parameters on the CLI
* [x] https://github.com/goadesign/goa/pull/701
### Logging
* [x] https://github.com/goadesign/goa/774
* [x] https://github.com/goadesign/goa/pull/792
### DSL
[...]

PR #758, opened 2016-09-03
Fix setting a default value to datetime

ikawaha:
issue #726

Issue #738, opened 2016-08-24
Generated test helpers missing "DO NOT MODIFY" comment heading


chriskaukis:
The generated code for the test helpers are missing the common header for all files that overridden by the generator.

stuartweir:
hey @chriskaukis where in the code is this located?

raphael:
Thanks for the fix!

PR #965, opened 2016-12-13
Backporting unmerged features/fixes to v1 branch.

ikawaha:
issue #964

* ~~[Added the ability to dynamically add/remove keys](https://github.com/goadesign/goa/pull/818)~~
*  ~~[Refactor JWT key resolver](https://github.com/goadesign/goa/pull/832)~~
* [x] [Fix setting a default value to datetime](https://github.com/goadesign/goa/pull/758)
* [x] [Add an interface instead of using testing.T](https://github.com/goadesign/goa/pull/733)
* [x] [address issue #738 by creating a title for the test ](https://github.com/goadesign/goa/pull/780)
* [x] [removed unneeded check which would make an unsafe concurrent modification](https://github.com/goadesign/goa/pull/736)
* [x] [Modify metrics so that names are normalized to legal chars](https://github.com/goadesign/goa/pull/771)


raphael:
Thank you for catching this! I now realize I didn't perform the diff correctly when I wrote the blog post :(

I went through the list and the only one that seems problematic is the `JWT`  interface change. This being a `v1` release I would like to keep the interface backwards compatible. Do you think you could remove the first two commits from this PR? I'll update the blog post to remove this entry. Thank you again, much appreciated!

ikawaha:
I remove the two `JWT` commits.

raphael:
I have retagged `v1.1.0` to include these commits. Thank you again.

PR #972, opened 2016-12-14
context values (user types) populated with lookup functions

michaelboke:
I created a way to make context value lookups possible.
Most of the time i am repeating some part of controller logic by looking up some models from the repositories. I could do it with generic middleware but and store the value in the context. But that doesn't seems to be nice because most of the time my action depends on the lookup values. And also we have a beautiful context structure to store these references. Now with some extra DSL we can create a lookup function and execute it from the context creation.

for example we can create a Lookup template in the api definition like this
```
var _ = API("test", func() {
	//....
	Lookup("Account", Account, func() { //name of the lookup and the usertype to populate in the context
		Param("id", UUID) // the arguments the lookup needs to have
	})
})
```

This will create a lookup interface we can later on create and give as an argument to the mount function
```
type LookupAccount interface {
	LookupAccount(id uuid.UUID) (*Account, error)
}
```

The DSL for the action, you could also define it on the resource.  When stacking routes (parent) the canonical action path also inherits all the lookups in its route. 
```
Action("show", func () {
	Routing(GET("/:account_id"))
	Description("returns a account")
	Params(func () {
		Param("account_id", UUID, "Account ID")
	})
	UseLookup("Account", func (){
		MustResolve()	 //this will indicate the lookup must succeed with a non nil value, otherwise the request will cancel with a 404
		MapRouteParam("account_id", "id") // if the route params do not match 1 on 1  on the lookup arguments we could remap them this way
	})

	Response(OK, Account)
	Response(NotFound)
})
```

this will add a the following attribute to the contexts.go
```
type ShowAccountContext struct {
	context.Context
	*goa.ResponseData
	*goa.RequestData
	AccountID     uuid.UUID
	Account       *Account // << populated by the lookup
}
```

the validation and invocation of the lookup in the controllers.go
```
func MountAccountController(service *goa.Service, ctrl CustomersController, lookupAccount LookupAccount) {
	initService(service)
	var h goa.Handler

	h = func(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
		// Check if there was an error loading the request
		//...

		// Lookup the context value Account
		tmp94, err := lookupAccount.LookupAccount(rctx.AccountID)
		if err != nil {
			return goa.ErrInternal("lookup of account failed with error", "error", err, "lookup", "account")
		}

		//this check is only added if MustResolve() is used in UseLookup
		if tmp94 == nil {
			return goa.ErrNotFound("account is not found by id", "lookup", "account")
		}
		rctx.Account = tmp94

		return ctrl.Show(rctx)
	}
	service.Mux.Handle("GET", "/api/stalling/:stalling_id/customers/:customer_id", ctrl.MuxHandler("Show", h, nil))
	//....
}
```

Also the initial creation of main.go the lookups are auto inserted into the Mount functions and a stub lookups.go is created.
main.go
```
	lookupAccount := NewLookupAccount()

	// Mount "account" controller
	c := NewAccountController(service)
	app.MountAccountController(service, c, lookupAccount)
```

and the stub lookups.go
```
package main

import (
	uuid "github.com/satori/go.uuid"
	"stalling/goa2/app"
)

type lookupAccount struct{}

// NewLookupAccount creates a new account lookup
func NewLookupAccount() app.LookupAccount {
	return &lookupAccount{}
}

// LookupAccount handles the lookup of the context value Account
func (l *lookupAccount) LookupAccount(id uuid.UUID) (*app.Account, error) {
	// Put your lookup logic here
	return nil, nil
}
```

Phew i hope i didn't forget anything yet.
What do you think of this proposal, is it something that could be a part of GOA.

I know tests need to be done and i would like to know how to use the validation to ensure the DSL correct.

### DIFF
--- a/design/apidsl/api.go
+++ b/design/apidsl/api.go
@@ -68,6 +68,9 @@ import (
 //				Required("header")
 //			})
 //		})
+//		Lookup("Account", AccountType, func() {	// Creates a new lookup template function
+//			Attribute("id", UUID)		// The input attributes
+//		})
 //	}
 //
 func API(name string, dsl func()) *design.APIDefinition {
@@ -475,7 +478,9 @@ func setupResponseTemplate(a *design.APIDefinition, name string, p interface{}) 
 	} else if tmpl, ok := p.(func(...string)); ok {
 		t := func(params ...string) *design.ResponseDefinition {
 			r := &design.ResponseDefinition{Name: name}
-			dslengine.Execute(func() { tmpl(params...) }, r)
+			dslengine.Execute(func() {
+				tmpl(params...)
+			}, r)
 			return r
 		}
 		a.ResponseTemplates[name] = &design.ResponseTemplateDefinition{
@@ -512,7 +517,9 @@ func setupResponseTemplate(a *design.APIDefinition, name string, p interface{}) 
 				// append input arguments
 				in[i] = reflect.ValueOf(params[i])
 			}
-			dslengine.Execute(func() { val.Call(in) }, r)
+			dslengine.Execute(func() {
+				val.Call(in)
+			}, r)
 			return r
 		}
 		a.ResponseTemplates[name] = &design.ResponseTemplateDefinition{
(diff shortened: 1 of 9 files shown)


raphael:
Closing this PR as per our discussion on Slack. This feels like something that would be better addressed via plugins rather than built-in DSL. The `v2` code generation algorithm will allow for this type of customization.

PR #974, opened 2016-12-14
Use go/build to determine if package exists for validations.

### LINKED ISSUE #973: Generating in vendored without source in gopath fails.
## The Issue
When generating using the [instructions here](https://goa.design/design/vendoring/#code-based-generation-and-vendoring) to vendor goa. It will fail if goa doesn't exist in your GOPATH and only in your vendor directory.

## Version of goa
`v1.1.0` and `master`

## Output:
```
$ go run gen/main.go
encoding for application/x-www-form-urlencoded: invalid Go package path "github.com/goadesign/goa/encoding/form"
encoding for application/x-www-form-urlencoded: invalid Go package path "github.com/goadesign/goa/encoding/form"exit status 1
main.go:1: running "go": exit status 1
```

## Probable Cause:
It looks like it's caused by the fact encoding doesn't check the vendor directory in [validation.go#L281-L291](https://github.com/goadesign/goa/blob/v1.1.0/design/validation.go#L281-L291)

## Possible solutions:
Use the `go/build` package to get information on the package. If it returns an error, we can safely assume it doesn't exist.
```golang
	dir, err := os.Getwd()
	if err != nil {
		panic(err)
	}
	_, err := build.Default.Import("github.com/goadesign/goa", dir, build.FindOnly))
	if err != nil {
		panic(fmt.Sprintf("Not found!"))
	}
```

Another solution is to just add the vendor directories recursively to the `gopaths` variable.

brycereitano:
Closes #973

### DIFF
--- a/design/validation.go
+++ b/design/validation.go
@@ -2,6 +2,7 @@ package design
 
 import (
 	"fmt"
+	"go/build"
 	"mime"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


brycereitano:
Closes #974

Issue #961, opened 2016-12-12
gen_client generating invalid Path when DateTime param is used in GET


rootpd:
Hi,

recently there was a bug with invalid generation of DateTime values in for `client` payload (see #924). Today I stumbled upon a similar scenario, but for GET parameters, which we didn't cover last time.

It looks like a `pathTemplate` [function](https://github.com/goadesign/goa/blob/v1/goagen/gen_client/generator.go#L793) is replacing all URL params with `%v` naively. It should use `time.RFC3339` formar for DateTimes and it probably should use similar switch-case statement as it was when [generating payload](https://github.com/goadesign/goa/blob/v1/goagen/gen_client/generator.go#L734) (can't tell for sure right now).

Right now it's causing invalid path being generated when DateTime param is used in GET and the server is returning `400 Bad Request`.

Thanks.

rootpd:
Technically it might be enough to alter `pathParamNames` function, so it returns `myDateVar.Format(time.RFC3339)` instead of `myDateVar` for time variables (and other kinds accordingly if necessary).

raphael:
Thank you for the PRs! I'll merge them for now. I think this requires a more fundamental fix though because the problem exists for other types as well. Ideally the code should use the `toString` function to generate the code that transform the values into strings before calling `Sprintf`.

rootpd:
Good point, I was afraid it would be the case :). I might have some time tomorrow to play with it in case you're busy but it takes some time to get into it. Anyway, thanks for the partial-fix merge.

raphael:
Sounds good, I'll leave the issue opened until the proper fix is made.

raphael:
This was fixed in #976 and #977

PR #976, opened 2016-12-15
Generating path for client is transforming all relevant types to string first.

rootpd:
I have a feeling that it could be done with a bit less duplicity within `initParams`. Opened to any ideas (or just push a change) :).

PR is fixing issue #961, which was partially fixed yesterday.


raphael:
Wow thanks for all the work on this. First pass through it looks good.

PR #977, opened 2016-12-15
Generating path for client is transforming all relevant types to string first.

rootpd:
I have a feeling that it could be done with a bit less duplicity within `initParams`. Opened to any ideas (or just push a change) :).

PR is fixing issue #961, which was partially fixed yesterday.

### DIFF
--- a/goagen/gen_client/generator.go
+++ b/goagen/gen_client/generator.go
@@ -134,7 +134,6 @@ func (g *Generator) Generate() (_ []string, err error) {
 			"join":               join,
 			"joinStrings":        strings.Join,
 			"multiComment":       multiComment,
-			"pathParamNames":     pathParamNames,
 			"pathParams":         pathParams,
 			"pathTemplate":       pathTemplate,
 			"signerType":         signerType,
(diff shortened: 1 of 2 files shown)


rootpd:
I see the failing Travis. Going for christmas party, will take care tomorrow (central europe time). :)

goagen/gen_client/generator.go
```diff
@@ -334,13 +333,50 @@ func (g *Generator) generateResourceClient(pkgDir string, res *design.ResourceDe
 				}
 			}
 		}
+		initParams := func(att *design.AttributeDefinition) []*paramData {
```
raphael: Looks like the same logic as the one used in `generateActionClient` - maybe it's time to make this its own function and call it from both places?

goagen/gen_client/generator.go
```diff
@@ -873,11 +889,16 @@ func (c *Client) {{ $funcName }}(resp *http.Response) ({{ decodegotyperef . .All
 `
 
 	pathTmpl = `{{ $funcName := printf "%sPath%s" (goify (printf "%s%s" .Route.Parent.Name (title .Route.Parent.Parent.Name)) true) ((or (and .Index (add .Index 1)) "") | printf "%v") }}{{/*
-*/}}{{ with .Route }}// {{ $funcName }} computes a request path to the {{ .Parent.Name }} action of {{ .Parent.Parent.Name }}.
-func {{ $funcName }}({{ pathParams . }}) string {
-	return fmt.Sprintf({{ printf "%q" (pathTemplate .) }}, {{ pathParamNames . }})
+*/}}// {{ $funcName }} computes a request path to the {{ .Route.Parent.Name }} action of {{ .Route.Parent.Parent.Name }}.
+func {{ $funcName }}({{ pathParams .Route }}) string {
+	var params []interface{}
```
raphael: It would be good to avoid having to create a slice for this. This method could be called multiple times per request being handled so the extra allocations / execution time could add up. Granted it is more complicated but it's possible and it seems better to take the time to do this right.

I'm thinking instead of using a temporary variable the code could use a known prefix, e.g. "param" and append an index to it. So in the loop it would be `{{ toString "p" (printf "param%d" $i) $param.Attribute }}` where `$i` is declared in the `range` expression (`{{ range $i, $param := .Params }}`). The code rendering the `fmt.Sprintf` would then also use the same range to build the arguments.

raphael:
Thank you for doing this! looks great. I suggested a couple of tweaks in the comments. Have fun at the party :)

rootpd:
Thanks for the hints and patience! This time it even looks greener than the last time :).

rootpd:
Found one more bug with the order of generated params, expect one more commit.

raphael:
Thank you for the great PR!

PR #980, opened 2016-12-18
Generate "Generated By" header for client code.

brycereitano:
I will create another branch to put onto master after this is reviewed.


raphael:
Thank you!

raphael:
Can you also send a PR to `master`? thanks!

raphael:
Good old Windows keeping us on our toes:

https://ci.appveyor.com/project/RaphaelSimon/goa-oqtis/build/1528

Do you think you can tweak the test so that it uses the correct path separator? (`filepath.Join` should do what you need). Thank you!

PR #981, opened 2016-12-20
support_custom_request_id_in_client

smessier:
@raphael added client.SetContextRequestID() to allow for sending a custom "X-Request-Id" header with client requests
added client/*_test.go files


raphael:
Thank you! could you also send a PR against the `master` branch?

Issue #986, opened 2016-12-22
Code based generation seems to be broken


ottogiron:
Hi,  I've been using code based generation for a while,  but a change in "goagen/gen_app/generator.go" has broken this option.

When runnig ***go run gen/main.go*** I get:

```
swagger/swagger.json
swagger/swagger.yaml
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x9e136]

goroutine 1 [running]:
panic(0x406780, 0xc42000c1f0)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
goa-cellar/vendor/github.com/goadesign/goa/goagen/codegen.(*Validator).recurse(0x0, 0xc420233280, 0xc420000001, 0x47430c, 0x1, 0xc4202d069a, 0x6, 0x2, 0xc4202d0600, 0x6)
        /Users/ogiron/go/src/goa-cellar/vendor/github.com/goadesign/goa/goagen/codegen/validation.go:102 +0xf26
goa-cellar/vendor/github.com/goadesign/goa/goagen/codegen.(*Validator).Code(0x0, 0xc420233280, 0x1, 0x47430c, 0x1, 0xc4202d069a, 0x6, 0x2, 0x0, 0x0, ...)
        /Users/ogiron/go/src/goa-cellar/vendor/github.com/goadesign/goa/goagen/codegen/validation.go:89 +0x9e
goa-cellar/vendor/github.com/goadesign/goa/goagen/codegen.(*Validator).recurse(0x0, 0xc420233300, 0xc420000000, 0x47e52f, 0x7, 0x474b0b, 0x3, 0x1, 0xc420255600, 0xc4203de000)
```

A test using cellar and vendoring goa with glide: 

***glide.lock***

```
name: github.com/goadesign/goa
 version: 7fc6b55040a7f8a800500c44c547692a0d7aed77
```

```go 
package main

import (
	_ "goa-cellar/design"

	"github.com/goadesign/goa/design"
	"github.com/goadesign/goa/goagen/codegen"
	genapp "github.com/goadesign/goa/goagen/gen_app"
	genclient "github.com/goadesign/goa/goagen/gen_client"
	"github.com/goadesign/goa/goagen/gen_js"
	genmain "github.com/goadesign/goa/goagen/gen_main"
	genswagger "github.com/goadesign/goa/goagen/gen_swagger"
)

func main() {
[...]

PR #988, opened 2016-12-26
WIP: Add constructor method to generators

ottogiron:
fixes https://github.com/goadesign/goa/issues/986

Hi @raphael, 

I'm creating an early PR with a proposal of the possible API for constructors methods for generators. It is based on a short article by Rob Pike about functional options API's 

https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html.

Also explained by Dave Cheney here: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis 

### Available options (settable as a functional option)
gen_app/options.go
```
package genapp

//Option a generator option definition
type Option func(*Generator)

//API The API definition
func API(API *design.APIDefinition) Option {
	return func(g *Generator) {
		g.API = API
	}
}

//From here not showing the implementation.

//OutDir Path to output directory
func OutDir(outDir string) Option 

...
//Target Name of generated package
func Target(target string) Option 

...
//NoTest Whether to skip test generation
func NoTest(noTest bool) Option 
...
```

### Definition of constructor
gen_app/generator.go
```

//NewGenerator returns an initialized instance of an Application Generator
func NewGenerator(options ...Option) *Generator {
	g := &Generator{}

	g.validator = codegen.NewValidator()

        //applying options
	for _, option := range options {
		option(g)
	}

	return g
}
```

### Configuring and running a generator
eg. gen/main.go
```
func main() {
	codegen.ParseDSL()
	codegen.Run(
		genapp.NewGenerator(
				genapp.API(design.Design),
				genapp.OutDir("app"),
                                 // other options
			),
            ...)
}
```

I have used this pattern for libraries in my current company, it is flexible and it has worked well. 

Other options for implementing this might be an static set of parameters (maybe the number of options won't change), or an options struct. 

Let me know what do you think, I would be glad to implement the solution based on what fits the best for the current codebase.

Thanks,

Otto.


raphael:
Thank you for the pointer! it seems like the functional option pattern could work well here. One  question I'm wondering is whether each `gen_xxx` package should define its own option function type or whether it would be better to declare it once (e.g. in the `codegen` package). The package that defines the type could also define the common options to all the generators (`API`, `OutputDir` and `Target` I believe). Each `gen_xxx` package could then add its specific option functions.

ottogiron:
This specific pattern depends on the firm of the option type including the concrete struct that is going to be configured, since it has to have access to public and private properties of that struct.  I tried to think in ways of achieving what you mentioned, but I could not think in practical ways of doing it.

raphael:
Hmm I see, agree that using an interface would not be practical - probably better to repeat the option functions in each package. It seems this PR is ready to merge then? Do you think you could do the same to the other `gen_xxx` packages?

ottogiron:
Sure, I will.

raphael:
Great! Could you also make PRs to backport the changes to the `v1` branch? Thank you!

PR #998, opened 2017-01-01
Fix minor typo in string format

bits01:


raphael:
Thank you! Do you think you could send a PR to backport the fix to the `v1` branch?

PR #1001, opened 2017-01-01
Adds constructor method and defines options for all the generators v1 backport

ottogiron:


raphael:
Could I coerce you into adding or updating a paragraph in the [docs](https://github.com/goadesign/goa.design/blob/master/content/design/vendoring.md) showing how to take advantage of the Option functions to write a code based generator?

ottogiron:
Sure.  Thank you for pointing out the right location, I was about to ask about that.

PR #1003, opened 2017-01-10
Add Tracer and AWS X-Ray middlewares.

raphael:
The Tracer middleware retrieves the trace ID and parent span ID from the
incoming request if present and stores the in the context. The middleware
provides a goa Doer wrapper function that extract the values from the context
and sets them in the outgoing request.

The middleware constructor accepts a percentage as argument which specifies the
sampling value. This value is used when the incoming request does *not* include
tracing headers to determine whether a new trace ID should be computed and set
in the context.

The xray middleware leverages the values set in the context by the Tracer
middleware to report tracing information to the AWS X-Ray service. The package
also makes it possible to attach further information to the traces such as
annotations and metadata. The middleware also reports errors to the service.

The xray package also provides a WrapClient function for wrapping an instance
of http.Client to trace requests made to external services.

### DIFF
--- a/middleware/context.go
+++ b/middleware/context.go
@@ -4,5 +4,12 @@ package middleware
 // It is private to avoid possible collisions with keys used by other packages.
 type middlewareKey int
 
-// ReqIDKey is the context key used by the RequestID middleware to store the request ID value.
-const reqIDKey middlewareKey = 1
+const (
+	// ReqIDKey is the context key used by the RequestID middleware to store the request ID value.
+	reqIDKey middlewareKey = iota + 1
+
+	// Keys used to record trace in context.
+	traceKey
+	spanKey
+	parentSpanKey
+)
--- a/middleware/xray/segment.go
+++ b/middleware/xray/segment.go
@@ -1,0 +1,313 @@
+package xray
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"sync"
+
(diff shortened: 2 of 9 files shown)


middleware/context.go
```diff
@@ -4,5 +4,12 @@ package middleware
...
-// ReqIDKey is the context key used by the RequestID middleware to store the request ID value.
-const reqIDKey middlewareKey = 1
+const (
+	// ReqIDKey is the context key used by the RequestID middleware to store the request ID value.
+	reqIDKey middlewareKey = iota + 1
+
+	// Keys used to record trace in context.
+	traceKey
+	spanKey
+	parentSpanKey
```
jim-slattery-rs: I love this trick to use a private type for the keys.

middleware/xray/segment.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// AddBoolMetadata adds a key-value pair that can be queried by AWS X-Ray.
+func (s *Segment) AddBoolMetadata(key string, value bool) {
+	s.addMetadata(key, value)
+}
+
+// addMetadata adds a key-value pair that can be queried by AWS X-Ray.
+// AWS X-Ray only supports annotations of type string, integer or boolean.
+func (s *Segment) addMetadata(key string, value interface{}) {
```
jim-slattery-rs: should this method Lock the mutex?

raphael: Hmm yes, good catch!

Issue #1006, opened 2017-01-14
CLI panic without requied params


riku179:
CLI should show what happen, when requied params are missing.

In my app, `is_member` is defined requied param in design.go
```go
Params(func() {
    Param("is_member", Boolean, "Required param")
    Required("is_member")
})
```

If CLI is executed without `is_member` , it panic.
```
./foo-cli signin jwt
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4d8ef1]

goroutine 1 [running]:
panic(0x760f60, 0xc4200140f0)......
```

This is because,  *tmp13 is nil in `tool/cli/commands.go` ,
```go
var tmp13 *bool
if cmd.IsMember != "" {
	var err error
	tmp13, err = boolVal(cmd.IsMember)
	if err != nil {
		goa.LogError(ctx, "failed to parse flag into *bool value", "flag", "--is_member", "err", err)
		return err
	}
}

resp, err := c.SigninJWT(ctx, path, *tmp13)   // <- panic here
```

Generated code should be like following.
```go
var tmp13 *bool
if cmd.IsMember != "" {
	var err error
	tmp13, err = boolVal(cmd.IsMember)
	if err != nil {
		goa.LogError(ctx, "failed to parse flag into *bool value", "flag", "--is_member", "err", err)
		return err
	}
} else {
     //	warn missing params and return error
}
resp, err := c.SigninJWT(ctx, path, *tmp13)
```

raphael:
Thank you for the detailed report!

PR #1007, opened 2017-01-15
Correctly check for nil values

raphael:
When calling client command with a pointer containing a required parameter.


raphael:
Fix #1006

PR #1018, opened 2017-01-29
Fix Validator nil reference exception

wardn:


raphael:
Thank you for the fix! I am wondering if it wouldn't be better to prevent the call to `recurse` in the first place rather than protecting against a nil target if `v` is `nil`. It seems like other things could be wrong in the caller so simply protecting here may hide issues.

wardn:
Yeah, I think Code() is all that's using it, but it seemed better to at least guard at the source in case other methods come along in the future that also use recurse.  Checks should probably be made upstream as well. I've only had a glance at the codebase so I'll leave it up to those in a better position to determine the correct checkpoints.  This at least gets me past the app gen nil reference issue I was experiencing.

raphael:
There's definitely a trade-off here but I would rather the execution clearly fails (panics) when a caller fails to provide a value than the execution finishes successfully but produces invalid code (because the value should not have been nil in the caller). So I'm going to close this and fix the caller. Thank you for the report!

raphael:
I actually don't understand how that could be `nil`. Validations are optional, yes, but here `v` is the validation code generator which should never be `nil`. Do you have a repro you could share?

wardn:
Here's a quick project to demonstrate...

https://github.com/wardn/nilvalidator

raphael:
Awesome, thank you!

raphael:
OK the problem is that you need to call `NewGenerator` instead of initializing the generator struct inline. This will initialize the `validator` field of the created generator properly. I'm wondering if you saw an example or documentation referring to creating the generator this way? If so we should fix that...

wardn:
Yeah I got it from https://github.com/goadesign/gorma-cellar/blob/master/gen/main.go

raphael:
I see, please see https://goa.design/design/vendoring/ for the latest / correct usage.

PR #1021, opened 2017-01-30
Payload function documentation

genez:
the type accepted by the Payload function is a *goa* type and *not* a Go type (struct)


raphael:
Closing this as per our conversation on Slack

PR #1022, opened 2017-01-31
ensure OutDir existence on main generation

wardn:


raphael:
Thank you! Can you please also make a PR for the `v1` branch.

Issue #658, opened 2016-07-29
Make it possible to provide default values for query string and header params


Tri125:
At the moment we can't specify a default value for params inside a resource or action.

bits01:
I assume URL params implies both path and query params.

Tri125:
I guess I should have called it Action/Resource Params. I'll edit to make it clearer.

raphael:
@bits01 setting default values on path parameters doesn't make a lot of sense because the mux enforces they have values. It might be good to fail actually if defaults are provided for those with a nice error message explaining this. So we're talking about query string parameters and header parameters.

PR #1024, opened 2017-02-01
Assign to default param values.

### LINKED ISSUE #754: Call Default() in Param() not passing default value to controller
Im specify
```go
Param("limit", Integer, func() {
	Description("Count rows")
	Minimum(20)
	Default(20)
})
```
But when am call GET /units without parameters, controller
```go
func (c *UnitController) List(ctx *app.ListUnitContext) error {
    // ctx.Limit equals 0 (not 20)
}
```

brycereitano:
This would close #754 (already closed) and half finish #658. I may set some time aside to cherry pick the changes for headers as well. I want to get this PR finalized and merged before that.

I made `codegen.printVals` public for use of printing the literal default values.
I added a lot of tests to make sure it was being handled properly everywhere. I may have gone a bit overboard.


raphael:
This is awesome, thank you!

PR #1030, opened 2017-02-02
Add an `ErrForbidden` error (#1029)

### LINKED ISSUE #1029: Add an `ErrForbidden` error
This kind of error would be returned when a user is not allowed to perform an operation. This is different from the `unauthorized` kind of response, which is related to authentication (ie, the user is not recognized and should login first). 

See discussion on [stackoverflow](http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses#6937030) for more details.

xcoulon:
Added an `ErrForbidden = NewErrorClass("forbidden", 403)`
error

Fixes #1029

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

### DIFF
--- a/error.go
+++ b/error.go
@@ -47,6 +47,9 @@ var (
 	// ErrUnauthorized is a generic unauthorized error.
 	ErrUnauthorized = NewErrorClass("unauthorized", 401)
 
+	// ErrForbidden is a generic forbidden error.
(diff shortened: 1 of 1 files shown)


raphael:
See comment in issue.

Issue #1035, opened 2017-02-07
Specify custom generated go filenames and cli command parameter name


hectorj2f:
We faced an issue on which the resource name have to differ from the generated filename and the name of the cli command argument.

The resource name is used as the filename prefix for the golang file, as defined here: (https://github.com/goadesign/goa/blob/master/goagen/gen_main/generator.go#L112). Moreover, the cli command argument seems to use the resource name as well:(https://sourcegraph.com/github.com/goadesign/goa@450e184b114536525efffa64e2b31733ee689d97/-/blob/goagen/gen_client/cli_generator.go#L837:27-837:29).

For instance, we want to define a command where the resource name can differ for the cli argument name, e.g:

```
var _ = a.Resource("obj_iterations", func() {
	a.Parent("obj")
	a.Action("list", func() {
		a.Routing(
			a.GET("obj"),
		)
		a.Description("List objs.")
		a.Response(d.OK, func() {
			a.Media(objList)
		})
		a.Response(d.InternalServerError, ...)
	})
```
We want to get a filename that uses underscores `obj_iterations.go` but that looks like that in our cli:
`cli create obj-iterations`. However we get the same cli parameter `cli create obj_iterations` which isn't what we want. Moreover, the use of underscores in cli apps isn't correct.

Is there a way of customizing this name for the cobra command ?

hectorj2f:
An easy solution could be to replace the `underscores` by `dashes` when using the `resource.Name` as cmd argument.

raphael:
Thanks for the detailed report! Replacing underscores with dashes for the CLI flags makes sense to me. It should probably done for all the flags though - not just resource names. Would you care to do the honor with a PR?

hectorj2f:
@raphael I made a PR with some changes, please have a look at it :).

goldcaddy77:
This can be closed since #1039 is merged, right?

raphael:
Yup, thank you!

PR #1039, opened 2017-02-07
gen_cli: Replace underscores by dashes for the generated cmd arguments

hectorj2f:
Related to this issues: https://github.com/goadesign/goa/issues/1035

@raphael Please, have a look! I made some initial changes that works in my branch. But I believe I might miss stuff. Also, I was thinking to pass a different parameter into the map instead of modifying the `Parent.Name` property.

ping @raphael

### DIFF
--- a/goagen/codegen/helpers.go
+++ b/goagen/codegen/helpers.go
@@ -167,35 +167,6 @@ func SnakeCase(name string) string {
 
 // KebabCase produces the kebab-case version of the given CamelCase string.
 func KebabCase(name string) string {
-	for u, l := range toLower {
-		name = strings.Replace(name, u, l, -1)
-		// Replace any potential underscore by a dash character following kebab-case alg.
-		name = strings.Replace(name, "_", "-", -1)
-	}
-	var b bytes.Buffer
-	var lastDash bool
-	ln := len(name)
-	if ln == 0 {
-		return ""
-	}
-	b.WriteRune(unicode.ToLower(rune(name[0])))
-	for i := 1; i < ln; i++ {
-		r := rune(name[i])
-		nextIsLower := false
-		if i < ln-1 {
-			n := rune(name[i+1])
-			nextIsLower = unicode.IsLower(n) && unicode.IsLetter(n)
-		}
-		if unicode.IsUpper(r) {
-			if !lastDash && nextIsLower {
-				b.WriteRune('-')
-				lastDash = true
-			}
-			b.WriteRune(unicode.ToLower(r))
-		} else {
-			b.WriteRune(r)
-			lastDash = false
-		}
-	}
-	return b.String()
+	name = SnakeCase(name)
+	return strings.Replace(name, "_", "-", -1)
 }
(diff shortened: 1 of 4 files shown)


goagen/codegen/helpers.go
```diff
@@ -164,3 +164,38 @@ func SnakeCase(name string) string {
 	}
 	return b.String()
 }
+
+// KebabCase produces the kebab-case version of the given CamelCase string.
+func KebabCase(name string) string {
```
raphael: Couldn't this simply call `SnakeCase` and then replace the `_` with `-`?

goagen/gen_client/cli_generator.go
```diff
@@ -162,6 +162,9 @@ func (g *Generator) generateCommands(commandsFile string, clientPkg string, func
 		}
 		return res.IterateActions(func(action *design.ActionDefinition) error {
 			name := codegen.Goify(action.Name, false)
+			// To avoid underscore in the resource name, we apply the kebabCase
+			// algorithm over the action.Parent.Name property.
+			action.Parent.Name = codegen.KebabCase(action.Parent.Name)
```
raphael: I think a better approach here would be to leave the name of the resource and action unchanged (they could be used for other things now and in the future) and instead call the `KebabCase` function from the template where the command and action names are defined (see `registerCmdsT`).

goagen/codegen/helpers.go
hectorj2f: Sure.

goagen/gen_client/cli_generator.go
hectorj2f: Alright!

hectorj2f:
@raphael 👍, I'll add a test for the KebabCase function

goagen/codegen/helpers_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/goadesign/goa/goagen/codegen"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestHelpers(t *testing.T) {
+	var (
+		stringTestA = "testAa"
```
kwk: Why not inline those to the `Expect` call below? It would simplify reading the tests a bit.

hectorj2f: I wanted to be specific but I have nothing against this suggestion :). I'll apply those changes

hectorj2f: Done @kwk

hectorj2f:
@raphael I addressed your comment, tests seem to be happy :).

raphael:
Looks great, thank you!

raphael:
Do you think you could backport these changes to the `v1` branch?

hectorj2f:
`Do you think you could backport these changes to the v1 branch?` I won't know how to do it. Could you ?

raphael:
OK :) FYI all that is required is to checkout the `v1` branch, cherry-pick the commit and create a new PR against the `v1` branch. I'll do it.

Issue #1040, opened 2017-02-08
gen_main: flag to specify package name and a flag to omit the `main.go` file


hectorj2f:
We'd like to generate all the controllers into a specified folder. This is possible at this moment. However, we would like to specify the package name, instead of `main`. Also, we want to omit the generation of the `main.go` file. We need the later cause we want to create our own main file that afterwards imports the controllers from the controller pkg.

raphael:
`gen_main` is meant to be used once when you start a new project. It's not meant to be used continuously. The files it generates are never overridden. It's as if you had written these files - just the tool helps you bootstrap. Adding more configuration to it would be going in the wrong direction in my opinion. If you look at the generated content it almost never changes and when it does it's just a matter of adding a couple of lines to implement the new method / interface or mount the controller.

PR #1041, opened 2017-02-08
[RFC-WIP] Added new flags to gen_main cmd

hectorj2f:
related to this: https://github.com/goadesign/goa/issues/1040

ping @raphael

### DIFF
--- a/goagen/gen_main/generator.go
+++ b/goagen/gen_main/generator.go
@@ -28,27 +28,31 @@ func NewGenerator(options ...Option) *Generator {
 
 // Generator is the application code generator.
 type Generator struct {
-	API       *design.APIDefinition // The API definition
-	OutDir    string                // Path to output directory
-	DesignPkg string                // Path to design package, only used to mark generated files.
-	Target    string                // Name of generated "app" package
-	Force     bool                  // Whether to override existing files
-	genfiles  []string              // Generated files
+	API             *design.APIDefinition // The API definition
+	OutDir          string                // Path to output directory
+	DesignPkg       string                // Path to design package, only used to mark generated files.
+	Target          string                // Name of generated "app" package
+	Force           bool                  // Whether to override existing files
+	Pkg             string                // Package name used instead of main
+	NoGeneratedMain bool                  // Whether to use the generated main.go or not
+	genfiles        []string              // Generated files
 }
 
 // Generate is the generator entry point called by the meta generator.
 func Generate() (files []string, err error) {
 	var (
-		outDir, designPkg, target, ver string
-		force                          bool
+		outDir, designPkg, target, ver, pkg string
+		force, noGeneratedMain              bool
 	)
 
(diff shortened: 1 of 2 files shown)


raphael:
Ah you already did the work :( Sorry but I believe these changes would make the tool more confusing to use. It would push for a model where `gen_main` is used every time the design changes when it is not the intent of the tool.

Can you describe exactly how you use it? maybe there's something else that can be done that works for you and stays inline with how the tool was designed. The one use case where I could see some benefit is if you add a new resource and you don't want to have to write the controller struct scaffold by hand. Any other change is trivial to apply. Maybe there could be a `controller`  command that accepts the name of a resource, a path, a package name and produces that scaffold - would that help your use case?

hectorj2f:
@raphael We want to generate the controllers created in `gen_main` and move them into a specific directory and a specific folder. We have our own `main.go` in our project which would like to explicitly call those controllers apart of doing other stuff.

`Maybe there could be a controller command that accepts the name of a resource, a path, a package name and produces that scaffold - would that help your use case?`

With this, you mean to create a new cmd like client, main, ... that does what we need but it's called `controller` ?

raphael:
Yes, something like:
```
goagen controller --res=foo --pkg=main --out=foo.go --design=github.com/foo/bar/design
```
that would generate the controller scaffold `foo.go` where:
* `--res` specifies the name of the resource and is compulsory
* `--pkg` specifies the name of the generated package (defaults to `main`)
* `--out` specifies the name of the relative path to the generated file (defaults to `<name of resource>.go`) 
* `--design` specifies the Go import path of the design package.

hectorj2f:
@raphael I think that would fit into our problem although we might want to also omit the generation of the controller scaffold.

raphael:
The controller scaffold would be the only thing generated in my proposal :) - it's also the only thing that's generated by `gen_main` apart from `main.go`. What is it that you want generated? Maybe just confusing terminology?

hectorj2f:
Okay, that might be the problem but the generator.go of `gen_main` generates two types of files: the mainFile, and the controllers ones. The main.go is the controller scaffold, right ?. I mean I need the other ones.

raphael:
OK we mean the same thing. I call the controller files the controller scaffolds because they are just scaffolding, they don't actually implement the controller logic (that's left to the user).
So it seems that proposal would work for you? One detail that's not clear to me is whether the `controller` generator should override existing files or not. It's scaffold so probably not -on the other hand you call it explicitly... Maybe we do the same as `main` and allow for a `--force` flag.

hectorj2f:
Yeah! I think that having a `--force` flag won't hurt anybody ;). So, I believe I should push a new PR with such a functionality.

raphael:
Sounds great, thank you!

hectorj2f:
I'll close this PR in favour of https://github.com/goadesign/goa/pull/1046. I linked to this one cause the conversation was done here.

PR #1046, opened 2017-02-09
Add new command `goagen controller`

hectorj2f:
Related to the discussion hold in: https://github.com/goadesign/goa/pull/1041#issuecomment-278413154

```
Available Commands:
  app         Generate application code
  bootstrap   Equivalent to running the "app", "main", "client" and "swagger" commands.
  client      Generate client package and tool
  commands    Lists all commands and flags in JSON
  controller  Generate controller scaffolding
  gen         Run third-party generator
  js          Generate JavaScript client
  main        Generate application scaffolding
  schema      Generate JSON Schema
  swagger     Generate Swagger
  version     Print the version number of goagen

Flags:
      --debug           enable debug mode, does not cleanup temporary files.
  -d, --design string   design package import path
  -o, --out string      output directory (default ".")

Use "goagen [command] --help" for more information about a command.
```

I set as default `controller` for the package name and the generated file `controller.go`. There are some conflicts with `pkg` from other commands (target pkg), so I created a new param called `pkg-name`. This flag specifies the name of the generated package. In the generated file, I coded a Factory function `NewControllerService` that returns the created controller scaffold (in gen_main is main.go).

```
Generate controller scaffolding

Usage:
  goagen controller [flags]

Flags:
      --force             overwrite existing files
      --pkg string        Name of generated Go package containing controllers supporting code (contexts, media types, user types etc.) (default "app")
      --pkg-name string   specify the name of the generated controller package (default "controller")
      --res string        specify the name of the resource and is compulsory (default "controller")

Global Flags:
      --debug           enable debug mode, does not cleanup temporary files.
  -d, --design string   design package import path
  -o, --out string      output directory (default ".")
```

ping @raphael


goagen/gen_controller/doc.go
```diff
@@ -0,0 +1,9 @@
+/*
+Package gencontroller provides a generator for a skeleton goa application.
+This generator generates the code for a basic "controller" package and is mainly intended as a way to
+create the controllers of new applications.
+The generator creates a controller.go file and one file per resource listed in the API metadata.
```
raphael: That description doesn't match right? Maybe something like:

>Package `gencontroller` generates the controller code for a given design resource. This generator is intended for use when resources are added to the design after the initial bootstrapping.

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,329 @@
...
+	var (
+		outDir, designPkg, target, ver, res, pkgName string
+		force                                        bool
+	)
+
+	set := flag.NewFlagSet("controller", flag.PanicOnError)
+	set.StringVar(&outDir, "out", "", "")
+	set.StringVar(&designPkg, "design", "", "")
+	set.StringVar(&target, "pkg", "app", "")
+	set.StringVar(&pkgName, "pkg-name", "controller", "")
```
raphael: Just reuse the existing `pkg` flag. No need to introduce a new one.

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,329 @@
...
+	set.StringVar(&res, "res", "controller", "")
```
raphael: I thought this flag (`res`) would identify which resource to generate the code for - it should not have a default value.

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// Generator is the application code generator.
+type Generator struct {
+	API       *design.APIDefinition // The API definition
+	OutDir    string                // Path to output directory
+	DesignPkg string                // Path to design package, only used to mark generated files.
+	Target    string                // Name of generated "app" package
+	Force     bool                  // Whether to override existing files
+	Pkg       string                // Name of the generated package
```
raphael: Just use `Target` - no need to introduce a new field (`Pkg`).

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,329 @@
...
+			return nil, err
+		}
+	}
+	imports := []*codegen.ImportSpec{
+		codegen.SimpleImport("io"),
+		codegen.SimpleImport("github.com/goadesign/goa"),
+		codegen.SimpleImport(imp),
+		codegen.SimpleImport("golang.org/x/net/websocket"),
+	}
+	err = g.API.IterateResources(func(r *design.ResourceDefinition) error {
```
raphael: I don't understand this code starting from here to the end of the file. The idea is to generate the the controller code for the one resource, why are we looping through all resources?

Also this should reuse the template and code from the `genmain` package so that when the `genmain` package changes how it generates code this generator output is also updated.

goagen/gen_controller/generator.go
hectorj2f: The result could be main or controller or whatever... that it would be the name of the generated file (analogous to the name of the MainFile in `gen_main`).

goagen/gen_controller/generator.go
hectorj2f: Writing something below to understand the problem here.

hectorj2f:
I will try to be more explicit to see if we can understand each other :).. 

We run the following commands (including the new version):
```
	$(GOAGEN_BIN) app -d ${PACKAGE_NAME}/${DESIGN_DIR}
	$(GOAGEN_BIN) controller -d ${PACKAGE_NAME}/${DESIGN_DIR} -o controller/ --pkg-name controller
	$(GOAGEN_BIN) client -d ${PACKAGE_NAME}/${DESIGN_DIR}
	$(GOAGEN_BIN) swagger -d ${PACKAGE_NAME}/${DESIGN_DIR}
```

I'll focus on what `controller` cmd does:

- In my design folder, I have the following files `a.go, b.go, c.go` that define resources and so forth.

- The output of the `app cmd` generated some functions and files that are invoked by the files generated by the controller command. The `Target` parameters is `app` which is the name of the package used by `controller/controller.go` (or bottle/bottle.go) to call the function `MountX` defined in the package `app` in the file `app/controllers.go`. That is why `Target` parameter isn't valid to be used as package name meaning by pkg name the name defined at the top of the go files. There is an example below:

```
file: app/controllers.go
func MountX(...) {...}

file: controller/controller.go
func NewControllerService() {
....
  app.MountX(service,...)
}
```

- controller command generates a part from the `a.go,b.go,c.go` files another named `controller.go`. This one originally could contain a `main()` function but if you want to get a main function why don't you use `gen main`. That is why we print instead of a main(), a function Factory that returns a service.

```
package <pkg-name>
...
	temp1 := controllers.NewXController(service)
	app.MountXController(service, temp1) // where app is the `target` parameter
```

Maybe we could omit the generation of `controllers.go` and just generate each one of the files `a.go`, `b.go` ... done by the controller cmd. Does it makes more sense to you ? This will omit the generation of the file `main.go` as done for `gen main`.

Another option could be to set a phone call to speed up this discussion and come to one agreement or not.

hectorj2f:
ping @raphael I might be confused, is your generated `bottle.go` what in my model would be the files `controller/a.go,controller/b.go` ?

raphael:
Thank you for the clarification, I understand the need for a new flag now to indicate the name of the controller package. Having `--pkg` and `--pkg-name` seems confusing though. It would seem better to use `--pkg` to denote the controller package name (instead of the `app` package) and have a different flag like `--app-pkg` which defaults to `app` which is the name of the package containing the generated `MountXXX` functions.

Another disconnect between what you are describing above and the way I was thinking about this is that in my mind the `--controller` would generate the code for a single resource. I can see how that would not work well with what you are trying to do though. A compromise here might be to make the `--res` flag optional and if provided filter the resources to only generate the code for the given resource. If not provided the controller code for all the resources would get generated. In any case the `controller` command should not generate a `main` or anything else but the controller scaffoldings.

hectorj2f:
@raphael I pushed some code to double check with you if it goes inline with your idea. :)

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,265 @@
...
+		codegen.SimpleImport("io"),
+		codegen.SimpleImport("github.com/goadesign/goa"),
+		codegen.SimpleImport(imp),
+		codegen.SimpleImport("golang.org/x/net/websocket"),
+	}
+
+	err = g.API.IterateResources(func(r *design.ResourceDefinition) error {
+		filename := filepath.Join(g.OutDir, codegen.SnakeCase(r.Name)+".go")
+
+		if g.Resource != "" {
```
raphael: This could just be
```
if g.Resource == "" || g.Resource == r.Name {
```

goagen/gen_controller/generator.go
```diff
@@ -0,0 +1,265 @@
...
+			g.generateControllerFile(filename, funcs, imports, r)
+		}
+
+		return nil
+	})
+
+	return g.genfiles, nil
+}
+
+func (g *Generator) generateControllerFile(filename string, funcs template.FuncMap, imports []*codegen.ImportSpec, r *design.ResourceDefinition) error {
```
raphael: Please refactor the `genmain` package slightly so that it exposes the function that generates the code (i.e. make a function out of https://github.com/goadesign/goa/blob/master/goagen/gen_main/generator.go#L112-L139) for a single controller. call it from here and remove all the copy/pasted code.

goagen/gen_controller/generator_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	JustBeforeEach(func() {
+		files, genErr = gencontroller.Generate()
+	})
+
+	AfterEach(func() {
+		os.RemoveAll(outDir)
+	})
+
+	Context("with a dummy API", func() {
```
raphael: This test does not apply to this generator.

goagen/gen_controller/generator.go
hectorj2f: Sure, i just code something quick for validation. I will address your comments

20zinnm:
+1, need the ability to generate specific files so as to not lose all database wiring by default.

goagen/gen_controller/generator.go
hectorj2f: done

goagen/gen_controller/generator_test.go
hectorj2f: done

hectorj2f:
ping @raphael I did some changes based on your comments :)

raphael:
The latest iteration looks great thank you! One last tweak :) instead of creating a new file in the util package please leave the functions in the `genmain` package and have the `gencontroller` package import it. It's ok for a generator to import another (for example `genswagger` imports `genschema`). Having  a `util` package is a little bit of an anti-pattern and something I'd like to get rid of eventually.

hectorj2f:
Done @raphael

raphael:
Thank you!

raphael:
This probably should be backported to `v1` as well.

20zinnm:
So should `goagen bootstrap` use the same format as this, with controllers in their own package?

raphael:
Hmm it's the opposite - the default controller package should be `main` (which can be overridden via the `--pkg` flag). Maybe the logic could be even smarter and infer the default package name from `--out`  if it's not the current folder.

raphael:
@hectorj2f fyi: https://github.com/goadesign/goa/pull/1048

raphael:
@hectorj2f are you backporting or do you want me to?

hectorj2f:
@raphael be my guest :)

raphael:
Alright, https://github.com/goadesign/goa/pull/1049

raphael:
@hectorj2f could I convince you to add a section about the new `controller` command on https://goa.design/implement/goagen/ ?
The source of the page is https://github.com/goadesign/goa.design/blob/master/content/implement/goagen.md with a copy at https://github.com/goadesign/goa.design/blob/master/content/implement/goagen.ja.md

hectorj2f:
Yeah i could add it, no pb @raphael

hectorj2f:
@raphael I'll take care of the documentation for the new cmd this week 👍

PR #1050, opened 2017-02-13
Fix goagen to not call panic when API definition is missing

tchssk:
Generators in goagen except gen_app call panic when API definition is missing.
This patch fixes it.


raphael:
This is great, thank you! Do you think you could backport this to the `v1` branch as well? thank you!

tchssk:
Done! #1052

PR #1053, opened 2017-02-14
Added new custom service listen and serve (for timeouts)

kchugalinskiy:
Without ability to set custom HTTP Server one is unable to set such important things as server timeouts. I've made this fix while trying to connect to my server from 1000 of clients, each dropping the connection without warning (network error simulation).

### DIFF
--- a/service.go
+++ b/service.go
@@ -151,10 +151,24 @@ func (service *Service) ListenAndServe(addr string) error {
 	return http.ListenAndServe(addr, service.Mux)
 }
 
+// ListenAndServeCustom starts a custom HTTP server.
+func (service *Service) ListenAndServeCustom(s *http.Server) error {
+	service.LogInfo("listen", "transport", "http", "addr", s.Addr)
+	s.Handler = service.Mux
+	return s.ListenAndServe()
+}
+
 // ListenAndServeTLS starts a HTTPS server and sets up a listener on the given host/port.
 func (service *Service) ListenAndServeTLS(addr, certFile, keyFile string) error {
 	service.LogInfo("listen", "transport", "https", "addr", addr)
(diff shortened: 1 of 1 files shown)


raphael:
Thank you ! Do you think this is really needed ? Each function is 3 lines which you can write yourself, they don't use private functions.

kchugalinskiy:
I'm not sure if this is needed by many people, but I've got some problems with multiple hanging connections, quickly running out of free ports. That was the reason. I'm sure that these lines are obvious but I've spent some time to figure out if there is some option to specify timeouts. Frankly speaking, I was surprised thad there is not.

raphael:
Yeah I don't doubt it took some troubleshooting to understand the problem and come up with a solution. It seems though that at the end of the day it's about understanding how the stdlib behaves and how to configure it. To say it differently: do you think having a function called `ListenAndServeCustom` exposed by goa would have helped you understand the problem faster?

kchugalinskiy:
I'm not pretty sure. But these functions (ListenAndServe) seemed a good place to start. Maybe we I should send a map with properties (which requires reflect dependency, not the best way to go) or just rename it some way. Like `ListenAndServeConfig` and pass some Config instead. This will incapsulate usage of http.Server, but will cause some trouble being in-sync with `http.Server` interface. Which way is better for your library? I'm not pretty sure. The only thing I'm sure that we NEED a way to configure underlying http.Server.

raphael:
I agree that being able to configure the underlying server is necessary. My point is that nothing needs to change or be added to goa for that to happen. In other words, this code:
```
	service.LogInfo("listen", "transport", "http", "addr", s.Addr)
 	s.Handler = service.Mux
	s.ListenAndServe()
```
can live anywhere -, no need to increase the complexity of the goa interface.

kchugalinskiy:
Ok, thanks for your attention

PR #1054, opened 2017-02-17
Fix gen_client to not crash when an array parameter is defined in an action using websocket

tchssk:
https://github.com/goadesign/goa/commit/0e7563b1eace965bb172548f8bf3174aa45e8370 adds a test context to reproduce the issue.
Then, https://github.com/goadesign/goa/commit/14b0f8d977ee9bd590e90b9c93c5302b61befdac is a patch to fix it.

PR #1055, opened 2017-02-17
Fix gen_client to not crash when an array parameter is defined in an action using websocket

tchssk:
This is a backporting of #1054 to v1.


raphael:
Thank you!

PR #1058, opened 2017-02-21
v2:  Fix goagen to buildable

tchssk:
This patch fixes some renamed packages and functions, and change goagen's package to `main`.


raphael:
Great, thanks. I'm currently working on making the account example client cli. Once I'm done my next item is to make the build green on v2.

PR #1073, opened 2017-02-24
Add nil check for user type in gen_app

### LINKED ISSUE #1060: Panic in Publicize when using Hash(Integer, PayloadType)
When using a Hash(Integer, PayloadType) goa generates a map[int]*PayloadType, which is correct.
However, in the Publicize function of the generated code in user_types.go, the following is generated:

```
	if ut.Triggers != nil {
		pub.Triggers = make(map[int]*GameengineAchievementTriggerPayload, len(ut.Triggers))
		for k2, v2 := range ut.Triggers {
			pubk2 := k2
			pubv2 := v2.Publicize()
			pub.Triggers[pubk2] = pubv2
		}
	}
```

This causes a panic if the client sends a null value for v2

tchssk:
Fix #1060.

### DIFF
--- a/goagen/codegen/publicizer.go
+++ b/goagen/codegen/publicizer.go
@@ -131,8 +131,14 @@ const (
 */}}{{ $k := printf "%s%d" "k" .depth }}{{ $v := printf "%s%d" "v" .depth }}
 {{ tabs .depth }}for {{ $k }}, {{ $v }} := range {{ .sourceField }} {
 {{ $pubk := printf "%s%s" "pub" $k }}{{ $pubv := printf "%s%s" "pub" $v }}{{/*
-*/}}{{ tabs .depth }}{{ publicizer .keyType $k $pubk .dereference (add .depth 1) true }}
-{{ tabs .depth }}{{ publicizer .elemType $v $pubv .dereference (add .depth 1) true }}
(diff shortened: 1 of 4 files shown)


tchssk:
Would you try this to your project? @luna-duclos

luna-duclos:
Just tested this, code looks good, and code works! 👍

raphael:
Thank you for the great PR with additional tests! I am wondering though if it wouldn't be better to put the test in the caller (https://github.com/goadesign/goa/blob/master/goagen/codegen/publicizer.go#L130) rather than in the `Publicize` function itself. `Publicize` should not be called on `nil` and protecting against that could hide other bugs (where the target should not have been `nil` but was because of a bug).

luna-duclos:
I think a function not blowing up on nil is generally good behaviour. I can see your point about masking other bugs however, but considering these bugs would usually be the same nil check .. I don't see the harm

raphael:
For example you may end up with code that is generated fine but does not compile because the `Publicize` method is missing. This is a lot harder to track than a clear error such as the one you hit where it took me 10 seconds to find the offending line. In general I dislike the idea of writing defensive code "just because" - it makes tracking issues down the line orders of magnitude more difficult.

tchssk:
I reverted the commit and moved nil check to publicizer for hash. Is it correct?

raphael:
Yes perfect, thank you for making the change!

PR #1080, opened 2017-02-27
ErrorResponse matches ErrorMedia.

### LINKED ISSUE #1038: Inconsistency between ErrorMedia and ErrorResponse
Goa's internal media type `ErrorMedia` which represents error has attribute `meta` defined as `Hash` (basically `Object`), but goa's type `ErrorResponse` which is used to generate error response has attribute `meta` defined as slice of maps (basically `Array of Objects`.

Attribute `meta` of type `ErrorResponse` should be changed to `map` to match `ErrorMedia`.

sotosof:
Fix #1038.

### DIFF
--- a/error.go
+++ b/error.go
@@ -107,7 +107,7 @@ type (
 		// Detail describes the specific error occurrence.
 		Detail string `json:"detail" xml:"detail" form:"detail"`
 		// Meta contains additional key/value pairs useful to clients.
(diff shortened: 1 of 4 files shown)


raphael:
Thank you for the great PR! It's a bit of a breaking change but since it also fixes a bug would you mind backporting that to the `v1` branch? (i.e. checkout the v1 branch, cherry-pick the commit and make a new PR using a different branch name e.g. `v1.error-response-fix`). Thanks again!

PR #1084, opened 2017-02-28
Avoid using keywords for arguments in TestHelpers

### LINKED ISSUE #1083: generated test fails with query parameter
I found generated code fails with `query` parameter.
maybe generated test code using `query` variable.

Example design is here.

```
package design

import (
	. "github.com/goadesign/goa/design"
	. "github.com/goadesign/goa/design/apidsl"
)

var _ = API("sample", func() {
	Title("sample")
	Description("sample")
	Scheme("http")
	Host("localhost:8080")
})

var _ = Resource("tests", func() {
	BasePath("/tests")
	DefaultMedia(TestMedia)

	Action("list", func() {
		Description("")
		Routing(
			GET("/"),
		)
		Params(func() {
			Param("query", String)
			Required("query") // <---- here
		})
		Response(OK)
		Response(NotFound)
	})
})

var TestMedia = MediaType("application/vnd.sample.v1+json", func() {
	Description("video")

	Attributes(func() {
		Attribute("id")
	})

	View("default", func() {
		Attribute("id")
	})
})
```

```
goagen bootstrap -d <GOPATH>/goatest/design
app
app/contexts.go
app/controllers.go
app/hrefs.go
app/media_types.go
app/user_types.go
app/test
app/test/tests_testing.go
main.go
tests.go
tool/sample-cli
tool/sample-cli/main.go
tool/cli
tool/cli/commands.go
client
client/client.go
client/tests.go
client/user_types.go
client/media_types.go
swagger
swagger/swagger.json
swagger/swagger.yaml
```

and run test 
```
go test ./...
# github.com/azihsoyn/goatest2/app/test
app/test/tests_testing.go:51: no new variables on left side of :=
app/test/tests_testing.go:51: cannot use url.Values literal (type url.Values) as type string in assignment
app/test/tests_testing.go:54: non-integer string index "query"
app/test/tests_testing.go:54: cannot assign to query["query"]
app/test/tests_testing.go:54: cannot use sliceVal (type []string) as type byte in assignment
app/test/tests_testing.go:58: query.Encode undefined (type string has no field or method Encode)
app/test/tests_testing.go:117: no new variables on left side of :=
app/test/tests_testing.go:117: cannot use url.Values literal (type url.Values) as type string in assignment
app/test/tests_testing.go:120: non-integer string index "query"
app/test/tests_testing.go:120: cannot assign to query["query"]
app/test/tests_testing.go:120: too many errors
```

Thanks.

tchssk:
Fix #1083.

### DIFF
--- a/goagen/gen_app/test_generator.go
+++ b/goagen/gen_app/test_generator.go
@@ -58,7 +58,11 @@ func (g *Generator) generateResourceTest() error {
 	if len(g.API.Resources) == 0 {
 		return nil
 	}
-	funcs := template.FuncMap{"isSlice": isSlice}
+	funcs := template.FuncMap{
+		"isSlice":       isSlice,
+		"initArguments": initArguments,
+		"tempvar":       tempvar,
+	}
 	testTmpl := template.Must(template.New("test").Funcs(funcs).Parse(testTmpl))
 	outDir, err := makeTestDir(g, g.API.Name)
 	if err != nil {
@@ -218,6 +222,36 @@ func (g *Generator) createTestMethod(resource *design.ResourceDefinition, action
 	}
 }
 
+var arguments map[string]bool
+
+func initArguments(method TestMethod) string {
+	arguments = make(map[string]bool)
+	for _, param := range method.Params {
+		arguments[param.Name] = true
+	}
+	for _, param := range method.QueryParams {
+		arguments[param.Name] = true
+	}
+	for _, header := range method.Headers {
+		arguments[header.Name] = true
+	}
+	if method.Payload != nil {
+		arguments[method.Payload.Name] = true
+	}
+	if method.ReturnType != nil {
+		arguments[method.ReturnType.Name] = true
+	}
+	return ""
+}
+
+func tempvar(s string) string {
+	if ok := arguments[s]; ok {
+		s = tempvar(s + "_")
+	}
+	arguments[s] = true
+	return s
+}
+
 // pathParams returns the path params for the given action and route.
 func pathParams(action *design.ActionDefinition, route *design.RouteDefinition) []*ObjectType {
 	return paramFromNames(action, route.Params())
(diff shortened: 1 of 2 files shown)


tchssk:
It might be better not to use local variables in the test helper functions.

goagen/gen_app/test_generator.go
```diff
@@ -58,7 +58,11 @@ func (g *Generator) generateResourceTest() error {
 	if len(g.API.Resources) == 0 {
 		return nil
 	}
-	funcs := template.FuncMap{"isSlice": isSlice}
+	funcs := template.FuncMap{
+		"isSlice":       isSlice,
+		"initArguments": initArguments,
```
raphael: Does `initArguments` need to be called when the template is rendered? It could be instead called by the code prior to rendering the template which would be better.

goagen/gen_app/test_generator.go
```diff
@@ -58,7 +58,11 @@ func (g *Generator) generateResourceTest() error {
...
+		"tempvar":       tempvar,
```
raphael: Maybe `escape` is a better name for this function than `tempvar`?

tchssk:
Thank you for your great revew.

I added a new commit fixing the matters.

tchssk:
Now local variables use `_` as suffix. Should we use prefix instead of suffix?

goagen/gen_app/test_generator.go
```diff
@@ -59,9 +60,8 @@ func (g *Generator) generateResourceTest() error {
 		return nil
 	}
 	funcs := template.FuncMap{
-		"isSlice":       isSlice,
-		"initArguments": initArguments,
-		"tempvar":       tempvar,
+		"isSlice": isSlice,
+		"escape":  escape,
```
raphael: `escape` should probably be a method on `Generator` this way you won't have to pass `$test` everywhere in the template (so `funcs` would be initialized with `"escape": g.escape`).

raphael:
Using a suffix is fine too.

goagen/gen_app/test_generator.go
tchssk: You mean `Generator` has `escape` as a method and `reservedNames` as a field, right?

In that way, Generator will have all reserved names in the service and use them for all test helpers. `reservedNames` also affects the functions which actually does not have reserved names. Is it OK?

tchssk: The way that `TestMethod` has `escape` as a method might be better. In this way, we won't have to pass `$test` everywhere in the template.

raphael: Yes sorry you're correct. `reservedNames` and `escape` should be scoped to `TestMethod` - not `Generator`.

tchssk: Thank you! I added the changes as two commits.

raphael:
Thank you!

Issue #1075, opened 2017-02-24
Add a `Bytes` Primitive Type to Goa `v1`


JKhawaja:
A Goa primitive type that maps to a `[]byte` in Go and possibly JSON objects in other places (e.g. swagger) could be a useful. It is upcoming in Goa `v2`, however a `v1` update with it could be useful (and simpler/faster to implement). Any thoughts or questions?

raphael:
Thanks for opening the issue! Just to be clear, adding a new primitive type requires updating all the code generators (`genapp`, `gen_client`, `gen_controller`, `gen_js`, `gen_main`, `gen_schema` and `gen_swagger`). In most cases I expect the required changes to be small though.

JKhawaja:
Will there be any other reason for a `Bytes` primitive to exist other than it mapping to a `json.RawMessage` type in the `gen_app` code? Which would imply that it is only useful when using either JSON encoding for producing or consuming.

raphael:
I think a `Byte` type would have many more applications. The question though is not what it would be used for, it's what needs to be done so that goagen isn't broken if someone uses `Byte` in the design. Also it wouldn't map to `json.RawMessage` but to `[]byte` in the generated code, that code has to be independent of encoding. (the controller can then simply type assert to `json.RawMessage`).

JKhawaja:
Will the goa JSON encoder unmarshal the request Payload to a Type definition if a field has a Type `[]byte` or a `*[]byte`? I don't think it will. That is why I am wondering if there are any specific use cases for typing an `Attribute` as `Byte`. Outside of request and response body types it does not seem like too many people would care about arrays of uint8s?

raphael:
The broader use case for bytes is just plain binary data rather than a valid set of UTF-8 characters. You're right that the default JSON behavior today will be to do the base 64 encode/decode which is not great. We can't use `json.RawMessage` as the Go type for `Byte` though, again it needs to be a type that works with any marshaler / Swagger / JavaScript etc. Primitive types that are defined in the DSL cannot be bound to a specific code generator / output. 

Maybe a solution here would be to add a special Metadata tag that allows overriding the type generated for the `app` package. This would be useful for other things as well (e.g. `int64` instead of `int`).

JKhawaja:
So, would it be a `struct:field:type` tag that gets called like `Metadata("struct:field:type", "json.RawMessage")`?

raphael:
[...]

PR #1085, opened 2017-02-28
code, comments, and tests for the new struct:field:type Metadata tag

JKhawaja:
- `design/apidsl/metadata.go` (comments)
- `goagen/codegen/types.go` (code)
- `goagen/codegen/types_test.go` (test)

Reference Issue: #1075


goagen/codegen/types.go
```diff
@@ -115,6 +115,11 @@ func goTypeDefObject(obj design.Object, def *design.AttributeDefinition, tabs in
 			typedef = "*" + typedef
 		}
 		fname := GoifyAtt(field, name, true)
+		if tname, ok := field.Metadata["struct:field:type"]; ok {
```
raphael: Might make sense to move the test so that the type name doesn't have to be computed if the metadata is present. So move the test to line 113, declare the variable `typedef` above and have a `if typedef == ""` before the computation done on lines 113 to 116. Wow lots of words for something that is so simple :)

JKhawaja:
Code is updated. Passed tests as well. :)

design/apidsl/metadata.go
```diff
@@ -18,6 +18,11 @@ import (
 //
 //        Metadata("struct:field:name", "MyName")
 //
+// `struct:field:type`: overrides the Go struct field type generated by default by goagen.
+// Applicable to attributes only.
+//
+//        Metadata("struct:field:type", "MyType")
```
raphael: Please update the comment to reflect the ability to optionally specify a Go import path.

goagen/gen_app/generator.go
```diff
@@ -415,3 +417,81 @@ func (g *Generator) generateUserTypes() error {
 	}
 	return utWr.FormatCode()
 }
+
+// mediaTypeMetadataImports will grab all imports specified in a struct:field:type Metadata tag and
+// add them to the imports for the media_types.go file.
+func (g *Generator) mediaTypeMetadataImports(s []*codegen.ImportSpec) []*codegen.ImportSpec {
```
raphael: We need only one function that accepts a `*design.AttributeDefinition`. This function should be recursive to handle recursive definitions properly, something like:
```go
func (g *Generator) attributeImports(att *design.AttributeDefinition, imports []*codegen.ImportSpec, seen []*design.AttributeDefinition) []*codegen.ImportSpec {
    for _, a := range seen {
        if att == a {
           return
        }
    }
    seen = append(seen, att)
    if tname, ok := field.Metadata["struct:field:type"]; ok {
        // TBD: append to imports (use appendImports)
    }
    switch t := att.Type.(type) {
        case *design.UserTypeDefinition:
            return appendImports(imports, g.attributeImports(t.AttributeDefinition, seen))
        case *design.MediaTypeDefinition:
            return appendImports(imports, g.attributeImports(t.AttributeDefinition, seen))
        case design.Object:
            t.IterateAttributes(func(at *design.AttributeDefinition) error {
               imports = appendImports(imports, g.attributeImports(t.AttributeDefinition, seen))
               return nil
            })
            return imports
        case *design.Array:
            return appendImports(imports, g.attributeImports(t.ElemType, seen))
        case *design.Hash:
            imports = appendImports(imports, g.attributeImports(t.KeyType, seen))
            return appendImports(imports, g.attributeImports(t.ElemType, seen))
     }
}
```
where `appendImports` accepts a slice of import specs and a slice of names. It returns a new slice of import specs with non-duplicate imports added. Both `MediaTypeDefinition` and `UserTypeDefinition` have a `AttributeDefinition` field that can be given to this function.

Note that I have not tested or even compiled the code snippet above, this is just on top of my head / pseudo code :)

goagen/gen_client/generator.go
```diff
@@ -615,6 +617,84 @@ func (g *Generator) generateUserTypes(pkgDir string) error {
 	return utWr.FormatCode()
 }
 
+// mediaTypeMetadataImports will grab all imports specified in a struct:field:type Metadata tag and
+// add them to the imports for the media_types.go file.
+func (g *Generator) mediaTypeMetadataImports(s []*codegen.ImportSpec) []*codegen.ImportSpec {
```
raphael: Please move the `attributeImports` function described above to the `codegen` package and use it from both the `genapp` and `genclient` package instead of copy/pasting.

JKhawaja: So, we don't want it to be a Generator method?

goagen/gen_app/generator.go
JKhawaja: I'm not sure what you are getting at here. Besides the fact that we should probably have recursion.

raphael: Mainly that we only need one function, not a function for media type and another one for user type. One function that accepts an attribute works for both cases.

goagen/gen_client/generator.go
raphael: Correct, I don't think it has to be?

goagen/gen_app/generator.go
JKhawaja: So, we won't need to range over `g.API.MediaTypes` or `.Types`?

JKhawaja: What is the `seen` check on a `[]*design.AttributeDefinition` slice for?

raphael: It's to avoid infinite recursion, there can be cycles where an attribute uses another (e.g. an object) which ends up using the first one (e.g. recursive type definition).

JKhawaja:
Something must be bugged with the Travis CI build system right now.

goagen/gen_app/generator.go
```diff
@@ -372,6 +372,10 @@ func (g *Generator) generateMediaTypes() error {
 		codegen.SimpleImport("unicode/utf8"),
 		codegen.NewImport("uuid", "github.com/satori/go.uuid"),
 	}
+	for _, v := range g.API.MediaTypes {
+		var seen []*design.AttributeDefinition
+		imports = codegen.AttributeImports(v.AttributeDefinition, imports, seen)
```
raphael: Why not just pass `nil` here?

goagen/gen_app/generator.go
```diff
@@ -405,6 +409,10 @@ func (g *Generator) generateUserTypes() error {
 		codegen.SimpleImport("github.com/goadesign/goa"),
 		codegen.NewImport("uuid", "github.com/satori/go.uuid"),
 	}
+	for _, v := range g.API.Types {
+		var seen []*design.AttributeDefinition
+		imports = codegen.AttributeImports(v.AttributeDefinition, imports, seen)
```
raphael: same as above (and in `genclient` package as well)

design/apidsl/metadata.go
```diff
@@ -18,6 +18,11 @@ import (
 //
 //        Metadata("struct:field:name", "MyName")
 //
+// `struct:field:type`: overrides the Go struct field type generated by default by goagen.
+// Applicable to attributes only.
```
raphael: Maybe add `The second optional tag value specifies the Go import path to the package defining the type if not built-in`.

design/apidsl/metadata.go
```diff
@@ -18,6 +18,11 @@ import (
...
+//
+//        Metadata("struct:field:type", "mypackage.MyType", "github.com/me/mypackage")
```
raphael: Just to be clear it might be good to have a first example that does not use a package. Also would be good to use a real use case for the example like the one you are after so maybe something like:
```go
//        Metadata("struct:field:type", "[]byte")
//        Metadata("struct:field:type", "json.RawMessage", "encoding/json")
```

JKhawaja: Or, we could even put in all three? :P

goagen/gen_app/generator.go
JKhawaja: would have to restructure the entire AttributeImports function if we pass a nil seen here.

raphael:
Looks great, could I coerce you into adding a test for the 2 tag value? last request :-)

JKhawaja:
Sure, but where would that test live?

goagen/codegen/import_spec_test.go
```diff
@@ -0,0 +1,142 @@
...
+import (
+	. "github.com/goadesign/goa/design"
+	. "github.com/goadesign/goa/design/apidsl"
+	"github.com/goadesign/goa/dslengine"
+	"github.com/goadesign/goa/goagen/codegen"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("AttributeImports", func() {
```
raphael: It would also be interesting to test the recursion in particular with cycles, something like:
```go
o := Object{}
child := &AttributeDefinition{Type: o}
po := Object{"child":child}
parent := &AttributeDefinition{Type:po}
o["parent"] = parent
```

raphael:
Thank you, looks like this is ready to be merged?

raphael:
Addresses #1075

JKhawaja:
I think we should also recommend that if you use this Metadata tag to overwrite a field type in the generated app code, then in the design code you should still specify the `Attribute` type as `String` so you are able to use a string example of whatever Type you are using: `Example("{}")`

I ran into the problem of overwriting an `Integer` with a custom array type, and Goa would error out my provided example (a string) as being of an incorrect type.

JKhawaja:
And yes, this can be merged. :D

raphael:
Well the issue with recommending string that you may want non-Go generators (Swagger, JavaScript, JSON Schema) to use integer in your example. Thank you for the great PR!

raphael:
Could you backport this to `v1`? (branch off of `v1`, cherry pick the commit from master and create a PR against `v1`), thank you!

JKhawaja:
Woops, we have a problem. Not sure how I did not catch this. *facepalm* .. but the validation method `Validate()` in the in the generated `app` code (`user_types.go`) is still checking for empty strings if the `Attribute` type is set as `String`. This will cause the Go compiler to yell at you if you are using a non-string type in the Metadata tag.

JKhawaja:
**Nil Issue:**
Also, a problem with `nil` in `Validate()`: The problem seems to be occurring in the non-public Types (in `user_types.go`) where all the fields are pointers. If the type specified in the Metadata tag is not a pointer then it cannot be evaluated as `nil`. (which, of course, produces an invalid indirect error when trying to perform the `Publicize` method).

SOLUTION: we could either tell the user to be sure to use a pointer type e.g. `*myType`, OR we could possibly update the code:

```
if tname, ok := field.Metadata["struct:field:type"]; ok {
			if len(tname) > 0 {
				typedef = "*" + tname[0]
			}
		}
```

of course, we might also want to check that the specified Type in the Metadata tag is not already a pointer.

raphael:
For that last issue the code generator needs to check whether to add the `*` or not, it shouldn't (cannot) come from the design. I think the method takes a `public` bool or something like that which you can use to know whether to prefix with `*` or not. Look at how the code does it in the rest of the method.

raphael:
For the first issue ("string" issue) probably the simplest way to do it would be to not generate validations on attributes that have the tag defined on it. There should be a comment in the `Metadata` function explaining that this is the case. I wouldn't worry about checking type compatibility, this is the responsibility of the user.

JKhawaja:
Should I open a new PR since the other one has been merged? Or, do you want to revert the merge?

raphael:
New PR makes sense yes  - good you already fixed one!

JKhawaja:
Sorry about taking so long to get back to this problem. I think I added the necessary code in.

raphael:
FYI I ended up adding the code that prevents the validation generation in https://github.com/goadesign/goa/pull/1169. This PR also tweaks the code a little bit so that it's possible to use the metadata on arrays as well like so:
```go
var  T = Type("T", func() {
    Attribute("data", ArrayOf(String, func() {
        Metadata("struct:field:type", "json.RawMessage", "encoding/json")
    })
})
```
The PR also fixes an issue where the custom package wasn't imported in the generated `contexts.go` file even if a payload type used it.

raphael:
PR #1170 backports the "struct:field:type" metadata support to v1.

Issue #1077, opened 2017-02-27
Codegen: Endpoint writer


tchssk:
Implemented in package https://github.com/goadesign/goa/tree/v2/codegen/writers.

tchssk:
https://github.com/goadesign/goa/pull/1090

PR #1090, opened 2017-03-01
[WIP] Implement EndpointsWriter()

tchssk:


codegen/writers/endpoint.go
```diff
@@ -29,19 +32,80 @@ type (
...
+
+func (e endpointWriter) OutputPath() string {
+	return e.outputPath
+}
+
 // EndpointsWriter returns the codegen.FileWriter for the endpoints of the given
 // service.
 func EndpointsWriter(api *design.APIExpr, service *design.ServiceExpr) codegen.FileWriter {
-	return nil
+	return codegen.FileWriter(endpointWriter{
```
raphael: I don't think you need the cast?

codegen/writers/endpoint.go
```diff
@@ -29,19 +32,80 @@ type (
...
+	methods := make([]*endpointMethod, len(service.Endpoints))
+	for i, v := range service.Endpoints {
+		methods[i] = &endpointMethod{
+			Name:        v.Name,
+			Description: v.Description,
+		}
+		if payload, ok := v.Payload.(*design.UserTypeExpr); ok {
+			methods[i].Payload = payload
+		}
+		if result, ok := v.Result.(*design.UserTypeExpr); ok {
```
raphael: A result type could also be a media type. At this level we don't need to differentiate so you can use the `design.UserType` interface I believe.

codegen/writers/endpoint_test.go
```diff
@@ -0,0 +1,110 @@
+package writers
+
+import (
+	"bytes"
+	"testing"
+
+	"goa.design/goa.v2/design"
+)
+
+func TestEndpointTmpl(t *testing.T) {
```
raphael: Would be good to be a bit more systematic, I'll provide an example.

codegen/writers/endpoint.go
tchssk: It was not needed. I fixed it. https://github.com/goadesign/goa/pull/1090/commits/675ab2cd0e6315d4ff8f935050c17d5647f31607

codegen/writers/endpoint.go
tchssk: OK. Let's fix when we know it's wrong.

codegen/files_test.go
```diff
@@ -12,19 +12,17 @@ func TestHeaderTmpl(t *testing.T) {
 	}{
 		"a minimum header": {
 			data: map[string]interface{}{
-				"Pkg": "endpoints",
+				"Pkg": "foo",
```
raphael: Nitpick: instead of `foo` etc. let's use `a`, `b`, ...

tchssk: Is using `foobar` in tests not good?

raphael: It's fine in general, the problem is that when you need more than 3 different values you're stuck (`foo`, `bar`, `baz` ??). Just trying to be consistent, not a big deal.

tchssk: OK. but we can use up to 13th🙂 https://www.ietf.org/rfc/rfc3092.txt

raphael: Haha I didn't know there was a standard for that :)

tchssk: I also learned recently. I'll choose variable names on a case by case.

tchssk:
I'll rebase this branch onto v2 and squash the commits.

4a6e38c Move header template to section file.
2778b13 [WIP] Delete files_test.go
117c0eb [WIP] Add a function to create Endpoint section
de71f1d [WIP] Fix the comment of Header()
6dd3d6f [WIP] Fix the test for headerTmpl
675ab2c [WIP] Remove cast
c5f4100 [WIP] Use Header section in EndpointsWriter
24f0d84 [WIP] Add Header section
f5069d3 [WIP] Add a test for headerTmpl
f40fc3e [WIP] Remove a temporary variable
7b29bf6 [WIP] Change the test name
c3cb49c [WIP] Fix the template for endpoint
7aeed65 [WIP] Add a test for endpoint
61e0a7f [WIP] Fix the template for endpoint
b1247d7 [WIP] Fix buildEndpointData
049d21a [WIP] Implement buildEndpointData
4186add [WIP] Set output path
b6b1ebe [WIP] Update EndpointsWriter

codegen/writers/endpoint.go
```diff
@@ -29,19 +32,85 @@ type (
...
+	return endpointWriter{
+		sections: []*codegen.Section{
+			codegen.Header("", "endpoints", []*codegen.ImportSpec{
+				&codegen.ImportSpec{Path: "context"},
+				&codegen.ImportSpec{Path: "goa.design/goa.v2"},
+				&codegen.ImportSpec{Path: "goa.design/goa.v2/examples/account/gen/services"},
+			}),
+			Endpoint(api, service),
+		},
+		outputPath: "gen/endpoints/", // TODO Set output path.
```
tchssk: Hmm. How should we set `ImportSpec` of `services` and `outputPath`?

raphael: The output path is the relative path from the output path specified by user. So compute a filename from the service name (using Go conventions: lowercase and underscores to separate words) and append what you have above with it.

The services Go import path should be computed and I'm realizing now that we'll need the absolute output path for that (and use `build.Import`). So we should add the output path specified on the command line as an argument to `EndpointsWriter`.

The output path returned by the writer should still be relative though (not all writers may need the absolute path as input).

tchssk: I fixed `outputPath`. https://github.com/goadesign/goa/pull/1090/commits/f0c7899b5e7fb796113a54807e856168796efc81

raphael:
Whoops, probably shouldn't have merged yet. Not sure reverting is a good idea though. Please open a new PR.

tchssk:
OK. I'll open a new PR.

tchssk:
Related to #1077.

PR #1092, opened 2017-03-01
Add a test for headerTmpl

tchssk:


raphael:
Merging, I'll make a subsequent PR with the example.

PR #1093, opened 2017-03-01
Add a function to create header section

tchssk:


tchssk:
If we keep the templates private, it's yes.

We might have to bundle a template and a function to create a section using that. Each writer need not to use the templates directly, just use these functions.

raphael:
Yes that's what I'm thinking too: the writers just call the shared section functions, the template is just a (private) constant in the same file as the functions. I'll merge this for now so I can also merge the test and modify it to provide an example.

PR #1097, opened 2017-03-02
Rework Header section test

raphael:
Make test more systematic by testing each parameter independently.


raphael:
@tchssk I tried to illustrate more systematic tests with this PR, it's very simple tests though. The idea is to test each argument / interesting values (e.g. single vs. multiple imports) independently. The test uses artificial / short values for the parameters to try and be consistent.

codegen/sections_test.go
```diff
@@ -0,0 +1,68 @@
...
+		s := Header(tc.Title, "testpackage", tc.Imports)
+		s.Render(buf)
+		actual := buf.String()
+		if actual != tc.Expected {
+			t.Errorf("%s: got %#v, expected %#v", k, actual, tc.Expected)
+		}
+	}
+}
+
+const (
```
tchssk: These `const` should be in scope of `TestHeader()`, right? We have to take care to the duplication of that when we add some tests in this package.

tchssk:
I understood. I left a comment in review.

codegen/sections_test.go
raphael: OK I'll move them in the test.

codegen/sections_test.go
```diff
@@ -0,0 +1,68 @@
...
+		title = "test title"
+	)
+	var (
+		imprt   = []*ImportSpec{&ImportSpec{Path: "test"}}
+		imports = append(imprt, &ImportSpec{Path: "other"})
+	)
+	cases := map[string]struct {
+		Title    string
+		Imports  []*ImportSpec
+		Expected string
```
tchssk: These fields seem to don't need to be public.

raphael: They don't but I like to do it this way to make the field names stand out when defining the test cases.

tchssk: I'll write test in the same way.

tchssk:
Now this looks good to me. Is this OK to merge?

raphael:
Yes please!

PR #1099, opened 2017-03-04
Keys allows []*ecdsa.PublicKey

nicholaswyoung:
If I understand the other documentation correctly, `Keys` can be `[]*ecdsa.PublicKey` too.


raphael:
Indeed, thank you!

PR #1102, opened 2017-03-05
generator for rest paths + tests

michaelboke:
create codegenerator for the rest paths,

please review, i do think there could be some improvements here.
For example the pathWriter is duplication of the endpointWriter generator (maybe a general one)


rest/codegen/paths.go
```diff
@@ -1,11 +1,138 @@
...
+{{- if .Params}}
+	return fmt.Sprintf("{{ .Path }}", {{join .Params ", "}})
+{{- else}}
+	return "{{ .Path }}"
+{{- end}}
+}
+{{end}}
+`
+
+var pathTmpl = template.Must(template.New("path").
```
raphael: As a convention I've been putting `var` declarations at the top of the file after `type` declarations (there can be exceptions when it's a short declaration used by the one function below - e.g. regexp declarations)

rest/codegen/paths_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		service = goadesign.ServiceExpr{
+			Name: "Account",
+		}
+
+		endpoint = goadesign.EndpointExpr{
+			Name:    "Show",
+			Service: &service,
+			//Payload: &goadesign.UserTypeExpr{AttributeExpr: &params },
```
raphael: Please remove the inline comments or add an explanation of why they are there.

rest/codegen/paths.go
```diff
@@ -1,11 +1,138 @@
...
+	}
+	return args
+}
+
+func generateRoutePathName(routeNum int, a *design.ActionExpr) string {
+	serviceName := strings.Title(a.Service.Name)
+	endpointName := strings.Title(a.Name)
+
+	if routeNum == 1 {
+		return fmt.Sprintf("%s%sAlternativePath", endpointName, serviceName)
```
raphael: Just curious: what was the thinking behind using the `AternativePath` suffix rather than say `Path2`?

rest/codegen/paths.go
```diff
@@ -1,11 +1,138 @@
...
+	return &pd
+}
+
+func generatePathArguments(r *design.RouteExpr) []string {
+	params := r.Params()
+	obj := goadesign.AsObject(r.Action.PathParams().Type)
+	args := make([]string, len(params))
+
+	for i, name := range params {
+		args[i] = fmt.Sprintf("%s %s", name, obj[name].Type.Name())
```
raphael: To support slices we probably need to call a `GoTypeRef` style function.

rest/codegen/paths.go
```diff
@@ -1,11 +1,138 @@
...
+	pd := pathData{
+		ServiceName:  a.Service.Name,
+		EndpointName: a.Name,
+		Routes:       make([]*pathRoute, len(a.Routes)),
+	}
+
+	for i, r := range a.Routes {
+		pd.Routes[i] = &pathRoute{
+			Name:      generateRoutePathName(i, a),
+			Path:      design.WildcardRegex.ReplaceAllString(r.FullPath(), "/%v"),
```
raphael: I know this is what `v1` is doing but I've been wondering about allowing the use of slices in paths (that has come up a few times). In that case this wouldn't work.

rest/codegen/paths.go
```diff
@@ -1,11 +1,138 @@
...
+		// Arguments describe all the function arguments with types
+		Arguments []string
+	}
+)
+
+// PathWriter returns the path generators writer.
 func PathWriter(r *design.RootExpr) codegen.FileWriter {
-	return nil
+	sections := []*codegen.Section{
+		codegen.Header("", "http", []*codegen.ImportSpec{
```
raphael: Would be nice to set a non empty title here.

rest/codegen/paths_test.go
```diff
@@ -0,0 +1,143 @@
+package codegen
```
raphael: There should be tests for the different types as well - in particular slices (Array) once they are supported.

rest/codegen/paths.go
michaelboke: Cosmetics it looked nice, but just adding a number for the alternatives is maybe a better idea with code generation in mind. I will change it.

raphael: I don't have a strong opinion to be honest, either way is fine by me - was just curious about the rationale, cosmetics is a valid point :)

rest/codegen/paths.go
michaelboke: Do you have a suggestion?

rest/codegen/paths.go
michaelboke: How will a path with a slice look like

Something like this
/test/1,2,3,4,5
/test/slice,is,here/

Than we could introduce a dynamic printf pattern "%d,%d,%d", or we join the param values first so the printf pattern will stay the same.

What do you prefer or suggest?

rest/codegen/paths.go
michaelboke: Good idea, it would be a combination of GoTypeRef and GoTypeName, should i create a stub in codegen?

rest/codegen/paths.go
raphael: Maybe "`r.API.Name` HTTP request path constructors"?

rest/codegen/paths.go
raphael: yes please

codegen/funcs.go
```diff
@@ -155,3 +156,59 @@ func KebabCase(name string) string {
...
+	return tname
+}
+
+// GoTypeName returns the Go type name for a data type.
+func GoTypeName(dt design.DataType) string {
+	switch actual := dt.(type) {
+	case design.Primitive:
+		return GoNativeType(dt)
+	case *design.Array:
+		return "[]" + GoTypeRef(actual.ElemType.Type)
```
raphael: Please add a TBD comment that other types need to be added here (map, object, user type, etc.)

codegen/funcs.go
```diff
@@ -155,3 +156,59 @@ func KebabCase(name string) string {
...
+	default:
+		panic(fmt.Sprintf("goa bug: unknown type %#v", actual))
+	}
+}
+
+// GoNativeType returns the Go built-in type from which instances of provided datatype can be initialized.
+func GoNativeType(t design.DataType) string {
```
raphael: Same as above add a line to the header comment about needing to support other types.

rest/codegen/paths.go
```diff
@@ -2,24 +2,57 @@ package codegen
...
 	"text/template"
 
 	"goa.design/goa.v2/codegen"
 	goadesign "goa.design/goa.v2/design"
 	"goa.design/goa.v2/rest/design"
 )
 
+// note: conversion array to string path https://play.golang.org/p/0QHmyJeFhR
+
 const pathT = `{{range $i, $route := .Routes}}
```
raphael: isn't the initial newline going to cause extra blank lines in the generated code?

rest/codegen/paths.go
```diff
@@ -2,24 +2,57 @@ package codegen
...
 }
-{{end}}
+{{end -}}
+
+{{- define "arguments" -}}
+{{range $i, $arg := . -}}
+{{if ne $i 0}}, {{end}}{{.Name}} {{goTypeRef .Type}}
+{{- end}}
+{{- end}}
+
```
raphael: These blank lines get generated no?

rest/codegen/paths.go
michaelboke: you are right i'm going to fix that

rest/codegen/paths.go
michaelboke: The trim marker does an incredible job here, the maker in the "{{- define" makes sure the whitespace is trimmed. So we do not need to glue everything together anymore.

rest/codegen/paths.go
```diff
@@ -52,17 +93,20 @@ type (
...
 	Parse(pathT))
 
 // PathWriter returns the path generators writer.
-func PathWriter(r *design.RootExpr) codegen.FileWriter {
+func PathWriter(api *goadesign.APIExpr, r *design.RootExpr) codegen.FileWriter {
+	title := fmt.Sprintf("%s HTTP request path constructors", api.Name)
 	sections := []*codegen.Section{
-		codegen.Header("", "http", []*codegen.ImportSpec{
-			{Path: "context"},
+		codegen.Header(title, "http", []*codegen.ImportSpec{
```
michaelboke: Are there any plans to stripout unused imports after generation? 
For example running `goimports` on every file after generation.

Because i include all the imports i think we would need. And in some cases not all the imports are needed. 
The scenarios where this will break at compile time:
- Simple paths with no params (no import needed)
- Paths with non sliced params (fmt needed)
- Paths with slices of string or interface (fmt, net/url, strings)
- Paths with slices of numbers or boolean (fmt, strconv, strings)

If there are no plans, we need to figure out a way to determine the required imports.

rest/codegen/paths_test.go
```diff
@@ -50,15 +50,125 @@ func ShowAccountPath3(id int32, view string) string {
 }
 
 `
+
+		pathWithStringSliceParam = `
+// ShowAccountPath returns the URL path to the Account service Show HTTP endpoint.
+func ShowAccountPath(slice_string []string) string {
+	encodedslice_string := make([]string, len(slice_string))
+	for i, v := range slice_string {
+		encodedslice_string[i] = url.QueryEscape(v)
```
michaelboke: I really dislike the variable name generation for the arguments names.
Also its open to create invalid variable names. For example: a path param with the name `slice-string` would break at compile time.
This should be translated into `sliceString` and `encodedSliceString`

Probably we would need to introduce the function `Goify` here to create safe variable names.
What are your thoughts?

rest/codegen/paths.go
raphael: I was referring to the blank lines in between each block or are these also trimmed?

rest/codegen/paths.go
raphael: Yes `v1` already does that and `v2` will as well. We cannot use `goimport` directly because we don't want to add imports (the added imports are not guaranteed to be correct). However there is code that removes extra imports.

rest/codegen/paths_test.go
raphael: Yes we need `Goify`.

rest/codegen/paths.go
michaelboke: Yes those are too trimmed because of the trim markers.

rest/codegen/paths_test.go
michaelboke: Ok i will add it

codegen/types.go
```diff
@@ -0,0 +1,268 @@
...
+		switch actual.Kind() {
+		case design.BooleanKind:
+			return "bool"
+		case design.Int32Kind:
+			return "int32"
+		case design.Int64Kind:
+			return "int64"
+		case design.UInt32Kind:
+			return "uint32"
+		case design.UInt64Kind:
```
raphael: The case is missing `IntKind` and `UIntKind`

michaelboke: Ok i added it, also added them to the path template

raphael:
Thank you!

Issue #1074, opened 2017-02-24
Generate swagger consumes/produces for every action


ZandorZ:
I have an action that returns mimetype "application/vnd.myapp.user+json; type=collection" 
defined it in dsl by a collectionof MediaType("application/vnd.myapp.user+json")

Is there a way to define Consumes/Produces in every Action , not only on API ?
I tried it but it says: Invalid use of Produces in resource "user" action "list"

i need a generate swagger file that has defined the respective mimetype it actually produces
something like this:

```yaml
/users:
    get:
      description:
        Returns a list users
      consumes:
        - application/vnd.myapp.user+json
      produces:
        - application/vnd.myapp.user+json; type=collection
....
```

raphael:
Hmm I don't see a reason not to always add the operation specific `Produces` field. Thanks for raising the issue!

raphael:
I just merged code that sets the `Produces` field for each operation if it differs from the API level field. Can you validate and close the ticket if that does what you need?

ZandorZ:
```golang

package design

import (
	. "github.com/goadesign/goa/design"
	. "github.com/goadesign/goa/design/apidsl"
)

var _ = API("myapp", func() {
	Title("My App API")
	Description("My App API")
	Host("localhost:8080")
	BasePath("/api")
	Scheme("http")
	Origin("*", func() {
		Methods("GET", "PUT", "POST")
	})
})

//User
var _ = Resource("user", func() {
	Description("User resources")
	BasePath("/user")

	//list
	Action("list", func() {
		Routing(GET(""))
		Description("Return users list")
		Response(OK, CollectionOf(UserType))
		Response(BadRequest, ErrorMedia)
		Produces("application/json")
	})

....
```


I'm still getting the following error:
invalid use of Produces in resource "user" action "list"

raphael:
The generator sets it by looking at the content or media type of the response, no need to specify it with `Produces`. Just re-generate with latest.

raphael:
Also note that you can override the content type with `ContentType` in the [media type definition](https://goa.design/reference/goa/design/apidsl/#func-mediatype-a-name-apidsl-mediatype-a).

ZandorZ:
Nice. Thank you

PR #1111, opened 2017-03-13
Generate Produces on operations when necessary

raphael:
That is when the set of mime types corresponding to the operation
responses differs from the overall API 'Produces' set.


raphael:
Issue #1074

Issue #1107, opened 2017-03-09
Codegen: HTTP server


michaelboke:
Code generator for  HTTP server.
Example:
https://github.com/goadesign/goa/blob/v2/examples/account/gen/transport/http/account_server.go

PR #1116, opened 2017-03-14
[WIP] http server codegen

michaelboke:
todo encode and decoder, header mapping

This pr is for project issue #1107


rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+		"statusConst": statusCodeToHttpConst,
+	}).Parse(serverT))
+
 // ServerWriters returns the server HTTP transport writers.
-func ServerWriters(r *design.RootExpr) (ws []codegen.FileWriter) {
-	return
+func ServerWriters(api *goadesign.APIExpr, root *design.RootExpr) []codegen.FileWriter {
+	fw := make([]codegen.FileWriter, len(root.Resources))
+	for i, r := range root.Resources {
+		title := fmt.Sprintf("%s server HTTP server transport", api.Name)
```
raphael: remove the second occurrence of `server`

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+		sections := []*codegen.Section{
+			codegen.Header(title, "http", []*codegen.ImportSpec{
+				{Path: "fmt"},
+				{Path: "io"},
+				{Path: "net/http"},
+				{Path: "strconv"},
+				{Path: "github.com/dimfeld/httptreemux"},
+				{Path: "goa.design/goa.v2"},
+				{Path: "goa.design/goa.v2/examples/account/gen/endpoints"},
```
raphael: These paths need to be computed :)

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
 	"goa.design/goa.v2/codegen"
+	goadesign "goa.design/goa.v2/design"
 	"goa.design/goa.v2/rest/design"
+	"net/http"
 )
 
+//todo: in encode headers and mappings
+//todo: in decode payload, query and paths
+
+const serverT = `
```
raphael: Just curious: what's the benefit of doing these with inline defines vs. different string constants? If we are going to do it this way then we should do it this way in all writers.

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+{{define "handlers_struct" -}}
+// {{.Name}}Handlers lists the {{.Name}} service endpoint HTTP handlers.
+type {{.Name}}Handlers struct {
```
raphael: I was thinking that instead of doing the goifys and transforms inline (and potentially forgetting to do them like here) it would be better to build an intermediary data structure that contains all the values that the template can readily use. That was the idea behind the `data` field of the writers. So instead of passing the resource directly as data we should build a data structure that is convenient for the template (i.e. identify the variables that the template requires and compute their value in code then build the input data structure with the results).

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+// New{{.Name}}Handlers instantiates HTTP handlers for all the {{.Name}} service endpoints.
+func New{{.Name}}Handlers(
+	e *endpoints.{{.Name}},
+	dec rest.RequestDecoderFunc,
+	enc rest.ResponseEncoderFunc,
+	logger goa.Logger,
+) *{{.Name}}Handlers {
+	return &{{.Name}}Handlers{
+		{{- range .Actions}}
+		{{.Name}}: NewCreate{{$.Name}}Handler(e.{{.Name}}, dec, enc, logger),
```
raphael: `Create` shouldn't be hard-coded. Same comment as above (and throughout the rest): compute the name of the handler constructors in the writer code so that it doesn't have to be done here. The input data structures could have a `Handlers` field with is a slice of `HandlerData` data structures which would define e.g. a `Constructor` field that contains the constructor name so that the template would simply do `{{ .Name }}: {{ .Constructor}}(e.{{ .EndpointName }}, ...`.

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+	goadesign "goa.design/goa.v2/design"
 	"goa.design/goa.v2/rest/design"
+	"net/http"
 )
 
+//todo: in encode headers and mappings
+//todo: in decode payload, query and paths
+
+const serverT = `
+{{define "handlers_struct" -}}
```
raphael: Just to be consistent with the rest of the code base please add a blank space after the start delimiters and before the end delimeters for example `{{ define "handlers_struct" -}}` or `{{ .Name }}`). Thanks!

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+// The middleware is mounted so it executes after the request is loaded and thus
+// may access the request state via the rest package ContextXXX functions.
+func New{{.Name}}{{.Resource.Name}}Handler(
+	endpoint goa.Endpoint,
+	dec rest.RequestDecoderFunc,
+	enc rest.ResponseEncoderFunc,
+	logger goa.Logger,
+) http.Handler {
+	var (
+		decodeRequest  = {{.Name}}{{.Resource.Name}}DecodeRequest(dec)
```
raphael: We should treat the case of an empty decoder specially (if the endpoint has no `Payload` type or the `Payload` type is `design.Empty`). In this case there is no need to decode and thus no need to declare the decoder.

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+		encodeResponse = {{.Name}}{{.Resource.Name}}EncodeResponse(enc)
+		encodeError    = {{if .HTTPErrors}}{{.Name}}{{.Resource.Name}}{{end}}EncodeError(enc, logger)
+	)
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		payload, err := decodeRequest(r)
```
raphael: See above

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+		}
+	})
+}
+{{end}}
+
+{{define "decode_request" -}}
+// {{.Name}}{{.Resource.Name}}DecodeRequest returns a decoder for requests sent to the
+// create {{.Resource.Name}} endpoint.
+func {{.Name}}{{.Resource.Name}}DecodeRequest(decoder rest.RequestDecoderFunc) DecodeRequestFunc {
+	return func(r *http.Request) (interface{}, error) {
```
raphael: That shouldn't return `interface{}`

rest/codegen/server.go
```diff
@@ -1,11 +1,288 @@
...
+		{{- end}}
+		{{- if gt (len .PathParams.Type) 0}}
+		params := httptreemux.ContextParams(r.Context())
+		var (
+			{{range $name, $attr := .PathParams.Type}}
+			{{goify $name false}} {{goTypeRef .Type}}
+			{{end}}
+		)
+		{
+			orgIDRaw := params["org_id"]
```
raphael: That needs to be templatized

rest/codegen/server.go
michaelboke: It is just a convenience, defining them as a separate templates and feeding the to the template parse with a name does the same thing. Only some more lines of code

rest/codegen/server.go
michaelboke: I do agree on that, i slowly got the feeling my template was getting too complicated to figure out the internals of the expression. Also it was getting less verbose on the tags.

rest/codegen/server.go
michaelboke: So for every section i create a different data structure. 

Such as : handlersData, constructorData, mountData, actionMountData, actionConstructorData, decoderData, encoderData and errorEncoderData. 

Is this correct?

raphael: At a high level that seems to make sense: caveat I haven't actually implemented a generator that used that scheme so we may find issues with it that I don't see now. Overall though seems like a good guideline.

rest/codegen/server.go
michaelboke: I do agree.

But what happens when i want to override the decoder with my own custom decoder and payload data. Than i should also override the constructor because the bindings are not there.

rest/codegen/server.go
michaelboke: Ok this could be the generated payload type.

rest/codegen/server.go
michaelboke: yes its on the todo

rest/codegen/server.go
raphael: I wouldn't worry too much about that use case, remember that this is the transport layer. If there's some default you want to set then you can also do it in the endpoint.

codegen/generators/server/generator.go
```diff
@@ -0,0 +1,30 @@
+/*
```
michaelboke: I have no idea why this file comes up as a new file, probably a remaining file after a merge/rebase from v2 branch. Is this file safe to be removed from this PR ?

michaelboke:
Not there yet,

TODO check the response encoder, headers and mapping from the response model.
TODO request decoder support for slices from query and path params 
TODO write extra checks for the decoder primitive types string/int32/64/float etc.

Did i left something out what is needed ?

codegen/generators/server/generator.go
raphael: yes safe to remove. It's been replaced by https://github.com/goadesign/goa/blob/v2/codegen/generators/server.go

raphael:
In terms of area to cover this seems to be it. There's the work of updating to the latest codegen package (see https://github.com/goadesign/goa/blob/v2/examples/account/gen/endpoints/account.go for an example) and updating the template code to rely on pre-processed data rather than calling `Goify` and co. Also need to be careful about each place where a service, action or parameter name occur and assess whether it needs to be goified (i.e. use a different field from the data struct). The convention I've been using in the endpoint file is `Name` for the design name and `VarName` for the goified version.

codegen/files/rest/server.go
```diff
@@ -1,11 +1,587 @@
...
+		{{- if or .Payload.HasPathParams }}
+		params := httptreemux.ContextParams(r.Context())
+		{{- end }}
+		var (
+			{{- range .Payload.AllParams }}
+			{{ .VarName }} {{goTypeRef .Type }}
+			{{- end }}
+		)
+{{ range .Payload.QueryParams }}
+	{{- if eq .Type.Name "string" }}
```
raphael: Might be a bit better to compare the `.Type,Kind` to `design.StringKind`

codegen/files/rest/server.go
```diff
@@ -1,11 +1,587 @@
...
+		{{ .VarName }} = make({{ goTypeRef .Type }}, len({{ .VarName }}RawSlice))
+		for i, rv := range {{ .VarName }}RawSlice {
+			{{- template "type_slice_conversion" . }}
+		}
+	{{- else }}
+		{{- template "type_conversion" . }}
+	{{- end }}
+{{- end }}
+{{- define "type_conversion" }}
+	{{- if eq .Type.Name "string" }}
```
raphael: Same as above for all of these, please use the `Kind` to compare

codegen/files/rest/server.go
```diff
@@ -1,11 +1,587 @@
...
+type {{ .HandlersStruct }} struct {
+	{{- range .ActionData }}
+	{{ .EndpointName }} http.Handler
+	{{- end }}
+}
+`
+
+const serverConstructorT = `{{ printf "%s instantiates HTTP handlers for all the %s service endpoints." .Constructor .ServiceName | comment }}
+func {{ .Constructor }}(
+	e *endpoints.{{ .ServiceName }},
```
michaelboke: I already goified that name in code

codegen/files/rest/server.go
```diff
@@ -1,11 +1,587 @@
...
+
+const serverHandlerT = `{{ printf "%s configures the mux to serve the \"%s\" service \"%s\" endpoint." .MountHandler .ServiceName .EndpointName | comment }}
+func {{ .MountHandler }}(mux rest.ServeMux, h http.Handler) {
+	{{- range .Routes }}
+	mux.Handle("{{ .Method }}", "{{ .Path }}", h)
+	{{- end }}
+}
+`
+
+const serverHandlerConstructorT = `{{ printf "%s creates a HTTP handler which loads the HTTP request and calls the \"%s\" service \"%s\" endpoint." .Constructor .ServiceName .EndpointName | comment }}
```
michaelboke: i just played around a bit, maybe this snippet will help to make the comment function wrap at 80 characters, without splitting words in half.

https://play.golang.org/p/iHXUWlmU2s

Please check i could create a separate PR for this with tests, unless you know a better way to split the comments

codegen/files/rest/server.go
```diff
@@ -1,11 +1,587 @@
...
+			return
+		}
+
+		res, err := endpoint(r.Context(), payload)
+
+		if err != nil {
+			encodeError(w, r, err)
+			return
+		}
+		if err := encodeResponse(w, r, res); err != nil {
```
michaelboke: Ok i will not render the encodeResponse if there are no responses defined. But i do not think we could check on design.Empty for the body because we can still set some headers. Otherwise we should check if body and header is empty.

codegen/files/rest/server.go
raphael: Thanks for putting that together! This seems like a good initial version, we'll see if it needs tweaking as real world example come up.

codegen/files/rest/server.go
raphael: So this touches on something that is quite fundamental and that is whether goa should enforce that the attributes used to define the HTTP endpoints params, bodies and headers must be a subset of the (non-http) endpoint payload and result types.

That's why I wrote *endpoint* result and not HTTP response. If we go with "all attributes used to define HTTP elements must be defined in the corresponding endpoint type (payload for HTTP request elements, result for HTTP response elements) then simply checking that the endpoint result type is `design.Empty` is enough to know that the response body is empty. Otherwise need to check whether all the attributes of the endpoint result are used to define response headers and if not the remaining attributes define the response body. See the DSL spec for examples.

Another approach would consist of making it possible to define attributes in the HTTP DSL that is not defined in the Endpoint DSL. There is a major drawback to this approach though which is that it makes impossible to generate the code that maps the service interface types to the HTTP types. The generated code may not be able to completely form the HTTP response because it's missing the extra fields that were HTTP specific from the service result type. The user would have to modify the generated type to initialize these values. Stepping back a little this breaks the nice layer abstraction.

So at this point my preference would be for the DSL engine to enforce that attributes used to define HTTP elements exist in the corresponding non-HTTP type. There may be extra validations / examples defined in the HTTP specific attributes but the names must match attributes from the payload (request) or result (response) type.

codegen/files/rest/server.go
raphael: OK then the version used in the comment is wrong :) Might be good to be consistent and follow the `X` vs `VarX` convention for the non-goified and goified versions.

raphael:
Superseded by #1191

PR #1118, opened 2017-03-15
Verify no error was returned during generation.

brycereitano:


raphael:
Thank you! can you please make a PR against master as well?

Issue #1117, opened 2017-03-15
gen_controller: Specify app package path


brycereitano:
Currently you can specify the package name, and will assume you have an `app` package at the present working directory. Ideally, you should be able to specify where the `app` package resides. Helpful when you are changing the output directory of `gen_controller`, possibly `gen_main` as well.

raphael:
The `--app-pkg` flag of the `controller` command now accepts a Go import path. The import path can be absolute or relative so that the default is still just `app`.

PR #1119, opened 2017-03-15
Fix handling of --app-pkg flag

raphael:
In goagen controller command. The flag now accepts a Go import path
and generates the import accordingly


raphael:
Fix #1117

PR #1127, opened 2017-03-17
Generate tests correctly when headers are defined only on resource

### LINKED ISSUE #1125: Goagen generates faulty tests when headers are required outside of actions
Similar to #915 but not quite.

If I add a required header on a resource:

```
var _ = Resource("users_v1", func() {
	BasePath("/v1/users")
	Description("Provides a set of endpoints to manage the users")

	// Headers
	Headers(func() { // Headers describe relevant action headers
		Header("Authorization", String)
		Required("Authorization")
	})

	Action("show", func() {
		Description("Retrieve the user identified by the given ID. The special id 'me' can be used to obtain info about the current user (the one who provided the auth token)")
		Routing(GET("/byID/:id"))
```

the generated test file doesn't contain the header parameter

It works correctly if the Headers section is inside an action.

matteosuppo:
Fix #1125

Signed-off-by: Matteo Suppo matteo.suppo@gmail.com


raphael:
Thank you!

PR #1134, opened 2017-03-21
Pointer Code Fix

JKhawaja:
The code here will allow for generated app code to compile when using the Metadata struct:field:type tag.


JKhawaja:
Not sure why the build is failing. Tests passed for me.

PR #1135, opened 2017-03-21
Generated App Code -- Pointer Problem Fix

JKhawaja:
Source fixes for generating code with pointers properly for attributes using metadata `struct:field:type` tag.


goagen/codegen/types.go
```diff
@@ -113,7 +113,11 @@ func goTypeDefObject(obj design.Object, def *design.AttributeDefinition, tabs in
 		typedef := ""
 		if tname, ok := field.Metadata["struct:field:type"]; ok {
 			if len(tname) > 0 {
-				typedef = tname[0]
+				if (field.Type.IsPrimitive() && private) || field.Type.IsObject() || def.IsPrimitivePointer(name) {
```
raphael: Might be good to remove the copy/paste here and define a variable below `typedef := ""` on line 113 that can be used both here and below. Something like:0
```go
pointer := ""
if (field.Type.IsPrimitive() && private) || field.Type.IsObject() || def.IsPrimitivePointer(name) {
    pointer = "*"
}
```
then this code becomes `typedef = pointer + tname[0]` and the code below (line 125) becomes `typedef = pointer + typedef`

raphael:
Thank you, looks good! I added a comment in the commit to clean up a little bit the code. I'll merge once it's addressed. Thanks again.

raphael:
Awesome, thank you!

raphael:
Does this need to be backported to `v1`?

JKhawaja:
Yes, it does. But I am not sure I want to submit a backport PR for `v1` until I figure out how to make sure that attributes that use the metadata tag do not also generate validation code. That way it will be a single Metadata Tag `struct:field:type`  `v1` PR.

Issue #1106, opened 2017-03-09
Codegen: Service writer


tchssk:
Implemented in package https://github.com/goadesign/goa/tree/v2/codegen/writers.

PR #1139, opened 2017-03-23
Implement Service()

tchssk:
#1106

### DIFF
--- a/codegen/files/service.go
+++ b/codegen/files/service.go
@@ -1,18 +1,22 @@
 package files
 
 import (
+	"path/filepath"
+	"text/template"
+
 	"goa.design/goa.v2/codegen"
 	"goa.design/goa.v2/design"
 )
 
 type (
-	// serviceData contains the data necessary to render the service template.
+	// serviceData contains the data necessary to render the service
+	// template.
 	serviceData struct {
-		// Name is the service interface name
+		// Name is the service name.
 		Name string
-		// Description is the service description.
-		Description string
-		// Methods list the interface methods.
+		// VarName is the service struct name.
+		VarName string
+		// Methods lists the service struct methods.
 		Methods []*serviceMethod
 	}
 
@@ -20,27 +24,113 @@ type (
 	serviceMethod struct {
 		// Name is the method name.
 		Name string
-		// Description is the method description.
-		Description string
-		// Payload is the payload type.
-		Payload *design.UserTypeExpr
-		// Result is the result type.
-		Result *design.UserTypeExpr
+		// Payload is the payload of the method.
+		Payload servicePayload
+		// HasPayload is true if the payload type is not empty.
+		HasPayload bool
+		// Result is the result of the method.
+		Result serviceResult
+		// HasResult is true if the result type is not empty.
+		HasResult bool
+	}
+
+	servicePayload struct {
+		// Name is the payload name.
+		Name string
+		// Fileds lists the payload fields.
+		Fields map[string]string
+	}
+
+	serviceResult struct {
+		// Name is the result name.
+		Name string
+	}
+
+	// serviceFile is the codgen file for a given service.
+	serviceFile struct {
+		service *design.ServiceExpr
 	}
 )
 
-// Service returns the file for the given service.
-func Service(api *design.APIExpr, service *design.ServiceExpr) codegen.File {
-	return nil
+// serviceTmpl is the template used to render the body of the service file.
+var serviceTmpl = template.Must(template.New("service").Parse(serviceT))
+
+// Service returns the service file for the given service.
+func Service(service *design.ServiceExpr) codegen.File {
+	return &serviceFile{service}
 }
 
-// serviceT is the template used to write a service definition.
-const serviceT = `
-type (
-	{{ comment .Description 1 }}
-	{{ .Name }} interface {{{ range .Methods }}
-		{{ comment .Description 2 }}
-		{{ .Name }}(context.Context, {{ gotypename .Payload }}) ({{ gotypename .Result }}, error)
+// Sections returns the service file sections.
+func (s *serviceFile) Sections(genPkg string) []*codegen.Section {
+	var (
+		data *serviceData
+	)
+	{
+		methods := make([]*serviceMethod, len(s.service.Endpoints))
+		for i, v := range s.service.Endpoints {
+			fields := make(map[string]string)
+			if o, ok := v.Payload.Attribute().Type.(design.Object); ok {
+				for key, value := range o {
+					fields[key] = codegen.GoTypeName(value.Type)
+				}
+			}
+			methods[i] = &serviceMethod{
+				Name: codegen.Goify(v.Name, true),
+				Payload: servicePayload{
+					Name:   codegen.Goify(v.Payload.Name(), true),
+					Fields: fields,
+				},
+				HasPayload: v.Payload != design.Empty,
+				Result: serviceResult{
+					Name: codegen.Goify(v.Result.Name(), true),
+				},
+				HasResult: v.Result != design.Empty,
+			}
+		}
+		data = &serviceData{
+			Name:    s.service.Name,
+			VarName: codegen.Goify(s.service.Name, true),
+			Methods: methods,
+		}
 	}
+
+	var (
+		header, body *codegen.Section
+	)
+	{
+		header = codegen.Header(s.service.Name+"Services", "services",
+			[]*codegen.ImportSpec{
+				&codegen.ImportSpec{Path: "context"},
+				&codegen.ImportSpec{Path: "goa.design/goa.v2"},
+				&codegen.ImportSpec{Path: genPkg + "/services"},
+			})
+		body = &codegen.Section{
+			Template: serviceTmpl,
+			Data:     data,
+		}
+	}
+
+	return []*codegen.Section{header, body}
+}
+
+// OutputPath is the path to the generated service file relative to the output
+// directory.
+func (s *serviceFile) OutputPath(reserved map[string]bool) string {
+	svc := codegen.SnakeCase(s.service.Name)
+	return UniquePath(filepath.Join("services", svc+"%d.go"), reserved)
+}
+
+// serviceT is the template used to write an service definition.
+const serviceT = `type (
+	// {{ .VarName }} is the {{ .Name }} service interface.
+	{{ .VarName }} interface {
+{{ range .Methods }}		// {{ .Name }} implements the {{ .Name }} endpoint.
+		{{ .Name }}(context.Context{{ if .HasPayload }}, *{{ .Payload.Name }}{{ end }}) {{ if .HasResult }}({{ .Result.Name }}, error){{ else }}error{{ end }}
+{{ end }}	}
+{{ range .Methods }}{{ if .HasPayload }}
+	{{ .Payload.Name }} struct {
+{{ range $key, $att := .Payload.Fields }}		{{ $key }} {{ $att }}
+{{ end }}	}
+{{ end }}{{ end -}}
 )
 `
--- a/codegen/files/service.go
+++ b/codegen/files/service.go
@@ -46,6 +46,8 @@ type (
 	serviceResult struct {
 		// Name is the result name.
 		Name string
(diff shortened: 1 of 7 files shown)


codegen/files/service.go
```diff
@@ -1,46 +1,136 @@
...
+func (s *serviceFile) OutputPath(reserved map[string]bool) string {
+	svc := codegen.SnakeCase(s.service.Name)
+	return UniquePath(filepath.Join("services", svc+"%d.go"), reserved)
+}
+
+// serviceT is the template used to write an service definition.
+const serviceT = `type (
+	// {{ .VarName }} is the {{ .Name }} service interface.
+	{{ .VarName }} interface {
+{{ range .Methods }}		// {{ .Name }} implements the {{ .Name }} endpoint.
```
raphael: Seems like we should also have `VarName` and `Name` for the method? so that the comment could be:
```go
// {{ .VarName }} implements the {{ .Name }} endpoint.
```

codegen/files/service_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		multipleEndpoints = design.ServiceExpr{
+			Name: "Multiple",
+			Endpoints: []*design.EndpointExpr{
+				&a,
+				&b,
+			},
+		}
+		_ = multipleEndpoints
```
raphael: just curious about what this does?

codegen/files/service.go
```diff
@@ -1,46 +1,136 @@
...
+// Sections returns the service file sections.
+func (s *serviceFile) Sections(genPkg string) []*codegen.Section {
+	var (
+		data *serviceData
+	)
+	{
+		methods := make([]*serviceMethod, len(s.service.Endpoints))
+		for i, v := range s.service.Endpoints {
+			fields := make(map[string]string)
+			if o, ok := v.Payload.Attribute().Type.(design.Object); ok {
```
tchssk: Oops thank you, replaced it. https://github.com/goadesign/goa/pull/1139/commits/c9581fe033bfaea8f32911dee13d72677c271422

codegen/files/service.go
```diff
@@ -1,46 +1,136 @@
...
+		}
 	}
+
+	var (
+		header, body *codegen.Section
+	)
+	{
+		header = codegen.Header(s.service.Name+"Services", "services",
+			[]*codegen.ImportSpec{
+				&codegen.ImportSpec{Path: "context"},
```
tchssk: Thank you. https://github.com/goadesign/goa/pull/1139/commits/76a17d34a4cb7410b88d46b1aed22b1bf3f2e5b3

codegen/files/service_test.go
tchssk: They were just debug codes. I deleted them. https://github.com/goadesign/goa/pull/1139/commits/2c93fdb8a4c82035684b2f2c2bda7dde84a68657

codegen/files/service.go
tchssk: https://github.com/goadesign/goa/pull/1139/commits/5d6a6185f39d30e821b02b5debc18cc332b245b8

raphael:
Thanks for all the changes, looks good. Are you working on recursively tracking the types that need to be generated? seems like it's the last item remaining.

tchssk:
Yes, I'm working on that.

It seems we have to improve `GoTypeName` and `GoNativeType` before using `Walk()` so I'll add some commits first.

tchssk:
I'm going forward step by step so I pushed commits about `GoNativeType` first. Please take a look these commits.

- https://github.com/goadesign/goa/pull/1139/commits/de8d8c60cc9b91e59119e59787a031e60ca8d355 adds tests for `GoNativeType()`.
- https://github.com/goadesign/goa/pull/1139/commits/6587e5cfd321d31691240c9e3ba85e99bc0429a6 supports Map, Object and UserType in `GoNativeType()`.

raphael:
The `GoNativeType` implementation and tests look good. Something I've been wondering is whether we should add back `DateTime` as a type in `v2`. I was thinking initially that it could just be `String` with a special `datetime format`. But after thinking it through and looking at real APIs it seems having a DateTime as first class type in the design might be better. What do you think?

tchssk:
`v1` has `DateTimeKind` and has `DateTime` as `Primitive` so we can design `v2` as same way as that. Maybe it's useful for OpenAPI (Swagger) but `v2` supports also gRPC.

It's a big topic for `v2`. Do we open a new issue and discuss about that? At this point, I can continue this PR to support only existing types and we can add more types support as needed in the future.

tchssk:
I pushed a new commit.

- 3264349 tracks the types recursively in `Service()`.

tchssk:
In the way you shown, I could track only top level fields (means `Parent` in your example) in the payload. It seems we have to track the types recursively using `ut.Walk()`.

We also might need a function like `GoTypeDef()` in `v1` to write each type definitions.

raphael:
Yes you're right - it has to be recursive. And also agree on needing something like `GoTypeDef`. Last commit looks great BTW.

tchssk:
I pushed new commits.

41a41a2 implements `GoTypeDef()`.
9c6ddd8 renders results in `Service()`.
47757ca tracks the types recursively in `Service()`.

I didn't add an argument to indent codes in `GoTypeDef()` but the codes will be `gofmt` ed, right?

raphael:
Looks good overall! a couple of comments:

**Cycles**

I think the walker code needs to handle cycles, consider this design:
```go
var Parent = Type("parent", func() {
    Attribute("c", Child)
})

var Child = Type("child", func() {
    Attribute("p", "parent")
})
```
This is a legal definition that will cause the walker to recurse infinitely. I think it needs to recurse only if the type is not already in the hash.

** Indenting **

Yes the code gets `gofmt`ed however the idea so far has been to render code that was properly formatted and use `gofmt` as a last resort. However it might be better to give up on trying to generate code that is properly indented and just rely on `gofmt`. The added complexity involved in generating the proper level of indentation (esp. with recursive algorithms) may not be worth it. However if we do that we should do it consistently everywhere.

So @michaelboke @tchssk let's do a quick vote here :) do we indent the generated code using `gofmt` or via the templates? I vote `gofmt`.

tchssk:
`gofmt` +1

tchssk:
>Cycles

ea09cbe avoids infinite recursion in `Service()`.

michaelboke:
Gofmt but keep formatting as good as possible in templates. At least I'm
trying
Op do 30 mrt. 2017 om 18:09 schreef Taichi Sasaki <notifications@github.com>

raphael:
Agree on keeping formatting as good as possible. The idea here is mainly to remove the tracking of indentation - that is remove the `depth` argument from all the template functions and instead always format properly but with a level of indentation of 0.

tchssk:
I tried to write the codes from the idea.

f0ed658 removes indents and use `gofmt` for tests.

tchssk:
Once it seems finished. or Do I have anything to do more?

codegen/types.go
```diff
@@ -108,6 +108,48 @@ var (
...
+			keyDef = "*" + keyDef
+		}
+		elemDef := GoTypeDef(actual.ElemType.Type)
+		if design.IsObject(actual.ElemType.Type) {
+			elemDef = "*" + elemDef
+		}
+		return fmt.Sprintf("map[%s]%s", keyDef, elemDef)
+	case design.Object:
+		return goTypeDefObject(actual)
+	case *design.UserTypeExpr:
```
raphael: Please use `design.UserType` and not `design.UserTypeExpr` so that it also works for media types. Would be good to have tests for this function that tests all the different types (primitive, array, map, object, user and media types).

codegen/types.go
```diff
@@ -125,13 +167,20 @@ func GoTypeName(dt design.DataType) string {
 		return GoNativeType(dt)
 	case *design.Array:
 		return "[]" + GoTypeRef(actual.ElemType.Type)
+	case *design.Map:
+		return fmt.Sprintf("map[%s]%s", GoNativeType(actual.KeyType.Type), GoNativeType(actual.ElemType.Type))
+	case design.Object:
+		return "map[string]interface{}"
+	case *design.UserTypeExpr:
```
raphael: Same as above, please use `design.UserType`.

codegen/types.go
```diff
@@ -165,6 +214,12 @@ func GoNativeType(t design.DataType) string {
 		}
 	case *design.Array:
 		return "[]" + GoNativeType(actual.ElemType.Type)
+	case *design.Map:
+		return fmt.Sprintf("map[%s]%s", GoNativeType(actual.KeyType.Type), GoNativeType(actual.ElemType.Type))
+	case design.Object:
+		return "map[string]interface{}"
+	case *design.UserTypeExpr:
```
raphael: ditto

codegen/types.go
```diff
@@ -143,7 +143,7 @@ func goTypeDefObject(o design.Object) string {
 	var ss []string
 	ss = append(ss, "struct {")
 	o.WalkAttributes(func(name string, at *design.AttributeExpr) error {
-		ss = append(ss, fmt.Sprintf("%s %s", name, GoTypeName(at.Type)))
+		ss = append(ss, fmt.Sprintf("\t%s %s", name, GoTypeName(at.Type)))
```
michaelboke: I do not think this you should add a tab in here. Let the template be responsible for this

tchssk: Do you mean should this returns like

```go
struct {
IntField int
StringField string
}
```

instead of

```go
struct {
	IntField int
	StringField string
}
```

?

michaelboke: Ah wait, ignore my first comment. I missed the part where you were building a structure definition here.
Sorry for the confusion

tchssk: No problem, thank you :)

tchssk:
a97bc6be uses `UserType` instead of `UserTypeExpr`.

codegen/types.go
tchssk: a97bc6be

codegen/types.go
tchssk: a97bc6be

codegen/types.go
tchssk: a97bc6be

codegen/types.go
```diff
@@ -132,8 +132,8 @@ func GoTypeDef(dt design.DataType) string {
 		return fmt.Sprintf("map[%s]%s", keyDef, elemDef)
 	case design.Object:
 		return goTypeDefObject(actual)
-	case *design.UserTypeExpr:
-		return GoTypeName(actual)
+	case design.UserType:
+		return GoTypeName(actual.Attribute().Type)
```
raphael: I think you can just call `GoTypeName(actual)` no?

tchssk: hmm it doesn't work fine.

```go
--- FAIL: TestGoTypeDef (0.00s)
	types_test.go:36: UserTypeExpr: got "UserType", expected "bool"
	types_test.go:36: MediaTypeExpr: got "application/vnd.goa.example", expected "bool"
FAIL
FAIL	goa.design/goa.v2/codegen	0.012s
```

raphael: I think the test isn't correct. `GoTypeDef` should return the type name for user types.  Also `GoTypeName` should return the `TypeName` field of the embedded `UserTypeExpr` for `MediaTypeExpr`.

tchssk: Thank you. I'll fix it.

tchssk: caab119 fixes `GoTypeDef()` about `UserType`.

tchssk:
fc7a5b1 adds tests about empty payload and empty result.

tchssk:
Travis failed with error that `WalkAttributes()` methods of design.Object not found. This seems to be caused by #1164.

raphael:
Yes this PR moved the walkers to the code generation package (`files.WalkAttributes`) because the design and DSL engine don't use it, only code generation does.

tchssk:
Shoud I rebase this branch onto v2?

raphael:
Yeah, that would probably make sense

raphael:
Actually I'll merge and rebase since I need some of the changes in this PR (the `GoTypeName` changes).

PR #1140, opened 2017-03-25
Make another pass at the paths file.

raphael:
Mainly to make sure that names of generated functions are valid Go identifiers.


raphael:
@michaelboke do you think you could review and merge? Thank you!

Issue #1137, opened 2017-03-22
Resource inheritance child inherits query string params


lockwobr:
I am running into a bug I believe were if you add parent relationship to a resource it is inheriting query params.

Example: User is parent.

     Action("show", func() {
		Routing(
			GET("/:UserId"),
		)
		Description("Retrieve account with given id.")
		Params(func() {
			Param("UserId")
			Param("view", String, "Which view of user would you like?", func() {
				Default("default")
				Enum("default", "short")
			})
		})
		Response(OK)
		Response(NotFound)
	})

Swagger yaml output were balance is child of user, and you can see view param is in there.

    /api/user/{UserId}/balance:
    get:
      description: Retrieve account with given id.
      operationId: Balance#show
      parameters:
      - description: Unique user ID
        in: path
        name: UserId
        required: true
        type: integer
      - default: default
        description: Which view of user would you like?
        enum:
        - default
        - short
        in: query
        name: view
        required: false
        type: string
      responses:
        "200":
          description: OK
          schema:
            $ref: '#/definitions/GoaBalance'
        "204":
          description: No Content
      schemes:
      - http
      summary: show Balance
      tags:
      - Balance

I guess for reference here is part of my vendor file so you track the version I am using. Might be over kill but.

       		{
			"checksumSHA1": "Sf/0bt+aQfG4ng+Y061mu37KegI=",
			"path": "github.com/goadesign/goa/cors",
			"revision": "148222cf85d4578a0cfc591bd0c943d11d3038dd",
			"revisionTime": "2017-02-22T20:45:24Z"
		},
		{
			"checksumSHA1": "kv/fDGC+gKkUH9c0dviifYmHAIk=",
			"path": "github.com/goadesign/goa/design",
			"revision": "148222cf85d4578a0cfc591bd0c943d11d3038dd",
			"revisionTime": "2017-02-22T20:45:24Z"
		},
		{
			"checksumSHA1": "4XSXzkn+NcsVpXlIIyNVhp9dehk=",
			"path": "github.com/goadesign/goa/design/apidsl",
			"revision": "148222cf85d4578a0cfc591bd0c943d11d3038dd",
			"revisionTime": "2017-02-22T20:45:24Z"
		},
		{
			"checksumSHA1": "FJv1gV2WiyoZ+lqyprF43ul2OtI=",
			"path": "github.com/goadesign/goa/dslengine",
			"revision": "148222cf85d4578a0cfc591bd0c943d11d3038dd",
[...]

PR #1144, opened 2017-03-28
Fix 'AllParams' so that it does not return parent query string params

raphael:
Only parent path parameters.


raphael:
Fix #1137

PR #1147, opened 2017-03-30
Use kebak-case for resource names in gen_client

### LINKED ISSUE #1146: Case discrepancy in generated code
I’m a newbie to goagen, proceeding incrementally from the cellar example.

There appears to be a discrepancy between various code generators in goagen regarding the case of identifiers the tool chain produces.

I have in my design.go file:
  var _ = Resource("ADC", func() {
        BasePath("/adc")
        DefaultMedia(ADCMedia)
        ...

The generated file tool/cli/commands.go contains a declaration of InitADCCommand struct.  But later on there’s a call to new(InitAdcCommand), which fails to compile.  Note the difference of the sub-word “ADC” vs “Adc”.

tchssk:
Fix #1146.

This seems a regression by #1039.

### DIFF
--- a/goagen/gen_client/cli_generator.go
+++ b/goagen/gen_client/cli_generator.go
@@ -681,7 +681,7 @@ func new{{ goify $security.SchemeName true }}Signer({{ signerSignature $security
 {{ end }}{{ end }}
 `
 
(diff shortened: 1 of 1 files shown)


raphael:
Thank you!

PR #1152, opened 2017-03-31
Fix problem not starting with appengine

tikasan:
for https://github.com/armon/go-metrics/pull/47

goa are using packages that are not permitted by GAE of GCP PaaS and can not be used.
This can be solved by not including the syscall containing code of the problem package in compilation.
I like goa, so I would like to use it on GAE as well. So I requested it

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -1,4 +1,4 @@
-// +build !js
+// +build !js,!appengine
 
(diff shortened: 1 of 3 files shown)


tikasan:
sry miss commit

PR #1153, opened 2017-03-31
Fix problem not starting with appengine

tikasan:
for https://github.com/armon/go-metrics/pull/47

goa uses packages not permitted by GAE of GCP PaaS and can not be used on GAE.
This can be solved by not including syscall containing the code of the package in question in the compilation.
I like Gore, so I would like to use it on GAE. So I requested it.

There is a problem with the dependency package, but if it is solved by goa it will be usable, so please consider.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -1,4 +1,4 @@
-// +build !js
+// +build !js,!appengine
 
 package goa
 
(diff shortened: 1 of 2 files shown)


raphael:
Thank you, good catch! Do you think you could backport the fix to the `v1` branch as well? (just branch off of `v1` and cherry-pick this commit). Thanks again!

tikasan:
Certainly it seems better to do it.
I'll try.

PR #1172, opened 2017-04-08
Add ability to use DSL when defining Hash attributes

raphael:


raphael:
Here is an example that takes advantage of the DSLs:
```go
// ValidateKey is a DSL function that defines a pattern based validation.
func ValidateKey() {
	Pattern("^foo")
}

// TypeValue is a DSL function that attaches metadata.
func TypeValue() {
	Metadata("struct:field:type", "json.RawMessage", "encoding/json")
}

// Mappings is a Hash type with validation attached to the keys and metadata attached to the values.
var Mappings = HashOf(String, String, ValidateKey, TypeValue)
```

PR #1175, opened 2017-04-11
Xray transport for v1

dominicm:
Note: This PR would break compatibility with Go1.6 since this code relies on accessing the context in http.Request.


raphael:
Thank you!

PR #1181, opened 2017-04-13
Fix setting a default value to Number

deadcheat:
Fixed func PrintVal in goagen/codegen/finalizer.go that returns int value string when setting zero as a default value to Number type.


raphael:
Great, thank you! That change broke a few tests, would you mind fixing them as well? see https://travis-ci.org/goadesign/goa/jobs/221614278

deadcheat:
yes, i'll do it soon.

deadcheat:
fixed test and squashed commits. thank you!

raphael:
Thank you!

PR #1182, opened 2017-04-13
Fix setting a default value to Number

deadcheat:
same fixtures as #1181 to branch v1.

> Fixed func PrintVal in goagen/codegen/finalizer.go that returns int value string when setting zero as a default value to Number type.


raphael:
Thank you!

PR #1184, opened 2017-04-13
Expose record context response

dominicm:


middleware/xray/segment.go
```diff
@@ -175,37 +200,41 @@ func (s *Segment) RecordResponse(resp *http.Response) {
...
+func (s *Segment) RecordContextResponse(ctx context.Context) {
+	resp := goa.ContextResponse(ctx)
+	if resp == nil {
+		return
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	if s.HTTP == nil {
```
jim-slattery-rs: maybe from this line on down could be factored out as a shared method that could also be reused by `RecordResponse`?
```go
func (s *Segment) recordResponse(status, respLength int)
```

middleware/xray/segment.go
```diff
@@ -175,37 +200,41 @@ func (s *Segment) RecordResponse(resp *http.Response) {
...
+// RecordContextResponse traces a context response if present in the context
+//
+// It sets Throttle, Fault, Error and HTTP.Response
+func (s *Segment) RecordContextResponse(ctx context.Context) {
+	resp := goa.ContextResponse(ctx)
+	if resp == nil {
+		return
+	}
+
+	s.Lock()
```
jim-slattery-rs: maybe from this line on down could be factored out as a shared method that could also be reused by `RecordResponse`?
```go
func (s *Segment) recordResponse(status, respLength int)
```

raphael:
Thank you!

PR #1206, opened 2017-04-25
Use new to initialize variables in endpoint writer

tchssk:
#1077.

This is a fix the generated codes to match the example.


raphael:
Thank you! I have more changes for the `endpoint` and `service` files that I'll push soon.

PR #1207, opened 2017-04-26
set Error in when recording

smessier:
@raphael set xray.Segment.Error to true when recording error unless Fault or Throttle already set


raphael:
Thank you! could you also make a PR against `master` please?

raphael:
nevermind :)

PR #1214, opened 2017-05-04
Added "can be used in" to the top of all DSL's

deltaskelta:
Some things I was unsure about so I noted them with a TODO. I will change them and resubmit if there is extra feedback...

Todos are on the following lines so far:

```
Desktop/goa/design/apidsl/api.go|154 col 4| // TODO: check if origin can really be used in action
Desktop/goa/design/apidsl/api.go|259 col 4| // TODO: can it really be used in Resource (no examples in docs)
Desktop/goa/design/apidsl/api.go|310 col 4| // TODO: can it really be used in Resource (no examples)
Desktop/goa/design/apidsl/api.go|437 col 4| // TODO: where is function used?
Desktop/goa/design/apidsl/media_type.go|367 col 4| // TODO: is this correct?
Desktop/goa/design/apidsl/security.go|310 col 4| // TODO: is this correct?
Desktop/goa/design/apidsl/type.go|8 col 4| // TODO: are these correct? They can be used as values inline, but they can also be used
```

### DIFF
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -8,6 +8,7 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// Files used in: Resource
 // Files defines an API endpoint that serves static assets. The logic for what to do when the
 // filename points to a file vs. a directory is the same as the standard http package ServeFile
 // function. The path may end with a wildcard that matches the rest of the URL (e.g. *filepath). If
@@ -52,6 +53,7 @@ func Files(path, filename string, dsls ...func()) {
 	}
 }
 
+// Action used in: Resource
 // Action implements the action definition DSL. Action definitions describe specific API endpoints
 // including the URL, HTTP method and request parameters (via path wildcards or query strings) and
 // payload (data structure describing the request HTTP body). An action belongs to a resource and
@@ -112,6 +114,7 @@ func Action(name string, dsl func()) {
 	}
 }
 
+// Routing used in: Action
 // Routing lists the action route. Each route is defined with a function named after the HTTP method.
 // The route function takes the path as argument. Route paths may use wildcards as described in the
 // [httptreemux](https://godoc.org/github.com/dimfeld/httptreemux) package documentation. These
@@ -126,6 +129,7 @@ func Routing(routes ...*design.RouteDefinition) {
 	}
 }
 
+// GET used in: Routing
 // GET creates a route using the GET HTTP method.
 func GET(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "GET", Path: path}
@@ -137,6 +141,7 @@ func GET(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// HEAD used in: Routing
 // HEAD creates a route using the HEAD HTTP method.
 func HEAD(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "HEAD", Path: path}
@@ -148,6 +153,7 @@ func HEAD(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// POST used in: Routing
 // POST creates a route using the POST HTTP method.
 func POST(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "POST", Path: path}
@@ -159,6 +165,7 @@ func POST(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// PUT used in: Routing
 // PUT creates a route using the PUT HTTP method.
 func PUT(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "PUT", Path: path}
@@ -170,6 +177,7 @@ func PUT(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// DELETE used in: Routing
 // DELETE creates a route using the DELETE HTTP method.
 func DELETE(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "DELETE", Path: path}
@@ -181,6 +189,7 @@ func DELETE(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// OPTIONS used in: Routing
 // OPTIONS creates a route using the OPTIONS HTTP method.
 func OPTIONS(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "OPTIONS", Path: path}
@@ -192,6 +201,7 @@ func OPTIONS(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// TRACE used in: Routing
 // TRACE creates a route using the TRACE HTTP method.
 func TRACE(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "TRACE", Path: path}
@@ -203,6 +213,7 @@ func TRACE(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// CONNECT used in: Routing
 // CONNECT creates a route using the CONNECT HTTP method.
 func CONNECT(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "CONNECT", Path: path}
@@ -214,6 +225,7 @@ func CONNECT(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// PATCH used in: Routing
 // PATCH creates a route using the PATCH HTTP method.
 func PATCH(path string, dsl ...func()) *design.RouteDefinition {
 	route := &design.RouteDefinition{Verb: "PATCH", Path: path}
@@ -225,6 +237,7 @@ func PATCH(path string, dsl ...func()) *design.RouteDefinition {
 	return route
 }
 
+// Headers can be used in: Action, Response, Resource
 // Headers implements the DSL for describing HTTP headers. The DSL syntax is identical to the one
 // of Attribute. Here is an example defining a couple of headers with validations:
 //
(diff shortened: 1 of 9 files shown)


design/apidsl/api.go
```diff
@@ -122,6 +124,7 @@ func Description(d string) {
 	}
 }
 
+// BasePath used in: API
```
raphael: Also Resource

design/apidsl/api.go
```diff
@@ -147,6 +150,8 @@ func BasePath(val string) {
 	}
 }
 
+// Origin can be used in: Resource, API, Action
+// TODO: check if origin can really be used in action
```
raphael: `Origin` can only be called in `API` and `Resource` - not `Action`

design/apidsl/api.go
```diff
@@ -194,34 +199,39 @@ func Origin(origin string, dsl func()) {
...
-// Credentials sets the allow credentials response header. Used in Origin DSL.
+// Credentials can be used in: Origin
+// Credentials sets the allow credentials response header.
 func Credentials() {
 	if cors, ok := corsDefinition(); ok {
 		cors.Credentials = true
 	}
 }
 
+// TermsOfService can be used in: Origin
```
raphael: `TermsOfService` can be used in `API` and `API` only.

design/apidsl/api.go
```diff
@@ -291,6 +306,8 @@ func License(dsl func()) {
 	}
 }
 
+// Docs can be used in: API, Action, Resource, Files
```
raphael: Not `Resource`

design/apidsl/api.go
```diff
@@ -322,13 +340,15 @@ func Name(name string) {
...
 
+// Email can be used in: Contact
 // Email sets the contact email.
 func Email(email string) {
 	if c, ok := contactDefinition(); ok {
 		c.Email = email
 	}
 }
 
+// URL can be used in: Contact, License.
```
raphael: and `Docs` - I guess the comment should be updated too.

design/apidsl/api.go
```diff
@@ -411,6 +434,7 @@ func Package(path string) {
 	}
 }
 
+// TODO: where is function used?
```
raphael: In `Consumes` and `Produces` as well

design/apidsl/api.go
```diff
@@ -552,9 +579,9 @@ func Trait(name string, val ...func()) {
 	}
 }
 
-// UseTrait executes the API trait with the given name. UseTrait can be used inside a Resource,
-// Action, Type, MediaType or Attribute DSL. UseTrait takes a variable number
-// of trait names.
+// Trait can be used in: Resource, Action, Type, MediaType, Attribute
```
raphael: Should be `UseTrait` can be used...

design/apidsl/attribute.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// Attribute can be used in: View, Type, Attribute (nested), Attributes (child),
+// MediaType
```
raphael: Not `MediaType` (has to be under `Attributes`)

design/apidsl/attribute.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// Attribute can be used in: View, Type, Attribute (nested), Attributes (child),
```
raphael: Not sure I understand the distinction between `nested` and `child`, in both cases it's in a DSL function.

design/apidsl/attribute.go
```diff
@@ -235,6 +238,7 @@ func parseAttributeArgs(baseAttr *design.AttributeDefinition, args ...interface{
 	return dataType, description, dsl
 }
 
+// Header can be used in: Headers, APIKeySecurity, JWTSecurity
```
raphael: Should document `Member` below, can appear in `Payload`, `Param` can appear in `Params`, `Default` can appear in `Attribute`

design/apidsl/attribute.go
```diff
@@ -282,6 +286,7 @@ func Default(def interface{}) {
 	}
 }
 
+// Example can be used in: Attribute
```
raphael: All the attribute functions below (`Example` to `MaxLen`) can all be used in `Attribute`, `Header`, `Param`, `HashOf` or `ArrayOf`

design/apidsl/attribute.go
```diff
@@ -522,6 +535,7 @@ func MaxLength(val int) {
 	}
 }
 
+// Required can be used in: Attribute, Attributes, Header, Headers, Payload, Type
```
raphael: Not `Attribute` or `Header`. Also `Params`.

design/apidsl/media_type.go
```diff
@@ -12,6 +12,7 @@ import (
 // Counter used to create unique media type names for identifier-less media types.
 var mediaTypeCount int
 
+// MediaType can be used in: ResponseTemplate, (also as a top level DSL)
```
raphael: Why the parenthesis around `also as a top level DSL`?

design/apidsl/media_type.go
```diff
@@ -113,6 +114,7 @@ func MediaType(identifier string, apidsl func()) *design.MediaTypeDefinition {
 	return mt
 }
 
+// Media can be used in: Response, ResponseTemplate
```
raphael: Remove the redundant comment at the end

design/apidsl/media_type.go
```diff
@@ -202,6 +206,7 @@ func TypeName(name string) {
 	}
 }
 
+// ContentType can be used in: Type, MediaType
```
raphael: Only `MediaType`

design/apidsl/media_type.go
```diff
@@ -213,6 +218,7 @@ func ContentType(typ string) {
 	}
 }
 
+// iew can be used in: MediaType
```
raphael: `View`, also `Response`.

design/apidsl/media_type.go
```diff
@@ -355,6 +364,8 @@ func Link(name string, view ...string) {
 	}
 }
 
+// TODO: is this correct?
+// CollectionOf can be used in: Media, Response, ResponseTemplate
```
raphael: There are more places, maybe it's simpler to say something like `wherever a media type can`, e.g `Attribute("foo", CollectionOf(Bar)`

design/apidsl/metadata.go
```diff
@@ -5,10 +5,10 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// Metadata can be used in: Attributes, MediaTypes, Action, Response, Resource, API
```
raphael: `MediaType` (no `s`)

design/apidsl/resource.go
```diff
@@ -5,6 +5,7 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// Resource can be used in: (it is a top level DSL)
```
raphael: Maybe just `Resource is a top level DSL`? same with all other top level DSLs.

design/apidsl/security.go
```diff
@@ -300,6 +307,8 @@ func inHeader(headerName string) {
 	dslengine.IncompatibleDSL()
 }
 
+// TODO: is this correct?
```
raphael: yes

design/apidsl/type.go
```diff
@@ -5,6 +5,10 @@ import (
 	"github.com/goadesign/goa/dslengine"
 )
 
+// TODO: are these correct? They can be used as values inline, but they can also be used
```
raphael: Just top level DSL

design/apidsl/type.go
```diff
@@ -59,6 +63,7 @@ func Type(name string, dsl func()) *design.UserTypeDefinition {
 	return t
 }
 
+// ArrayOf can be used in: (it is a top level DSL)
```
raphael: `ArrayOf` can be used wherever a type can be used, e.g. `Attribute("foo", ArrayOf(String))`

design/apidsl/type.go
```diff
@@ -114,6 +119,7 @@ func ArrayOf(v interface{}, dsl ...func()) *design.Array {
 	return &design.Array{ElemType: &at}
 }
 
+// HashOf can be used in: (it is a top level DSL)
```
raphael: Same as `ArrayOf`, it can be used wherever a (DSL) type can.

design/apidsl/action.go
```diff
@@ -126,6 +129,7 @@ func Routing(routes ...*design.RouteDefinition) {
 	}
 }
 
+// GET used in: Routing
```
raphael: Maybe we should say that `Routing` does not take a DSL, `GET` (and the other HTTP method functions) is just an argument to `Routing`.

design/apidsl/attribute.go
deltaskelta: I guess conecptually, to me at the time of writing, nested indicated that it was able to be used within itself, but I see it is confusing to follow. I will just take out the parentheses I guess, unless you have any other ideas?

design/apidsl/media_type.go
deltaskelta: I was kind of following that pattern for the other top level DSL's, API was written like that but I think now maybe it looks better out of parentheses so I will change the others too.

design/apidsl/media_type.go
deltaskelta: for CollectionOf I think it is a good idea to say it can be used where a MediaType can be used, but then where can someone go to see all the possible places that a MediaType identifier can be used? Should we add a list to the MediaType docstring to include all the places a MediaType identifier can be used?

design/apidsl/type.go
deltaskelta: This has the same issue as MediaType in that there is still no place to see all the places that the type identifier can be used.

deltaskelta:
I fixed most of what you said. I think adding the notes to `CollectionOf` `HashOf`... etc that they can be used respectively  wherever MediaType and Type can be used is good, but I think there needs to be some list then in MediaType and Type as to where their identifiers can be used? What do you think?

I also tried to run make locally, but I am getting an error that says `goagen: duplicate DSL goa API` which I am not sure how to interpret

raphael:
This looks great, thank you! I think explaining where types can be used is probably too much for a header comment. Maybe we could add "see https://goa.design/design/types/ for more info" or something like that?

deltaskelta:
ok, I'll try to start working on that in a different PR soon.

raphael:
Ok merging this then, thanks again!

Issue #1202, opened 2017-04-24
BasePath parameters are referenced by a Parent function causes a panic


ikawaha:
Compiling a design whose `BasePath` parameters are referenced by a `Parent` function causes a panic.

This panic seems to occur only in goa 1.2.

## Example design

```go
package design

import (
	. "github.com/goadesign/goa/design"
	. "github.com/goadesign/goa/design/apidsl"
)

var _ = API("panic", func() {
})

var _ = Resource("parent", func() {
	BasePath("/:SomeParam")
	CanonicalActionName("index")

	Action("index", func() {
		Routing(GET("/"))
		Response(OK)
	})
})

var _ = Resource("panic", func() {
	Parent("parent")
	BasePath("panic")

	Action("panic", func() {
		Routing(GET("/"))
		Response(OK)
	})
})
```

**Log**
```
$ goagen bootstrap -d sample/design
exit status 2
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x12fe7cb]

goroutine 1 [running]:
github.com/goadesign/goa/design.(*dupper).DupType(0xc420201510, 0x1648be0, 0xc4201bff20, 0xc420201580, 0xc420201520)
        /Users/ikawaha/go/src/github.com/goadesign/goa/design/dup.go:67 +0x1db
github.com/goadesign/goa/design.Dup(0x1648be0, 0xc4201bff20, 0xc420201870, 0xc420201950)
        /Users/ikawaha/go/src/github.com/goadesign/goa/design/dup.go:7 +0x15d
github.com/goadesign/goa/design.(*ActionDefinition).initQueryParams(0xc420093900)
        /Users/ikawaha/go/src/github.com/goadesign/goa/design/definitions.go:1610 +0x2d1
github.com/goadesign/goa/design.(*ActionDefinition).Finalize(0xc420093900)
        /Users/ikawaha/go/src/github.com/goadesign/goa/design/definitions.go:1480 +0x98
github.com/goadesign/goa/design.(*ResourceDefinition).Finalize.func2(0xc420093900, 0xc4201bfad0, 0x143bc2c)
        /Users/ikawaha/go/src/github.com/goadesign/goa/design/definitions.go:875 +0x3d
github.com/goadesign/goa/design.(*ResourceDefinition).IterateActions(0xc4201c60e0, 0xc420201c90, 0x0, 0x0)
[...]

PR #1215, opened 2017-05-05
Fix potential panic caused by referencing parent path params

ikawaha:
Issue https://github.com/goadesign/goa/issues/1202

I think that this bug seems to be due to fix #1144, but I could not add tests.
It would be appreciated if you could add a test, if necessary.


raphael:
Thank you! I need to convince myself that this is the right approach :) Do you know why the key is missing?

ikawaha:
This fix is incorrect. 

It seems that the cause refers to that action even though the parent's action has not been finalized.

I have no idea... 😱

raphael:
:) OK thanks for looking into it! Sounds like the DSL evaluation order is incorrect, I'll take a look.

PR #1216, opened 2017-05-05
tracing with adaptive sampler

smessier:
@raphael backport adaptive sampling from v2 tracer middleware
maintains but deprecates existing Tracer middleware signature
separated sampling algorithm as Sampler interface for use outside of middleware

### DIFF
--- a/middleware/sampler.go
+++ b/middleware/sampler.go
@@ -23,9 +23,7 @@ type (
 		counter         uint32
 	}
 
-	fixedSampler struct {
-		samplingPercent int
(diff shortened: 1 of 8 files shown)


raphael:
Looks great thank you! I left a comment in the `v2` PR: the only tweak that would be nice to do is to replace the definition of `fixedSampler` so that it's just a `int` instead of a struct with an embeded int. Once that change is made I'll merge all 3 PRs.

raphael:
Thank you!

PR #1218, opened 2017-05-05
tracing with adaptive sampler

smessier:
@raphael change tracing middleware to use Sampler defined separately
added unit tests for tracing middleware

### DIFF
--- a/rest/middleware/tracing/sampler.go
+++ b/rest/middleware/tracing/sampler.go
@@ -23,9 +23,7 @@ type (
 		counter         uint32
 	}
 
-	fixedSampler struct {
-		samplingPercent int
-	}
+	fixedSampler int
 )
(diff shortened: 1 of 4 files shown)


rest/middleware/tracing/sampler.go
```diff
@@ -0,0 +1,102 @@
...
+	adaptiveSampler struct {
+		sync.Mutex
+		lastRate        int64
+		maxSamplingRate int
+		sampleSize      uint32
+		start           time.Time
+		counter         uint32
+	}
+
+	fixedSampler struct {
```
raphael: Couldn't that just be `fixedSampler int`?

smessier: I will make that change.

PR #1232, opened 2017-05-19
Add an argument to CollectionOf()

tchssk:
It makes it possible to specify the collection identifier.

### DIFF
--- a/design/apidsl/media_type.go
+++ b/design/apidsl/media_type.go
@@ -370,22 +370,30 @@ func Link(name string, view ...string) {
 	}
 }
 
-// CollectionOf can be used in: Wherever a MediaType can be used..
+// CollectionOf creates a collection media type from its element media type and an optional
+// identifier. A collection media type represents the content of responses that return a collection
+// of resources such as "list" actions. This function can be called from any place where a media
+// type can be used.
 //
-// e.g. Attribute("foo", CollectionOf(Bar))
-// CollectionOf creates a collection media type from its element media type. A collection media
-// type represents the content of responses that return a collection of resources such as "list"
-// actions. This function can be called from any place where a media type can be used.
+// If an identifier isn't provided then the resulting media type identifier is built from the
+// element media type by appending the media type parameter "type" with value "collection".
 //
-// The resulting media type identifier is built from the element media type by appending the media
-// type parameter "type" with value "collection".
+// Examples:
 //
-//  CollectionOf(BottleMedia)  // If the identifier of BottleMedia is "vnd.goa.bottle',
-//                             // Content-Type will be "vnd.goa.bottle; type=collection".
+//   // Define a collection media type using the default generated identifier
+//   // (e.g. "vnd.goa.bottle; type=collection" assuming the identifier of BottleMedia
+//   // is "vnd.goa.bottle") and the default views (i.e. inherited from the BottleMedia
+//   // views).
+//   var col = CollectionOf(BottleMedia)
 //
-// The collection identifier can be specified as second argument.
-//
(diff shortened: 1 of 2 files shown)


tchssk:
I don't know why travis was failed :(

design/apidsl/media_type.go
```diff
@@ -376,9 +376,17 @@ func Link(name string, view ...string) {
 // CollectionOf creates a collection media type from its element media type. A collection media
 // type represents the content of responses that return a collection of resources such as "list"
 // actions. This function can be called from any place where a media type can be used.
+//
 // The resulting media type identifier is built from the element media type by appending the media
 // type parameter "type" with value "collection".
-func CollectionOf(v interface{}, apidsl ...func()) *design.MediaTypeDefinition {
+//
+//  CollectionOf(BottleMedia)  // If the identifier of BottleMedia is "vnd.goa.bottle',
+//                             // Content-Type will be "vnd.goa.bottle; type=collection".
```
raphael: Should be `identifier` not `Content-Type` I think. By default the identifier is used to build the content type but that can be overridden with `ContentType`. So maybe something like:
```go
// Examples:
//
//   // Define a collection media type using the default generated identifier
//   // (e.g. "vnd.goa.bottle; type=collection" assuming the identifier of BottleMedia
//   // is "vnd.goa.bottle") and the default views (i.e. inherited from the BottleMedia
//   // views).
//   var col = CollectionOf(BottleMedia)
//
//   // Another collection media type using the same element media type but defining a
//   // a different default view.
//   var col2 = CollectionOf(BottleMedia, "vnd.goa.bottle.alternate; type=collection;", func() {
//        View("default", func() {
//           Attribute("id")
//           Attribute("name")
//        })
//   })
```

design/apidsl/media_type.go
```diff
@@ -376,9 +376,17 @@ func Link(name string, view ...string) {
 // CollectionOf creates a collection media type from its element media type. A collection media
```
raphael: Please remove the first 3 lines:
```go
// CollectionOf can be used in: Wherever a MediaType can be used..
 //
 // e.g. Attribute("foo", CollectionOf(Bar))
```
(I know you didn't add them but they don't add much value and where `MediaType` can be used is already explained below). @deltaskelta fyi - we probably should remove them from the other places that say "wherever...".

Also we should probably change the first line to something like:
```go
// CollectionOf creates a collection media type from its element media type and an optional identifier.

File: design/apidsl/media_type.go [hunk 3]
```diff
@@ -376,9 +376,17 @@ func Link(name string, view ...string) {
...
 // type represents the content of responses that return a collection of resources such as "list"
 // actions. This function can be called from any place where a media type can be used.
+//
 // The resulting media type identifier is built from the element media type by appending the media
```
raphael: Replace this with:
```go
// If an identifier isn't provided then the resulting media type identifier ...
```

design/apidsl/media_type.go
```diff
@@ -376,9 +376,17 @@ func Link(name string, view ...string) {
...
 // type parameter "type" with value "collection".
-func CollectionOf(v interface{}, apidsl ...func()) *design.MediaTypeDefinition {
+//
+//  CollectionOf(BottleMedia)  // If the identifier of BottleMedia is "vnd.goa.bottle',
+//                             // Content-Type will be "vnd.goa.bottle; type=collection".
+//
+// The collection identifier can be specified as second argument.
+//
+//  CollectionOf(BottleMedia, "vnd.goa.bottles")  // Content-Type will be "vnd.goa.bottles".
```
raphael: I guess these last 4 lines are not needed with the example above.

tchssk:
I added a fixing commit!

raphael:
Would you mind backporting this to v1? Thank you!

PR #1238, opened 2017-05-25
sort the "produces" section of the swagger.yaml, swagger.json

jim-slattery-rs:
Currently the "produces" section is generated non-deterministically.
For example, sometimes like this:
```yaml
produces:
- application/vnd.goa.error
- application/vnd.my.object+json
```
...sometimes in another order:
```yaml
produces:
- application/vnd.my.object+json
- application/vnd.goa.error
```

This change preserves the order of the elements provided by `action.IterateResponses`.


raphael:
Thank you! could you please backport (cherry-pick) to the `v1` branch as well?

PR #1239, opened 2017-05-25
sort the "produces" section of the swagger.yaml, swagger.json

jim-slattery-rs:
a `v1` companion to: https://github.com/goadesign/goa/pull/1238


raphael:
Thanks!

PR #1245, opened 2017-06-02
fix typo in error level

ehames:


raphael:
Thanks for the PR, this is actually not a typo but a way to make sure that all level strings are 4 characters. It's not the first time that this is mistaken for a typo though so v2 will use `ERROR`.

PR #1246, opened 2017-06-04
examples/code_regen, but built in to goagen

iancmcc:
This PR adds a `--regen` arg to `goagen main` and `goagen controller`, which has the same effect as [the code_regen example](https://github.com/goadesign/examples/blob/master/code_regen), but without an external script or Makefile. It works fine with code generated using existing goagen (that is, uses the same `// Controller_action: start_implement` indicator).

### DIFF
--- a/goagen/gen_controller/generator.go
+++ b/goagen/gen_controller/generator.go
@@ -30,6 +30,7 @@ type Generator struct {
 	DesignPkg string                // Path to design package, only used to mark generated files.
 	AppPkg    string                // Name of generated "app" package
 	Force     bool                  // Whether to override existing files
+	Regen     bool                  // Whether to regenerate scaffolding in place, retaining controller impls
 	Pkg       string                // Name of the generated package
 	Resource  string                // Name of the generated file
 	genfiles  []string              // Generated files
@@ -39,7 +40,7 @@ type Generator struct {
 func Generate() (files []string, err error) {
 	var (
 		outDir, designPkg, appPkg, ver, res, pkg string
-		force                                    bool
+		force, regen                             bool
 	)
 
 	set := flag.NewFlagSet("controller", flag.PanicOnError)
@@ -50,6 +51,7 @@ func Generate() (files []string, err error) {
 	set.StringVar(&res, "res", "", "")
 	set.StringVar(&ver, "version", "", "")
 	set.BoolVar(&force, "force", false, "")
+	set.BoolVar(&regen, "regen", false, "")
 	set.Bool("notest", false, "")
 	set.Parse(os.Args[1:])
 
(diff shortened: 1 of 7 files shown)


raphael:
Thank you for doing this, looks great at first glance! I need to spend a bit more time reviewing it. In the mean time I was wondering if you could add documentation about this new flag to the website, specifically in this file: https://github.com/goadesign/goa.design/blob/master/content/implement/goagen.md? It would be helpful if the doc described a little bit the expected workflow (e.g. first use `bootstrap` then after the design is updated run `regen` to re-generate the controllers. For this to work the comments ... `start` ...  and ... `end` ... need to be kept etc. )

iancmcc:
Creating a PR for that now. In the meantime, there are two areas of this PR that I'm not super happy with, and perhaps you have suggestions.

* Imports - the regenerated files have the default imports, the one part of the implemented controllers that can't (currently) get saved. `goimports` fixes it immediately, of course, but a `goagen controller --regen` followed immediately by a `go build` will likely fail due to missing imports. Not ideal. Could conceivably just always save imports.

* Failures - in one case, specifically where the existing files are read in to extract the controller impls, a failure would (currently) result in the files getting wiped, since an empty map of extracted controllers is returned. Would a `log.Fatal` or equivalent there be appropriate?

Thanks!

raphael:
Thanks for raising the issues.

* Wrt import it should be possible to leverage the Go stdlib `astutil` package to parse the existing imports and "add them back". As an example https://github.com/goadesign/goa/blob/master/goagen/codegen/workspace.go#L250 uses the package to remove imports that are not used.

* Wrt failures maybe I'm missing something but if the generator returns an error instead of writing the empty map then wouldn't that just do what we need?

raphael:
The changes and the PR look great, thank you! It would be nice to be able to preserve the import statements though - do you have a handle on how to do that?

iancmcc:
There! Returning an error was, yeah, simple; somehow I had forgotten that `extractControllerBody` was a function I had literally just introduced and I was therefore wary of changing the signature. Chalk it up to lack of sleep.

Imports were simple once you pointed me there, thanks. Trimming the quotes is a little hacky, but better that than rewriting codegen or double-processing the file with ast.

Still working on the website update.

raphael:
Looks great! Could I coerce you into writing a test for the import preserving logic? I'll merge once that's done, thanks again.

iancmcc:
Tests added for the regeneration, the behavior of regen when the files don't exist yet, and the import preservation. Also, technically, controller generation in the first place, although that may be covered elsewhere, but it made the tests flow better.

raphael:
Looks great, thank you. OK one last thing :) do you think you could backport (cherry-pick) this PR onto the `v1` branch?

PR #1249, opened 2017-06-06
examples/code_regen, but built in to goagen (#1246)

iancmcc:
Add `--regen` goagen flag for `main` and `controller` commands.


raphael:
Thank you!

PR #1251, opened 2017-06-08
There is no typo in EROR

ehames:
Since there have been multiple PRs rejected on this line, I think a comment would make it clearer that the spelling is deliberate.


raphael:
Good idea, thanks!

PR #1254, opened 2017-06-09
xray middleware to periodically redial daemon address

jim-slattery-rs:
xray middleware to periodically redial daemon address because the IP address may change over time.


raphael:
Thank you! Could you please backport to `v1`? thanks!

PR #1265, opened 2017-06-12
xray middleware to periodically redial daemon address

jim-slattery-rs:
A `v1` companion to: https://github.com/goadesign/goa/pull/1254
> xray middleware to periodically redial daemon address because the IP address may change over time.


raphael:
Thank you!

PR #1271, opened 2017-06-19
Fix default package name for gen_controller

tchssk:
It should be `main`. https://github.com/goadesign/goa/pull/1048


tchssk:
Oh I missed #1048...

raphael:
This change makes sense to me, @hectorj2f do you agree?

hectorj2f:
@tchssk @raphael why should it be `main` ? Is it any special scenario why to generate controllers in the main pkg ? I set the default to `controller` cause it organizes the generated code accordingly.

hectorj2f:
@raphael I just saw the comment in the initial issue. I missed to change it there.

PR #1272, opened 2017-06-19
Smart package name for gen_controller

tchssk:
This implements https://github.com/goadesign/goa/pull/1046#issuecomment-279244125.

| --out | --pkg | generated package |
| --- | --- | --- |
| `.` (default) | `main` (default) | `main` |
| `foo` | `main` (default) | `foo` |
| `.` (default) | `bar` | `bar` |
| `foo` | `bar` | `bar` |


raphael:
Looks great, thank you for doing this. Do you think you could remove the first commit "Fix default package name for gen_controller" from the PR? since that commit is already in a different PR. Also it would be nice to have a test that replicates the table you have in the PR comment. Thank you!

tchssk:
Thank you. I'll treat the commits and add tests.

tchssk:
I'v rebased onto master and added tests. Would you take a look? @raphael

goagen/gen_controller/generator.go
```diff
@@ -91,6 +94,7 @@ func (g *Generator) Generate() (_ []string, err error) {
 			err      error
 		)
 		if g.Resource == "" || g.Resource == r.Name {
+			fmt.Println(g.OutDir, g.Pkg)
```
raphael: Is that `fmt.Println` really intended?

goagen/gen_controller/generator.go
```diff
@@ -108,6 +112,30 @@ func (g *Generator) Generate() (_ []string, err error) {
 	return g.genfiles, err
 }
 
+func (g *Generator) smartenPkg() {
+	if g.Pkg != "main" {
+		return
+	}
+	pwd, err := filepath.Abs(".")
```
raphael: Just curious: why not use `os.Getwd`?

goagen/gen_controller/generator_test.go
```diff
@@ -71,6 +72,67 @@ var _ = Describe("Generate", func() {
...
+					content, err := ioutil.ReadFile(filepath.Join(outDir, "foo.go"))
+					Ω(err).ShouldNot(HaveOccurred())
+					Ω(string(content)).Should(HavePrefix("package bar"))
+				})
+			})
+		})
+
+		Context("without --out", func() {
+			BeforeEach(func() {
+				os.Args = []string{"goagen", "--out=" + outDir, "--design=foo", "--version=" + version.String()}
```
raphael: Looks like this is still using `--out`? Maybe the case where `--out` is specified but the value is empty should be another test case.

goagen/gen_controller/generator_test.go
```diff
@@ -71,6 +72,67 @@ var _ = Describe("Generate", func() {
...
+				Ω(files).Should(HaveLen(1))
+				content, err := ioutil.ReadFile(filepath.Join(outDir, "foo.go"))
+				Ω(err).ShouldNot(HaveOccurred())
+				Ω(string(content)).Should(HavePrefix("package main"))
+			})
+
+			Context("with --pkg", func() {
+				BeforeEach(func() {
+					pkg = "bar"
+					os.Args = []string{"goagen", "--out=" + outDir, "--pkg", pkg, "--design=foo", "--version=" + version.String()}
```
raphael: same comment as above

goagen/gen_controller/generator.go
tchssk: Oops just a debug print.

tchssk: https://github.com/goadesign/goa/pull/1272/commits/2e47e580c34568a4ceee567b4e776cbc6802e82d

goagen/gen_controller/generator.go
tchssk: It's more clear. I'v replaced it. https://github.com/goadesign/goa/pull/1272/commits/a5e82032c8e582d704cf80c54fac77b1c318766f

goagen/gen_controller/generator_test.go
tchssk: gen_controller receives `--out` from goagen main every time, right?
https://github.com/tchssk/goa/blob/f54c5877db449a4f81d61bebff88d8139c10d7c6/goagen/main.go#L281-L289

raphael:
Thank you!

tchssk:
Should I backport this to v1? @raphael

raphael:
yes please, thank you!

PR #1277, opened 2017-06-22
Fix default package name for gen_controller

tchssk:
This is a backport of #1271 to v1.


tchssk:
Thank you for reviewing.

raphael:
Thanks!

PR #1280, opened 2017-06-24
add LookupWithResult and ServeLookupResult to Mux interface

### LINKED ISSUE #1279: Lookup should treat url path parameter
I'm using goa and [kami](https://github.com/guregu/kami) in same project (under refactoring).

So, I want to handle request which router can handle.

goa has Lookup method.
https://github.com/goadesign/goa/blob/master/mux.go#L78-L81

but this cannot treat url include path parameter

```
package main

import (
	"log"
	"net/http"
	"net/url"

	"github.com/goadesign/goa"
)

func main() {
	service := goa.New("test")
	service.Mux.Handle("GET", "/v1/bottles", func(http.ResponseWriter, *http.Request, url.Values) {
		log.Println("can lookup")
	})
	service.Mux.Handle("GET", "/v1/bottles/:id", func(http.ResponseWriter, *http.Request, url.Values) {
		log.Println("cannot lookup")
	})
	service.Mux.Lookup("GET", "/v1/bottles")     // can lookup
	service.Mux.Lookup("GET", "/v1/bottles/1")   // cannot lookup
	service.Mux.Lookup("GET", "/v1/bottles/:id") // can lookup
}
```

Now httptreemux support Lookup. (https://github.com/dimfeld/httptreemux/issues/52)

I want goa to use httptreemux's Lookup.

Thanks.

azihsoyn:
resolve #1279 

This include interface change.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -23,8 +23,12 @@ type (
 		// handler registered with Handle. The values argument given to the handler is
 		// always nil.
 		HandleNotFound(handle MuxHandler)
-		// Lookup returns the MuxHandler associated with the given HTTP method and path.
+		// Lookup returns the MuxHandler associated with the given HTTP method and path. (DEPRECATED)
 		Lookup(method, path string) MuxHandler
+		// LookupWithResult call httptreemux Lookup internally.
+		LookupWithResult(w http.ResponseWriter, r *http.Request) (lr httptreemux.LookupResult, found bool)
+		// ServeLookupResult handle request without Lookup.
+		ServeLookupResult(w http.ResponseWriter, r *http.Request, lr httptreemux.LookupResult)
 	}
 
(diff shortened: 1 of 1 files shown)


raphael:
Thanks for the PR! The changes make the `httptreemux` dependency part of the goa package interface which is something that the current design explicitly avoids. The idea is that one can substitute the router with any other router as long as it implements the goa `ServeMux` interface. This change makes that impossible as the interface would now have a direct dependency on `httptreemux`.

To achieve what you want duplicate the mux goa code and configure the mux directly as you want. As long as the mux implements `goa.ServeMux` you can use it with your goa service. Hope this makes sense!

azihsoyn:
Thank you for review!

PR #1282, opened 2017-06-26
Process resource without a parent first

fixl:
Resources without a parent may be parents of other resources. For
deeper levels of nesting this is a problem.

E.g. Z -> Y -> X isn't sorted properly, but X needs to be processed first.


raphael:
Thanks for doing this! Could you please add a test that this fixes?

fixl:
Added. Was a little trickier than I initially thought.

raphael:
Looks great, thank you for doing this!

PR #1284, opened 2017-06-26
Smart package name for gen_controller

tchssk:
This is a backport of #1272 to v1.


raphael:
I believe the build should be fixed with the latest v1 - could you please rebase with the latest? Thank you!

tchssk:
I'v rebased but same error...

raphael:
The errors in the last build seem like genuine errors (test is asserting `package foo` but generated code has `package controller`).

tchssk:
It seems this error is caused because there is no backport of #1048.

raphael:
Good catch! thank you, #1290 fixes the missing backport.

tchssk:
Thank you. Finally the tests has been passed.

raphael:
Great thank you! One last request :) do you think you could update https://github.com/goadesign/goa.design/blob/master/content/implement/goagen.md to describe the behavior?

tchssk:
Thank you for notifying. I'll update it soon.

PR #1289, opened 2017-06-28
Backport to v1: Process resource without a parent first (#1282)

fixl:
* Process resource without a parent first

Resources without a parent may be parents of other resources. For
deeper levels of nesting this can be a problem.

* Iterate over every previous element to ensure no parent was missed


raphael:
Thank you!

PR #1296, opened 2017-07-07
Changeable FileSystem for FileHandler

tchssk:
goa can serve static assets via the DSL `Files` but it always uses the native file system. This patch makes it possible to use another `FileSystem` for that.

### DIFF
--- a/service_test.go
+++ b/service_test.go
@@ -158,7 +158,7 @@ var _ = Describe("Service", func() {
 			}
 		})
 
-		It("creates a handle", func() {
+		It("creates a handler", func() {
 			Ω(muxHandler).ShouldNot(BeNil())
 		})
 
@@ -341,7 +341,7 @@ var _ = Describe("Service", func() {
 			os.RemoveAll(outDir)
 		})
 
-		It("creates a handle", func() {
+		It("creates a handler", func() {
 			Ω(muxHandler).ShouldNot(BeNil())
 		})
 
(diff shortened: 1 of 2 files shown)


raphael:
Thanks! the patch looks good. Do you think you could add a simple test as well?

tchssk:
I added a test.

service_test.go
```diff
@@ -307,6 +309,65 @@ var _ = Describe("Service", func() {
...
+			ctrl := s.NewController("test")
+			handler = ctrl.FileHandler("/swagger.json", "public/swagger/swagger.json")
+			muxHandler = ctrl.MuxHandler("testAct", handler, nil)
+		})
+
+		AfterEach(func() {
+			os.RemoveAll(outDir)
+		})
+
+		It("creates a handle", func() {
```
raphael: handle -> handler

raphael:
Thank you the test looks great! (spotted a typo - left a comment). I guess it doesn't test overriding `FileSystem` but it's OK since the test proves that the default implementation is correctly invoked.

tchssk:
https://github.com/goadesign/goa/pull/1296/commits/4d8adde3da050dd6c01d5bfbc255c6e301e105e0 fixes typos.

raphael:
Thank you!

raphael:
I guess this could be backported to `v1` as well

tchssk:
Thank you! I'll backport this to `v1`.

PR #1303, opened 2017-07-17
Made an error message in the test generator slightly more useful.

lsh246:
I found this was helpful when I was writing some testcases.


raphael:
Makes sense, thank you! Do you think you could backport that change to the `v1` branch as well? This means branching off the `v1` branch, cherry picking the commit and sending a different PR. Thank you!

lsh246:
Done - I think. Made a PR in the `v1` branch from my fork.

PR #1307, opened 2017-07-18
Elide type in composite literals

tchssk:
This is a cosmetic change.


raphael:
Thanks! the `encoding_redux` branch is the latest and already has that change implemented so I'm just going to close this.

PR #1309, opened 2017-07-20
add var for template at work with 'use struct'

lyykfi:


raphael:
Thanks for the fix! do you think you could backport it to the `v1` branch? simply checkout the branch, cherry-pick the commit, push and create a new PR.

PR #1310, opened 2017-07-20
Fix bootstrap pkg, tooldir and regen flag

### LINKED ISSUE #1308: Panic when passing in --pkg, --tooldir, --regen to `goagen bootstrap` in Windows command line
Tested this with all three flags individually. Not tested on Mac. Here's sample output from trying --regen with the adder example. Got the similar output, with slightly different call stacks, with the other two flags (pkg and tooldir).

```
...\go\src\github.com\goadesign\examples\adder>goagen bootstrap --design github.com/goadesign/examples/adder/design --out ./ --regen
exit status 2
flag provided but not defined: -regen
Usage of app:
  -design string

  -force

  -notest

  -out string

  -pkg string
         (default "app")
  -version string

panic: flag provided but not defined: -regen

goroutine 1 [running]:
flag.(*FlagSet).Parse(0xc04218e540, 0xc042058060, 0x4, 0x4, 0x0, 0x0)
        G:/Go/src/flag/flag.go:926 +0xb9
github.com/goadesign/goa/goagen/gen_app.Generate(0x0, 0x0, 0xa4a180, 0xc04218b22
0, 0x0)
        C:/Users/Lawrence/go/src/github.com/goadesign/goa/goagen/gen_app/generat
or.go:51 +0x307
main.main()
        C:/Users/Lawrence/go/src/github.com/goadesign/examples/adder/goagen33171
8743/main.go:29 +0x7d
```

lsh246:
Some generators just ignore the flag but they must still define it so that bootstrap can use it.

Fixes #1308

### DIFF
--- a/goagen/gen_app/generator.go
+++ b/goagen/gen_app/generator.go
@@ -37,8 +37,8 @@ type Generator struct {
 // Generate is the generator entry point called by the meta generator.
 func Generate() (files []string, err error) {
 	var (
-		outDir, target, ver string
-		notest              bool
+		outDir, toolDir, target, ver string
+		notest, regen                bool
 	)
 
 	set := flag.NewFlagSet("app", flag.PanicOnError)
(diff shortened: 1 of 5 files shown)


raphael:
This is great, thank you!

PR #1320, opened 2017-07-24
Add appengine context middleware

Oralordos:
Adds a middleware that makes sure that the goa contexts are also App Engine contexts. If the user is not on App Engine, the middleware will do nothing.

### DIFF
--- a/middleware/with_appengine_context.go
+++ b/middleware/with_appengine_context.go
@@ -1,0 +1,19 @@
+// +build appengine
+
+package middleware
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/goadesign/goa"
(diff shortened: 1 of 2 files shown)


raphael:
Thank you for doing this. I've been trying to think through whether this is something that makes sense in a library. At the end of the day it is very little code and pretty much all boilerplate. It seems to make more sense to me as an example "here you want to use AppEngine do this" rather than a reusable package. Do you think you could repackage this as an example? there's already a basic app engine example in https://github.com/goadesign/examples/tree/master/appengine - seems like adding this middleware there and highlighting it in the README would work well. Does that make sense to you?

raphael:
closing this ticket as it should be moved to the examples repo.

Issue #1173, opened 2017-04-09
Multiple errors response


tpoxa:
Hello. For server side validation my client needs to have response like this.

HTTP CODE 422

JSON
```
[
{"field":"name", "error":"Field is required"},
{"field":"email", "error":"Field is required"},
]
```
How to implement this with goa.design?

Thank you

raphael:
You can replace the `ErrInvalidRequest` with your own error class that produces the format you want. However when multiple errors are produced `MergeErrors` produces a `500` in this case because it doesn't know about any other error type than `ErrorResponse` which is probably not what you want. This is a common request so I would gladly accept a PR that tweaks `MergeErrors` so it works with custom error types as well.

PR #1321, opened 2017-07-27
Add ability to customize errors response structure.

brycereitano:
This provides the interface for merging custom error response structs.

Tests and documentation to come after review of approach.

### DIFF
--- a/error.go
+++ b/error.go
@@ -95,6 +95,16 @@ type (
 		Token() string
 	}
 
+	// ServiceMergeableError is the interface implemented by ServiceErrors that can merge
+	// another error into a combined error.
+	ServiceMergeableError interface {
+		// ServiceMergeableError extends from the ServiceError interface.
+		ServiceError
+
+		// Merge updates an error by combining another error into it.
+		Merge(other error) error
+	}
+
 	// ErrorResponse contains the details of a error response. It implements ServiceError.
 	// This struct is mainly intended for clients to decode error responses.
 	ErrorResponse struct {
@@ -254,6 +264,8 @@ func (e *ErrorResponse) Token() string { return e.ID }
 // ServiceError if not already one - producing an internal error in that case. The merge algorithm
 // is:
 //
+// * If any of e or other implements ServiceMergableError, it is handled by its Merge method.
+//
 // * If any of e or other is an internal error then the result is an internal error
 //
 // * If the status or code of e and other don't match then the result is a 400 "bad_request"
@@ -269,11 +281,20 @@ func MergeErrors(err, other error) error {
 		if other == nil {
 			return nil
 		}
-		return asErrorResponse(other)
+		return asServiceError(other)
 	}
 	if other == nil {
-		return asErrorResponse(err)
+		return asServiceError(err)
 	}
+
+	// If either error is a mergable error.
+	if me, ok := err.(ServiceMergeableError); ok {
+		return me.Merge(other)
+	}
+	if mo, ok := other.(ServiceMergeableError); ok {
+		return mo.Merge(err)
+	}
+
 	e := asErrorResponse(err)
 	o := asErrorResponse(other)
 	switch {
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -265,6 +299,14 @@ func (e *ErrorResponse) Token() string { return e.ID }
 // Merge returns the updated error. This is useful in case the error was initially nil in
 // which case other is returned.
 func MergeErrors(err, other error) error {
+	// If either err is mergable, merge them.
```
raphael: This should probably be moved after the `nil` tests below

error.go
```diff
@@ -250,6 +260,30 @@ func (e *ErrorResponse) ResponseStatus() int { return e.Status }
...
+	if e.Meta == nil && len(o.Meta) > 0 {
+		e.Meta = make(map[string]interface{})
+	}
+	for k, v := range o.Meta {
+		e.Meta[k] = v
+	}
+	return e
+}
+
 // MergeErrors updates an error by merging another into it. It first converts other into a
```
raphael: This comment should be moved to the `ErrorResponse.Merge` method since the logic may now be different. The new comment should explain that the algo first checks whether err is mergeable and is so merges o into it then whether o is mergeable and finally creates a mergeable error with status code 500 if none is.

error.go
```diff
@@ -95,6 +95,16 @@ type (
 		Token() string
 	}
 
+	// ServiceMergeableError is the interface implemented by ServiceErrors that can merge
+	// another error into a combined error.
+	ServiceMergeableError interface {
+		// ServiceMergeableError extends from the ServiceError interface.
+		ServiceError
+
+		// Merge will handle the logic of merging two errors together.
```
raphael: will handle -> handles

error.go
brycereitano: Fair enough, the applications customized error handling should know how to handle a `goa.ErrorResponse` either way.

brycereitano: Oh actually, there is an issue with that. That may turn a custom mergeable error into a `*goa.ErrorResponse`, stringifying the result into `Detail`.

brycereitano: Maybe it would be better if `ErrorResponse` didn't implement ServiceMergableError. Removes the possibility of the old merging logic to be used.

raphael: that makes sense to me, good catch!

brycereitano:
@raphael Updated so that `ErrorResponse` doesn't implement `ServiceMergableError`. As it can cause the old merging logic to be called if `err` happens to a `goa.ErrorResponse`.

error.go
```diff
@@ -254,6 +264,8 @@ func (e *ErrorResponse) Token() string { return e.ID }
 // ServiceError if not already one - producing an internal error in that case. The merge algorithm
 // is:
 //
+// * If any of e or other implements ServiceMergableError, it is handled by it's Merge method.
```
seubert: *its

brycereitano: Nice catch, I'll get it updated.

brycereitano:
Once this is merged, this should close  #1173.

error_test.go
```diff
@@ -300,24 +310,22 @@ var _ = Describe("Merge", func() {
 		})
 	})
 
-	Context("with a nil argument", func() {
-		const code = "foo"
+	Context("with a nil target", func() {
```
raphael: nitpick: the test is calling `MergeErrors` so really it's still an argument and not a target?

error_test.go
```diff
@@ -326,178 +334,226 @@ var _ = Describe("Merge", func() {
...
-							Ω(mErr.Meta[k]).Should(Equal(v))
-						}
-					}
-					Ω(mErr.Meta[commonKey]).Should(Equal(metaValues2[commonKey]))
+				It("calls user defined merge", func() {
+					Ω(mErr.(*MergeableErrorResponse).MergeCalled).Should(Equal(1))
+				})
+			})
+
+			Context("with MergableError", func() {
```
raphael: would be good to also have a test where the first error is NOT a mergeable error but the second one is that validates the second error `Merge` method is called.

error_test.go
brycereitano: The other tests used a similar terminology. However, I think `target` is fine, as it's the `error` that is modified.

raphael:
Looks great - thank you! Do you think you could also send a PR against `master`? thanks!

PR #1325, opened 2017-08-03
Fix ResponseData.Write behavior to match net/http.ResponseWriter.Write.

sotosof:
From documentation of net/http.ResponseWriter.Write

    If WriteHeader has not yet been called, Write calls
    WriteHeader(http.StatusOK) before writing the data.

Without this fix, ResponseData which went through middleware which depends
on documented behavior of net/http.ResponseWriter.Write, basically which doesn't
call net/http.ResponseWriter.WriteHeader explicitly, ended in non Written
state. This leads into situation that NotFound handler will write into
ResponseData although middleware did the write already.

### DIFF
--- a/context.go
+++ b/context.go
@@ -45,6 +45,8 @@ type (
 		Status int
 		// Length is the response body length.
 		Length int
+		// Flag if WriteHeader has been called.
+		wroteHeader bool
 	}
 
 	// key is the type used to store internal values in the context.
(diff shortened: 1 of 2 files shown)


raphael:
Thank you! is the `writeHeader` field really necessary? couldn't the PR just be:
```go
 // Write records the amount of data written and calls the underlying writer.
  func (r *ResponseData) Write(b []byte) (int, error) {
 +	if !r.Written() {
 +		r.WriteHeader(http.StatusOK)
 +	}
        r.Length += len(b)
```

sotosof:
Yep, but I just didn't want to depend on special value of `Status` which can be set by caller (`0`).

raphael:
A few things:

1. `0` is not a valid HTTP status and the code is already assuming that it won't be set to that. 
2. Fixing this bug is an orthogonal issue and does not need to change that, it could be a different PR is you feel strongly about it (would need a real use case though).
3. A bit moot given 1 and 2 but the only place the code should test `wroteHeader` is `Written`. The code in `Write` should use `Written` and not `wroteHeader` directly. Again I don't think `wroteHeader` is necessary in the first place.

raphael:
Thanks you! Do you think you could cherry pick this fix against the v1 branch as well?

PR #1326, opened 2017-08-03
v1: Fix ResponseData.Write behavior to match net/http.ResponseWriter.Write.

sotosof:
Same as #1325 but into v1 branch.


raphael:
Thank you!

PR #1329, opened 2017-08-04
Only open generated files once.

raphael:
This is required for performance reasons especially on Windows. Unfortunately
this also means that files must now be explicitly closed.


tchssk:
This pull request contains some breaking changes (e.g. https://github.com/goadesign/goa/blob/c486c800bad5e75367d9f9bb4ffe3f9a109adffd/goagen/codegen/workspace.go#L263). Is it OK? @raphael

raphael:
Good catch - there was no need to break that. Let me push a follow up PR to maintain backwards compatibility, thank you for bringing that up.

PR #1350, opened 2017-08-09
DMO-1294 xray: request-level segment shouldn't be marked "subsegment"

jim-slattery-rs:
This was causing an X-Ray trace through multiple services to show up as a tree of subsegments under a single service. So queries like `service("xyz")` wouldn't show anything, even though the trace did flow through that service.


raphael:
Thank you!

PR #1351, opened 2017-08-09
DMO-1294 xray: request-level segment shouldn't be marked "subsegment"

jim-slattery-rs:
Companion PR for master; see: https://github.com/goadesign/goa/pull/1350


raphael:
Thanks!

PR #1355, opened 2017-08-15
DMO-1294 introduce `xray.WrapDoer`, replacing `xray.WrapClient`

jim-slattery-rs:
`WrapClient` was appropriate only for one-time use with a single Context.
`WrapDoer` can be setup one time and then reused across any number of Contexts.


raphael:
Thank you!

PR #1358, opened 2017-08-20
doc: Fix JWT validation handler example

klauspost:
Fix the example in the documentation by asserting the claims type and adding a missing return nil at the end.


raphael:
Great, thank you!

raphael:
Could you please back-port (cherry-pick) this fix to the `v1` branch as well?  Thank you!

PR #159, opened 2016-01-07
Add Gzip Middleware

tylerb:
This PR adds a Gzip middleware for easily gzipping your responses. This middleware takes advantage of `sync.Pool` and `gzip.Reset()` to re-use buffers to significantly reduce allocations and GC.

Please let me know if additional work needs to be done before merge.

PR #1368, opened 2017-08-29
Improve gzip middleware

klauspost:
Change gzip middleware to only be employed if:

* Response is above specified size (Default: 256 bytes)
* Status code is one of. (Default: OK, Created, Accepted)
* Content type has one of X prefixes. (Default: Common types+ all `application/vnd.`)

Options added are `AddContentTypes`, `OnlyContentTypes`, `AddStatusCodes`, `OnlyStatusCodes` and `MinSize`.

Expanded tests to cover these.

In writing this I came across [this check](https://github.com/goadesign/goa/blob/master/middleware/gzip/middleware.go#L68) which I cannot find the reason behind. So if `"Content-Encoding"` header is set to `"gzip"` on the *request*, we will not use gzip? I have left it, but it seems like an odd check to have.

### DIFF
--- a/middleware/gzip/middleware.go
+++ b/middleware/gzip/middleware.go
@@ -98,7 +98,7 @@ func (grw *gzipResponseWriter) WriteHeader(n int) {
 }
 
 type (
-	// Option allows to override processing parameters.
+	// Option allows to override default parameters.
 	Option func(*options) error
 
 	// options contains final options
--- a/middleware/gzip/middleware.go
+++ b/middleware/gzip/middleware.go
@@ -98,8 +98,8 @@ func (grw *gzipResponseWriter) WriteHeader(n int) {
 }
 
 type (
-	// option allows to override processing parameters.
-	option func(*options) error
+	// Option allows to override processing parameters.
+	Option func(*options) error
 
 	// options contains final options
 	options struct {
@@ -169,7 +169,7 @@ var defaultStatusCodes = []int{
 
 // AddContentTypes allows to specify specific content types to encode.
 // Adds to previous content types.
-func AddContentTypes(types ...string) option {
+func AddContentTypes(types ...string) Option {
 	return func(c *options) error {
 		dst := make([]string, len(c.contentTypes)+len(types))
 		copy(dst, c.contentTypes)
@@ -182,7 +182,7 @@ func AddContentTypes(types ...string) option {
 // OnlyContentTypes allows to specify specific content types to encode.
 // Overrides previous content types.
 // no types = ignore content types (always compress).
-func OnlyContentTypes(types ...string) option {
+func OnlyContentTypes(types ...string) Option {
 	return func(c *options) error {
 		if len(types) == 0 {
 			c.contentTypes = nil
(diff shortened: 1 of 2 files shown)


klauspost:
> `middleware/gzip/middleware.go:172:39: exported func AddContentTypes returns unexported type gzip.option, which can be annoying to use`

A linter can also be annoying to use ;)

raphael:
Looks great thank you! Reviewed the code and all the review items ended up being all very minor stuff amounting more to taste than to constructive feedback. Do you think you could backport (cherry-pick) to `v1`? thank you!

raphael:
Turns out one of the review items was to not use a global variable to store the default option and that turns out to be more than taste. The code could modify the default options which could have undesirable side effects (one of them being random [test failures](https://travis-ci.org/goadesign/goa/builds/269752977)). https://github.com/goadesign/goa/pull/1369 fixes that. Going to merge it to fix the build but feel free to comment/modify.

raphael:
Ah forgot to address the question around the dubious check. I don't understand it either - seems like we should remove it. @tylerb long shot I know [the PR](https://github.com/goadesign/goa/pull/159) is more than one year old :) but maybe you remember why [this check](https://github.com/goadesign/goa/blob/master/middleware/gzip/middleware.go#L261) might be needed?

tylerb:
I believe it is checking to see if the data has already been encoded as gzip. Probably more defensive than anything.

raphael:
woah thanks for the quick repsonse :) 
Wait a minute... the [inital PR](https://github.com/goadesign/goa/pull/159/files#diff-4c166b743d95629a9ab37eb6fe3df22aR339) checked the *response* header (which is probably a good idea) - the current code checks the *request* header (which is probably wrong).

tylerb:
Sure thing. 

Yes it did. Not sure how that got changed. Odd.

raphael:
Fixed in 1370

faguirre1:
i'm not getting these headers in the response:

	grw.Header().Set(headerContentEncoding, encodingGzip)
	grw.Header().Set(headerVary, headerAcceptEncoding)

is this a bug ?

faguirre1:
this is an example response headers i got using curl.

> Host: graphs.rh-sma.dev:9999
> Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
> Connection: keep-alive
> Accept-Encoding: gzip, deflate
> Accept-Language: en-US,es-AR;q=0.8,es;q=0.6,en;q=0.4
> Upgrade-Insecure-Requests: 1
> User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36
> 
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Length: 3527
< Content-Type: text/html; charset=utf-8
< Last-Modified: Sat, 24 Jun 2017 15:03:53 GMT
< Date: Sat, 16 Sep 2017 15:41:41 GMT
< 
{ [1528 bytes data]
* transfer closed with 1999 bytes remaining to read
* stopped the pause stream!

PR #1374, opened 2017-08-31
Added Method Not Allowed default handler

### LINKED ISSUE #1372: Using wrong HTTP method on endpoint returns 404 - 405 expected
If i request with the wrong method, which is not defined in the api design, I get a 404 - Not Found error.

**POST:**
```
http post localhost:8080/api/login username=alice password=secure
HTTP/1.1 401 Unauthorized
Content-Length: 72
Content-Type: application/vnd.goa.error
Date: Wed, 30 Aug 2017 11:48:37 GMT
```
**PUT:**
```
http put localhost:8080/api/login username=alice password=secure
HTTP/1.1 404 Not Found
Content-Length: 72
Content-Type: application/vnd.goa.error
Date: Wed, 30 Aug 2017 11:48:37 GMT
```
I expected to get something like
```
http put localhost:8080/api/login username=alice password=secure
HTTP/1.1 405 Method Not Allowed
Content-Length: 72
Content-Type: application/vnd.goa.error
Allow: POST
Date: Wed, 30 Aug 2017 11:48:37 GMT
```
It seems to be a very confusing behaviour, since the endpoint exists.

CuBiC3D:
This PR changes the the way how goa handles not existent HTTP methods on existing endpoints and resolve #1372.

The new public `HandleMethodNotAllowed` is able to take a self defined handler. Therefore it was necessary to introduce a new `MethodNotAllowedHandler` which extends the the `MuxHandler` arguments by `map[string]httptreemux.HandlerFunc`. This allows to retrieve the allowed methods and handler by the MethodNotAllowed handler.

A default handler is included, which sets the appropriate status code, `Allow` header and a new error class.

This changes can be breaking in some edge cases, but introduce the correct behaviour described in https://tools.ietf.org/html/rfc7231#section-6.5.5

### DIFF
--- a/error.go
+++ b/error.go
@@ -69,8 +69,8 @@ var (
 	// ErrNotFound is the error returned to requests that don't match a registered handler.
 	ErrNotFound = NewErrorClass("not_found", 404)
 
-	// ErrMethodNotAllowed is the error returned to requests that have a registered handler
-	// but the used HTTP method does not match.
+	// ErrMethodNotAllowed is the error returned to requests that match the path of a registered
+	// handler but not the HTTP method.
 	ErrMethodNotAllowed = NewErrorClass("method_not_allowed", 405)
 
 	// ErrInternal is the class of error used for uncaught errors.
@@ -249,8 +249,8 @@ func NoAuthMiddleware(schemeName string) error {
 	return ErrNoAuthMiddleware(msg, "scheme", schemeName)
 }
 
-// MethodNotAllowedError is the error produced to requests that have a registered handler
-// but the used HTTP method does not match.
+// MethodNotAllowedError is the error produced to requests that match the path of a registered
+// handler but not the HTTP method.
 func MethodNotAllowedError(method string, allowed []string) error {
 	msg := fmt.Sprintf("Method %s must be one of %s", method, strings.Join(allowed, ", "))
 	return ErrMethodNotAllowed(msg, "method", method, "allowed", strings.Join(allowed, ", "))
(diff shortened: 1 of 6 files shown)


error.go
```diff
@@ -69,6 +69,10 @@ var (
 	// ErrNotFound is the error returned to requests that don't match a registered handler.
 	ErrNotFound = NewErrorClass("not_found", 404)
 
+	// ErrMethodNotAllowed is the error returned to requests that have a registered handler
+	// but the used HTTP method does not match.
```
raphael: This might be a bit clearer:
```go
// ErrMethodNotAllowed is the error returned to requests that match the path of a registered
// handler but not the HTTP method.
```

error.go
```diff
@@ -245,6 +249,13 @@ func NoAuthMiddleware(schemeName string) error {
 	return ErrNoAuthMiddleware(msg, "scheme", schemeName)
 }
 
+// MethodNotAllowedError is the error produced to requests that have a registered handler
+// but the used HTTP method does not match.
```
raphael: Same as above

error.go
```diff
@@ -245,6 +249,13 @@ func NoAuthMiddleware(schemeName string) error {
...
+func MethodNotAllowedError(method string, allowed []string) error {
+	msg := fmt.Sprintf("Method %s must be one of %s", method, strings.Join(allowed, ", "))
```
raphael: maybe make a special case for when there's only one method? otherwise the message might be confusing (`Method GET must be one of POST`).

CuBiC3D:
Thanks, I'm glad to help. Nitpicks are hopefully fixed now :innocent:

raphael:
Yes perfect, thanks again. One last thing :) do you think you could backport this to the `v1` branch. Checkout v1 create a branch, cherry-pick this merge commit and create a PR. Thanks again.

CuBiC3D:
@raphael Sure, here you go #1375 ;)

k3a:
I am confused, was it merged? It doesn't seem to be part of v3@v3.4.2 http module

PR #1376, opened 2017-09-07
Expose Sign errors.

arangamani:


raphael:
Thank you! Do you think you could backport the fix to `v1` as well?

arangamani:
Done in https://github.com/goadesign/goa/pull/1377

PR #1382, opened 2017-09-13
Fixed ws client template

smousa:


raphael:
Thank you! Do you think you could backport this to `v1`? (create a branch off of `v1`, cherry-pick this commit then send a new PR).

PR #1384, opened 2017-09-13
Fixed ws client template (#1382)

smousa:
* add missing linebreak to websocket client template

* updated ws client template test


raphael:
Thank you!

Issue #1228, opened 2017-05-15
Null Scopes From Bearer Token


fluffybonkers:
I am having an issue with the bearer token submitted to my controllers returning "authorization failed" because of "scopes:null" using Postman. I am using Auth0 for authentication. The exact error in Postman is:

`{"id":"xOUR882s","code":"jwt_security_error","status":401,"detail":"authorization failed: required 'scopes' not present in JWT claim","meta":{"required":["read:meta"],"scopes":null}}`

However, my token does include the required scope `read:meta`. [jwt.io](url) gives back the following decoded bearer:

```
{
  "iss": "https://learnlogic.au.auth0.com/",
  "sub": "exJMkK7hXX56lrLwoTqna3s0jh7Gq67e@clients",
  "aud": "https://api.learn-logic.com",
  "exp": 1494855336,
  "iat": 1494768936,
  "scopes": "read:meta"
}
```
I am hoping someone can help me workout why, as I don't really understand what is going on in `middleware/jwt.go`.

I have the following `main.go` code:

```
b, err := ioutil.ReadFile("util/jwt.key")
	if err != nil {
		return
	}

	block, _ := pem.Decode([]byte(b))
	var cert *x509.Certificate
	cert, _ = x509.ParseCertificate(block.Bytes)
	rsaPublicKey := cert.PublicKey.(*rsa.PublicKey)
	fmt.Println(rsaPublicKey.N)
	fmt.Println(rsaPublicKey.E)
	fmt.Println(cert)

	var keyx = []jwt.Key{rsaPublicKey}

	var jwtResolver = jwt.NewSimpleResolver(keyx)

	app.UseJWTMiddleware(service, jwt.New(jwtResolver, nil, app.NewJWTSecurity()))
```

The certificate I am reading in is the same one used in [jwt.io](url) to decode the bearer token.

Any help is much appreciated.

raphael:
If you log the content of `claims` after the line below what do you get?

https://github.com/goadesign/goa/blob/master/middleware/security/jwt/jwt.go#L163

Maybe @xeger has a better idea :)

xeger:
[...]

PR #1399, opened 2017-09-21
Allow JWT scopes claim to be named 'scope' or 'scopes'

### LINKED ISSUE #1398: JWT security apparently incompatible with Auth0
Auth0 generates JWTs that include scopes using the claim `scope`, and, as far as I can tell, there's no way to control that. Goa is hardcoded to look for the claim `scopes`, and there's no way to control that, either, short of patching it. 

Is there any solution to this impasse? If not, would a PR that, say, adds `ScopesClaim` as a DSL func be welcome?

iancmcc:
This fixes #1398 and #1228.

### DIFF
--- a/middleware/security/jwt/jwt.go
+++ b/middleware/security/jwt/jwt.go
@@ -21,8 +21,9 @@ import (
 //
 //     1. Validate the "Bearer" token present in the "Authorization" header against the key(s)
 //        given to New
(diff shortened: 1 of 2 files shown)


raphael:
Looks great! Would you mind porting this to `v1` as well? (cherry-pick the merge commit). Thank you!

PR #1400, opened 2017-09-22
Sync MuxHandler init

klauspost:
If multiple calls are made at once several will try to initialize and are racing to do so.


raphael:
Good one! thanks :-) Would you care to backport that to `v1`?

PR #1403, opened 2017-09-29
Add ErrForbidden error variable

navruzm:


raphael:
Thanks for the PR! The `ErrXXX` variables are meant to only defines status code needed by goa itself. The intent is *not* to list all the possible status codes. Error classes only needed by user code should be defined there.

PR #1404, opened 2017-09-30
Sync MuxHandler init (#1400 -> v1)

klauspost:
If multiple calls are made at once several will try to initialize and are racing to do so.


raphael:
Thank you!

PR #1405, opened 2017-10-03
Fix Correspondence of packages to be should excluded on appengine

pei0804:
Syscall package does't permission on appengine.So, I fix it.

```
ERROR    2017-10-03 08:43:17,275 go_runtime.py:194] Failed to build Go application: (Executed command: /Users/jumpei/go_appengine/goroot-1.8/bin/go-app-builder -app_base /Users/jumpei/go/src/github.com/pei0804/hoge/server -arch 6 -dynamic -goroot /Users/jumpei/go_appengine/goroot-1.8 -gopath /Users/jumpei/go -nobuild_files ^^$ -incremental_rebuild -unsafe -print_extras_hash main.go)

2017/10/03 17:43:17 go-app-builder: Failed parsing input: parser: bad import "syscall" in github.com/goadesign/goa/goagen/utils/signal.go from GOPATH
```

[The App Engine SDK and workspaces (GOPATH) - The Go Blog](https://blog.golang.org/the-app-engine-sdk-and-workspaces-gopath)

PR #1408, opened 2017-10-04
Fix Correspondence of packages to be should excluded on appengine

pei0804:
cherry-pick v1 https://github.com/goadesign/goa/pull/1405


raphael:
Thank you!

PR #1420, opened 2017-11-09
Append a newline to the errors from goa

tchssk:
Now `goa` prints an error without a newline. The next prompt is printed to the same line like below.

```console
$ goa gen adder
no buildable Go source files in adder  $ 
```

This patch append a newline to the errors from `goa`.

```console
$ goa gen adder
no buildable Go source files in adder
$ 
```


raphael:
Great, thank you!

PR #1460, opened 2017-11-29
disallow_result_for_no_content

### LINKED ISSUE #1452: goa v2: http: multiple response.WriteHeader calls
Consider the following design:
```
package design

import (
	. "goa.design/goa/http/dsl"
)

var _ = API("agent", func() {
	Title("agent")
	Version("0.1")
	Server("http://localhost:8080")
})

var _ = Service("agent", func() {
	HTTP(func() {
		Path("/agent")
	})

	Method("show", func() {
		Payload(func() {
			Attribute("interface", String)
			Required("interface")
		})
		Result(String)
		HTTP(func() {
			GET("/{interface}")
			Response(StatusNoContent)
		})
	})
})
```
When calling the `show` endpoint you will receive an error in the server console:
`2017/11/20 13:27:29 http: multiple response.WriteHeader calls`

This is probably due to an issue with the `Response` saying there is no content (status 204), but the `Result` still allows for a result to be written.

nitinmohan87:
Fixes #1452

### DIFF
--- a/http/design/response.go
+++ b/http/design/response.go
@@ -134,6 +134,10 @@ func (r *HTTPResponseExpr) Validate() *eval.ValidationErrors {
 	}
 	if r.StatusCode == 0 {
 		verr.Add(r, "HTTP response status not defined")
+	} else {
+		if !bodyAllowedForStatus(r.StatusCode) && r.bodyExists() {
+			verr.Add(r, "Response body defined for status code %d which does not allow response body.", r.StatusCode)
+		}
 	}
 	return verr
 }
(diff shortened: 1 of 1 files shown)


http/design/response.go
```diff
@@ -134,6 +134,10 @@ func (r *HTTPResponseExpr) Validate() *eval.ValidationErrors {
 	}
 	if r.StatusCode == 0 {
 		verr.Add(r, "HTTP response status not defined")
+	} else {
```
raphael: You can write this like this:
```go
        verr.Add(r, "HTTP response status not defined")
    } else if !bodyAllowedForStatus(r.StatusCode) && r.bodyExists() {
        verr.Add(r, "Response body defined for status code %d which does not allow response body.", r.StatusCode)
    }
```

http/design/response.go
```diff
@@ -210,3 +214,31 @@ func (r *HTTPResponseExpr) Dup() *HTTPResponseExpr {
...
+}
+
+// bodyExists returns true if a response body is defined in the
+// response expression via Body() or Result() in the method expression.
+func (r *HTTPResponseExpr) bodyExists() bool {
+	if r.Body != nil && r.Body.Type == design.Empty {
+		return false
+	}
+	ep, ok := r.Parent.(*EndpointExpr)
+	if ok && ep.MethodExpr.Result != nil {
```
raphael: you can write this like this:
```go
return ok && ep.MethodExpr.Result != nil
```

http/design/response.go
```diff
@@ -210,3 +214,31 @@ func (r *HTTPResponseExpr) Dup() *HTTPResponseExpr {
...
+	case status == 204:
+		return false
+	case status == 304:
+		return false
+	}
+	return true
+}
+
+// bodyExists returns true if a response body is defined in the
+// response expression via Body() or Result() in the method expression.
```
raphael: It seems the function returns false if the response has a body but is defined at the service or API level. If this is OK it should probably be documented in the comment?

nitinmohan87: HTTP expression at the API or Service level can only contain error response as per the documentation. And all HTTP error codes can contain a response body. So I think we can have this function just check the method expression for now.

Issue #1462, opened 2017-11-30
goagen: How about move `end_implement` mark to the end of function definition?


tanatana:
Currently, `end_implement` mark is placed before response creation and return statement. In many cases, I delete them and rewrite them newly before the mark.

So when  we use `goagen main --regen`, response creation and return statement are re-generated and break the source code.

I feel it is better that put `end_miplement` mark at the end of function definition.

raphael:
If I understand correctly you would like the response creation and return statement to be generated once in the block but then not generated if the block already exists? That makes sense, would be happy to merge a PR that did that.

raphael:
I believe this can be closed?

tugusuke:
Yes, sorry I forgot to inform you. Thank you!

tanatana:
Yes, sorry I forgot to inform you.

raphael:
Thank you for the great contribution!

PR #1464, opened 2017-12-03
Return statement generated once v1

tanatana:
#1462

### DIFF
--- a/goagen/gen_main/generator.go
+++ b/goagen/gen_main/generator.go
@@ -375,9 +375,6 @@ func funcMap(appPkg string, actionImpls map[string]string) template.FuncMap {
 		"okResp":    okResp,
 		"targetPkg": func() string { return appPkg },
 		"actionBody": func(name string) string {
-			if actionImpls == nil {
-				return defaultActionBody
-			}
 			body, ok := actionImpls[name]
 			if !ok {
 				return defaultActionBody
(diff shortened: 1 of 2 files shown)


goagen/gen_main/generator.go
```diff
@@ -384,6 +384,16 @@ func funcMap(appPkg string, actionImpls map[string]string) template.FuncMap {
 			}
 			return body
 		},
+		"printResp": func(name string) bool {
+			if actionImpls == nil {
```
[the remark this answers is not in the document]
tanatana: Thank you, I am going to fix it.
By the way, the same processing is on line 378. Is it better to modify it the same way?
https://github.com/goadesign/goa/pull/1464/files/a828f7fe2be77a77c26165802a16662d5fde966d#diff-34b00d9b7f5787450c4e36c8cedbd6b1R378

goagen/gen_main/generator.go
```diff
@@ -375,19 +375,13 @@ func funcMap(appPkg string, actionImpls map[string]string) template.FuncMap {
...
 				return defaultActionBody
 			}
 			return body
 		},
 		"printResp": func(name string) bool {
-			if actionImpls == nil {
-				return true
-			}
 			_, ok := actionImpls[name]
 			if !ok {
```
raphael: You can also replace
```go
if !ok {
    return true
}
return false
```
with
```go
return !ok
```
:)

tanatana: 🙈

raphael:
Thank you! one last request I promise :) could you port this change to `master` please? just branch off of master and cherry-pick this commit into it then submit a new PR. Thanks again!

Issue #1454, opened 2017-11-22
goagen is also deleting Subversion .svn folders


dkeza:
Using goagen v1.3.0 on Windows 10 with svn (Subversion) version 1.5.6.

When I am generating new code with goagen, goagen is deleting this folders:

```
app\test
app
client
swagger
tool\cli
tool
```
and creating those folders again, with new files. That is all fine.

But, unfortunately, svn 1.5.6 relies on readonly hidden folder .svn in every child folder, in order to detect code changes, and goagen is also deleting all .svn subfolders. Because of that, my source version control is not working after running goagen.

Maybe we can add new flag to goagen command -i with name of folder which should be ignored?
I looked into code, and have found call to os.RemoveAll command in various places in various go files. Perhaps we should call some wrapper function, where this ignored folder should be checked?

I must say, I am newbie in Go, but I would gladly try to create PR with this change, when we need it.

raphael:
Thank you for the report! I would gladly accept a PR that fixed this. 

I think what needs to be done is look at the `gen_app`, `gen_client`, `gen_schema`, `gen_swagger` and `gen_js` packages and for each replace the call to remove the output directory with one that removes all the files in it except for a white-list which would only contain `.svn` for now (maybe `.git` too to allow for git submodules? - not sure). Note that the `gen_main` and `gen_controller` don't need to be changed because then don't delete folders.

PR #1465, opened 2017-12-03
Don't delete .svn directories

dkeza:
As described in #1454, goagen is not simply deleting all files and directories, but leaving directories and deleting only files. Files in .svn files are not deleted. In function allowDeleteDir is possible to add other directories, from which files should not be deleted.

### DIFF
--- a/goagen/gen_app/test_generator.go
+++ b/goagen/gen_app/test_generator.go
@@ -12,12 +12,11 @@ import (
 
 	"github.com/goadesign/goa/design"
 	"github.com/goadesign/goa/goagen/codegen"
-	"github.com/goadesign/goa/goagen/utils"
 )
 
 func makeTestDir(g *Generator, apiName string) (outDir string, err error) {
 	outDir = filepath.Join(g.OutDir, "test")
-	if err = utils.RemoveFiles(outDir); err != nil {
+	if err = os.RemoveAll(outDir); err != nil {
 		return
 	}
 	if err = os.MkdirAll(outDir, 0755); err != nil {
--- a/goagen/gen_app/test_generator.go
+++ b/goagen/gen_app/test_generator.go
@@ -12,11 +12,12 @@ import (
 
 	"github.com/goadesign/goa/design"
 	"github.com/goadesign/goa/goagen/codegen"
+	"github.com/goadesign/goa/goagen/utils"
 )
 
 func makeTestDir(g *Generator, apiName string) (outDir string, err error) {
 	outDir = filepath.Join(g.OutDir, "test")
-	if err = os.RemoveAll(outDir); err != nil {
+	if err = utils.RemoveFiles(outDir); err != nil {
 		return
 	}
 	if err = os.MkdirAll(outDir, 0755); err != nil {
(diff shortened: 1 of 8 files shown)


goagen/gen_app/test_generator.go
```diff
@@ -12,11 +12,12 @@ import (
 
 	"github.com/goadesign/goa/design"
 	"github.com/goadesign/goa/goagen/codegen"
+	"github.com/goadesign/goa/goagen/utils"
 )
 
 func makeTestDir(g *Generator, apiName string) (outDir string, err error) {
 	outDir = filepath.Join(g.OutDir, "test")
-	if err = os.RemoveAll(outDir); err != nil {
```
raphael: Do not change this as it deletes temporary folders created by tests

goagen/utils/files_test.go
```diff
@@ -0,0 +1,84 @@
...
+// Expects that RemoveFiles delete only files, and leaving directory structure intact.
+// Files from .svn directory should not be deleted.
+func TestRemoveFiles(t *testing.T) {
+
+	//Creating temp directory. When error occures, temp directory would remain, for visual check.
+	testDir := filepath.Join(".", time.Now().Format("20060102150405"))
+	currentDir := testDir
+	createDir(currentDir, t)
+
+	currentDir = testDir + sep + "app"
```
raphael: you can use `filepath.Join(testDir, "app")` instead of the above which might be a bit more readable.

goagen/gen_app/test_generator.go
dkeza: When I don't change it here, then is my .svn directory deleted in app/test. Because of that I changed it also at this place.

raphael: Sorry you're right, I got confused and thought this was the reason for the build failures.

raphael:
Do you have a handle on why the build is failing? you should be able to run `make` locally to reproduce the failures.

dkeza:
I will check why is build failing.

dkeza:
I had no success with make on my system (Linux Mint 18.3 64 bit). Somehow it doesn't work. But I think issue is that test functions expects that those directories doesn't exists. And now, I am only deleting files, and some part of code is detecting this as error.

This are this 2 files:
```
gen_app\test_generator_test.go
gen_schema\generator_test.go
```

BTW, what means
```
Ω
```
in those files?

raphael:
Makes sense, seems the new logic should also delete the directory if it's empty or alternatively the tests should be updated (I don't believe there's a case where the deleted directory would not get re-created). Ω is the name of a function in the gomega package (alias for `Expect`): https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L103

dkeza:
Maybe is this to complicated? Perhaps we should do this what I have done as workaround?
In my calling script, I first copy all .svn directories in temp directory, then I call goagen, and on end I am copying my .svn folders back.

raphael:
yeah I have been wondering that too. This gets even more complicated with v2 where the folders that are created depend on the service defined in the design. Maybe a better approach would be to document your workaround in the website? Maybe in https://github.com/goadesign/goa.design/blob/master/content/implement/goagen.md?

dkeza:
It is to complicated to implement this change, so we are giving up from this.

PR #1468, opened 2017-12-04
print_mounted_endpoints

### LINKED ISSUE #1453: Print routes when mounting them
This feature involves a few changes:

1. Enhance the generated `Server` struct in `gen/http/<service>/server/server.go` with a new field that captures the mounts.
2. Enhance the generated server `main.go` so  that before starting to listen to requests it iterates through the server mounted handlers and logs them.

Item 1 could look something like:
```go
type (
    Server struct {
        // Service is the name of the service served by this HTTP server.
       Service string
        // Mounts lists the mounted HTTP handlers.
        Mounts []*Mount
        .....  // HTTP handlers as currently generated
    }

    Mount struct {
        // Method is the name of the service method served by the mounted HTTP handler.
        Method string
        // Verb is the HTTP method used to match requests to the mounted handler.
        Verb string
        // Pattern is the HTTP request path pattern used to match requests to the mounted handler.
        Pattern string
        // Handler is the mounted HTTP handler.
        Handler http.Handler
    }
)
```
and 2 could look like:
```go
    // Start HTTP server using default configuration, change the code to
    // configure the server as required by your service.
    srv := &http.Server{Addr: *addr, Handler: handler}
    go func() {
        for _, m := range <service>Server.Mounts {
            logger.Printf("[INFO] service %q method %q mounted on %s %s", 
                <service>Server.Service, m.Method, m.Verb, m.Pattern)
        }
        logger.Printf("[INFO] listening on %s", *addr)
        errc <- srv.ListenAndServe()
    }()
}()
```
Also note that the current code generator optimizes the case where there is no need for a `Server` struct to be generated if there are only file servers exposed by the service and no regular method. This optimization needs to be removed and the file server case handled properly as well.

nitinmohan87:
Fixes #1453

### DIFF
--- a/design/api.go
+++ b/design/api.go
@@ -115,6 +115,9 @@ func (a *APIExpr) Random() *Random {
 // EvalName is the qualified name of the expression.
 func (a *APIExpr) EvalName() string { return "API " + a.Name }
 
+// Hash returns a unique hash value for a.
+func (a *APIExpr) Hash() string { return a.Name }
+
 // Finalize makes sure there's one server definition.
 func (a *APIExpr) Finalize() {
 	if len(a.Servers) == 0 {
--- a/design/api.go
+++ b/design/api.go
@@ -116,7 +116,7 @@ func (a *APIExpr) Random() *Random {
 func (a *APIExpr) EvalName() string { return "API " + a.Name }
 
 // Hash returns a unique hash value for a.
-func (a *APIExpr) Hash() string { return a.Name }
+func (a *APIExpr) Hash() string { return "_api_+" + a.Name }
 
 // Finalize makes sure there's one server definition.
 func (a *APIExpr) Finalize() {
--- a/examples/cellar/gen/http/swagger/server/server.go
+++ b/examples/cellar/gen/http/swagger/server/server.go
@@ -9,10 +9,45 @@
 package server
 
 import (
+	"context"
 	"net/http"
 
+	swagger "goa.design/goa/examples/cellar/gen/swagger"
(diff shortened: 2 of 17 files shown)


design/api.go
```diff
@@ -115,6 +115,9 @@ func (a *APIExpr) Random() *Random {
 // EvalName is the qualified name of the expression.
 func (a *APIExpr) EvalName() string { return "API " + a.Name }
 
+// Hash returns a unique hash value for a.
+func (a *APIExpr) Hash() string { return a.Name }
```
raphael: We may need to specify a prefix here so there's no risk of clashing with other constructs.

examples/cellar/gen/http/swagger/server/server.go
```diff
@@ -9,11 +9,46 @@
...
+) *Server {
+	return &Server{
+		Mounts: []*MountPoint{
+			{"../../gen/http/openapi.json", "GET", "/swagger/swagger.json"},
+		},
+	}
+}
+
+// Service returns the name of the service served.
+func (s *Server) Service() string { return "swagger" }
```
raphael: Do we need this function?

nitinmohan87: Yes.. It is used here - https://github.com/goadesign/goa/blob/bc1a832c804fb1ebca28ab8f4289b213c1ce3f00/examples/cellar/cmd/cellarsvc/main.go#L154

raphael:
Looks great, thank you!

PR #1470, opened 2017-12-05
Completely bypass metrics

klauspost:
Instead of running go-metrics, we dump metric collection earlier.

This also allows overriding go-metrics with a compatible collector.

Backwards compatible, so we keep the NoOpSink and associated stuff.

Only overhead to "no" metrics is key normalization.

We have to use a mutex since atomic values cannot contain interfaces.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -5,7 +5,7 @@ package goa
 import (
 	"regexp"
 	"strings"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	"github.com/armon/go-metrics"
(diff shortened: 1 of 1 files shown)


metrics.go
```diff
@@ -64,8 +85,18 @@ func NewMetrics(conf *metrics.Config, sink metrics.MetricSink) (err error) {
...
-	metriks.Store(m)
+func SetMetrics(m Collector) {
+	metriksMu.Lock()
+	metriks = m
+	metriksMu.Unlock()
+}
+
+// GetMetrics returns goa's metrics collector adapter interface.
+func GetMetrics() Collector {
+	metriksMu.Lock()
```
raphael: Could we use a RWMutex instead and simply use a read lock here? otherwise this causes all concurrent requests to have to be synchronized.

klauspost:
In situations like this, where locks is extremely fine grained (basically only held while a pointer is copied) a simple `Mutex` will be faster and have less contention than an `RWMutex`, which has much more housekeeping to do.

raphael:
Would you mind making a PR against the `v1` branch as well? Thank you!

PR #1471, opened 2017-12-06
Update help for Description function

k3a:
Description can be added to almost anything (not sure how to describe secirity, docs and fileserver definitians)

Actually when I tried to open generated swagger in the online swagger editor (https://editor.swagger.io/) it complained that Response should have Description. And it was not mentioned in the docs that Description can be inside ResponseTemplate yet it indeed worked so I decided this help should be updated. :)

### DIFF
--- a/design/apidsl/api.go
+++ b/design/apidsl/api.go
@@ -99,7 +99,7 @@ func Version(ver string) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


raphael:
This is great, thank you. Do you think you could backport this to the `v1` branch? i.e. cherry-pick this commit and create a new PR against `v1`. Thank you!

PR #1476, opened 2017-12-14
Completely bypass metrics (->v1)

klauspost:
Instead of running go-metrics, we dump metric collection earlier.

This also allows overriding go-metrics with a compatible collector.

Backwards compatible, so we keep the NoOpSink and associated stuff.

Only overhead to "no" metrics is key normalization.


raphael:
Thank you!

PR #1478, opened 2017-12-22
add_comments_http_response

nitinmohan87:
Fixes aliaser tool when generating aliases for HTTP DSL

### DIFF
--- a/cmd/aliaser/main.go
+++ b/cmd/aliaser/main.go
@@ -295,7 +295,7 @@ import (
 	// alias implementations.
 	aliasConstT = `const (
 {{- range $i, $n := .Names}}
-{{- $commLen := len $.Comments }}{{- if gt $commLen 0 }}
+{{- if $.Comments }}
 {{ index $.Comments $i }}
 {{- end }}
 {{ $n }} = design.{{ $n }}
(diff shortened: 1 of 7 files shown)


raphael:
Thanks for the PR! The idea of a linter is to help find inconsistencies, it shouldn't force nonsensical code or comments :-) The comments were done that way intentionally. Would you mind putting them back but keeping the other fixes this PR contains? Thank you!

cmd/aliaser/main.go
```diff
@@ -295,7 +295,9 @@ import (
 	// alias implementations.
 	aliasConstT = `const (
 {{- range $i, $n := .Names}}
+{{- $commLen := len $.Comments }}{{- if gt $commLen 0 }}
```
raphael: I believe this can be simply expressed as:
```go
{{ if $.Comments }}
```

design/mapped_attribute.go
```diff
@@ -152,3 +152,16 @@ func (ma *MappedAttributeExpr) Merge(other *MappedAttributeExpr) {
 		}
 	}
 }
+
+// FindKey finds the given key in the mapped attribute expression.
```
raphael: Is this supposed to be in this PR?

nitinmohan87: I can create a separate PR if you would like. Just thought I would add this here so that we can fix the plugins build.

raphael:
Thank you!

PR #1481, opened 2017-12-26
fix xray.WrapDoer to update segment in the context (v1)

jim-slattery-rs:
(This is a `v1` companion PR to https://github.com/goadesign/goa/pull/1480)

This way, the `middleware.TraceDoer` will include the latest span ID in the headers of the outgoing request.

Otherwise, the X-Ray service map would misrepresent a single request from A to B as two outgoing requests:
- `A` -> `B(client-side span)`
- `A` -> `B`

Now the service map will represent the single request as:
- `A` -> `B`

### DIFF
--- a/middleware/xray/wrap_doer.go
+++ b/middleware/xray/wrap_doer.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/goadesign/goa/client"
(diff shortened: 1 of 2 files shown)


raphael:
Thank you!

PR #1498, opened 2018-01-05
fix_error_encoder_generation

nitinmohan87:


design/result_type.go
```diff
@@ -86,6 +86,7 @@ var (
 		{"code", &AttributeExpr{
 			Type:        String,
 			Description: "an application-specific error code, expressed as a string value.",
+			Metadata:    MetadataExpr{"error:field": []string{}},
```
raphael: I think you can just assign `nil` instead of `[]string{}`.

design/service.go
```diff
@@ -96,11 +102,37 @@ func (s *ServiceExpr) Finalize() {
...
+	if !ok {
+		return verr
+	}
+	if o := AsObject(rt); o != nil {
+		var errField string
+		for _, n := range *o {
+			if _, ok := n.Attribute.Metadata["error:field"]; ok {
+				if errField != "" {
+					verr.Add(e, "error:field already set for %q attribute in %q result type", errField, rt.Identifier)
+				} else {
```
raphael: Don't think you need the `else` here.

dsl/metadata.go
```diff
@@ -14,6 +14,22 @@ import (
 // While keys can have any value the following names are handled explicitly by
 // goa when set on attributes.
 //
+// `error:field`: maps the attribute name in custom result type to the error
```
raphael: I'm thinking `struct:error:name` might be more consistent for the name of the metadata. Maybe this is more clear:

>`struct:error:name` identifies the attribute of a result type used to select the returned error when multiple errors are defined on the same method. The value of the field corresponding to the attribute with the `struct:error:name` metadata is matched against the names of the method errors as defined in the design. This makes it possible to define distinct transport mappings for the various errors (for example to return different HTTP status codes). There must be one and exactly one attribute with the `struct:error:name` metadata defined on result types used to define error results.

http/codegen/server.go
```diff
@@ -823,17 +823,14 @@ func {{ .ErrorEncoder }}(encoder func(context.Context, http.ResponseWriter) goah
...
-		{{- range $errs := .Errors }}
-		case {{ .Ref }}:
+	{{- range $ref := .Errors.Refs }}
+		case {{ $ref }}:
+			{{- range $.Errors.Get $ref }}
 
-	       {{- with .Response}}
-			{{- if .TagName }}
-				{{- if .TagRequired }}
+	      {{- with .Response}}
```
raphael: indenting seems off...

nitinmohan87: weird why I am having indenting issues.. never had this problem before.. probably an editor update messed up with my settings.

dsl/metadata.go
nitinmohan87: sounds good

raphael:
Looks great, thank you!

PR #1501, opened 2018-01-10
disallow_map_params_headers

### LINKED ISSUE #1479: Properly error out when an unsupported type is used as query string
`goa` generates incorrect code instead of erroring out when using a query string parameter of type `MapOf(String, Any)` for example. Properly catch unsupported types and report an error instead. Same applies to headers and path parameters.

nitinmohan87:
Fixes #1479


nitinmohan87:
Closing this in favor of https://github.com/goadesign/goa/pull/1508

PR #1508, opened 2018-01-24
Add MapParam DSL for HTTP

### LINKED ISSUE #1479: Properly error out when an unsupported type is used as query string

nitinmohan87:
Fixes #1479 
Adds a MapParams DSL to the HTTP endpoint expression which maps the query params to an entire payload (of map type) or an attribute in the payload object (attribute must be map type).


http/codegen/server.go
```diff
@@ -443,6 +444,27 @@ func {{ .RequestDecoder }}(mux goahttp.Muxer, decoder func(*http.Request) goahtt
...
+		{{- if .Required }}
+		if len({{ .VarName }}Raw) == 0 {
+			err = goa.MergeErrors(err, goa.MissingFieldError("{{ .Name }}", "query string"))
+		}
+		{{- end }}
+
+		{{- if not .Required }}
+		if len({{ .VarName }}Raw) != 0 {
+		{{- end }}
+			err2 := json.Unmarshal([]byte({{ .VarName }}Raw), &{{ .VarName }})
```
raphael: how does this work? {{ .VarName }}Raw would be of type `url.Values` which is `map[string][]string` - how can this be cases to `[]byte` and then JSON unmarhsalled? In general I don't think we should use JSON to marhsal/unmarshal. The encoding / decoding code everywhere type asserts or type converts as needed and we should do that here too.

http/design/endpoint.go
```diff
@@ -36,6 +36,12 @@ type (
 		// Metadata is a set of key/value pairs with semantic that is
 		// specific to each generator.
 		Metadata design.MetadataExpr
+		// MapQueryParams indicates that the query params are mapped to the
+		// payload in the method expression. If the pointer refers
+		// to a non-empty string, the query params are mapped to an attribute
+		// in the payload with the same name. If the pointer empty string,
```
raphael: missing "is an"

http/codegen/client.go
```diff
@@ -204,7 +218,7 @@ func {{ .RequestEncoder }}(encoder func(*http.Request) goahttp.Encoder) func(*ht
 			{{- end }}
 		{{- end }}
 	{{- end }}
-	{{- if .Payload.Request.QueryParams }}
+	{{- if or .Payload.Request.QueryParams }}
```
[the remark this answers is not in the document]
nitinmohan87: My mistake. Will take it out.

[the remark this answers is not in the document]
raphael: cool you probably know already but it was in another place as well.

http/codegen/server.go
nitinmohan87: `r.URL.Query().Get()` returns a string and it gets unmarshalled into `map[string][]string`.

raphael:
Thank you!

http/codegen/client.go
```diff
@@ -218,11 +218,24 @@ func {{ .RequestEncoder }}(encoder func(*http.Request) goahttp.Encoder) func(*ht
...
+		for key, value := range p{{ if .FieldName }}.{{ .FieldName }}{{ end }} {
+			{{ template "type_conversion" (typeConversionData .Type.KeyType.Type "keyStr" "key") }}
+			{{- if eq .Type.ElemType.Type.Name "array" }}
+			for _, val := range value {
+				{{ template "type_conversion" (typeConversionData .Type.ElemType.Type.ElemType.Type (printf "%sStr" "val") "val") }}
+				values.Add(keyStr, valStr)
+			}
+			{{- else }}
+			{{ template "type_conversion" (typeConversionData .Type.ElemType.Type (printf "%sStr" "value") "value") }}
+      values.Add(keyStr, valueStr)
```
raphael: indenting ^

Issue #1527, opened 2018-02-08
goa v2: Response in Service() section segfaults


psschroeter:
Response in Service() section segfaults in some DupAttribute method of goa gen. Same Response in Method() section seems to work as expected.

Example code:
https://gist.github.com/psschroeter/620b06b655f67df0fa54cf51b394445a

raphael:
This should be fixed in the latest v2 - can you please verify and if so close this ticket? thank you!

psschroeter:
Works now, thanks

Issue #1526, opened 2018-02-08
goa v2: Error() in service section generates improper code


psschroeter:
branch: v2 (master)
Error() in Service() section generates bad code. I.e. compiling with the code generates errors like:
gen/http/dummy/client/types.go:29:47: undefined: dummysvc.Error
gen/http/dummy/client/types.go:30:8: undefined: dummysvc.Error

If the same Error() is specified in a Method() section, everything seems to work fine.

Example gist:
https://gist.github.com/psschroeter/620b06b655f67df0fa54cf51b394445a

raphael:
This should be fixed in the latest v2 - can you please verify and if so close this ticket? thank you!

psschroeter:
Works now, thanks

PR #1529, opened 2018-02-08
Fix code generation when error response defined in service-level

nitinmohan87:
* Fixes an issue where the error struct is not initialized in `service.go` if `Error` is defined at the service level (#1526)
* Fixes segfault when error response is defined at the service level (#1527)

### DIFF
--- a/http/design/endpoint.go
+++ b/http/design/endpoint.go
@@ -554,19 +554,17 @@ func (e *EndpointExpr) Finalize() {
 		}
 	}
 
-	// Make sure error types are user types and have a body.
-	for _, herr := range e.HTTPErrors {
-		herr.Finalize(e)
-	}
-
 	// Inherit HTTP errors from service and root
 	for _, r := range e.Service.HTTPErrors {
-		r.Finalize(e)
 		e.HTTPErrors = append(e.HTTPErrors, r.Dup())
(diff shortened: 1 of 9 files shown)


http/design/endpoint.go
```diff
@@ -554,18 +554,20 @@ func (e *EndpointExpr) Finalize() {
...
 	}
 
+	// Make sure error types are user types and have a body.
+	for _, herr := range e.HTTPErrors {
+		herr.Finalize(e)
+	}
+
 	// Inherit HTTP errors from service and root
 	for _, r := range e.Service.HTTPErrors {
+		r.Finalize(e)
```
raphael: Can you explain why this change is necessary? (maybe add a comment to the code with that explanation). Why can't the `Finalize` be called on the dups which would seem safer?

raphael:
Thank you! looks great. I left one question in the code changes.

http/design/endpoint.go
nitinmohan87: When we dup the error expression, we also dup the HTTP response expression underneath the error (which means "dup"ing the response body and headers). When the HTTP error expression is defined at the API/service level the headers and body are not initialized when "dup"ing which results in nil pointer deference.

PR #1544, opened 2018-02-16
Initialize HTTP request context in BuildX request function

### LINKED ISSUE #1539: goa v2: http client/server code does not use context
The generated code for the client which makes the endpoint function doesn't do anything with the passed in context. 

For example, this dummy service code linked here: https://gist.github.com/psschroeter/dca69aa15ad9a08805799727060accca

Generates an upload function in gen/http/dummy/client with an Upload endpoint function:
```
// Show returns a endpoint that makes HTTP requests to the dummy service show
// server.
func (c *Client) Show() goa.Endpoint {
	var (
		encodeRequest  = EncodeShowRequest(c.encoder)
		decodeResponse = DecodeShowResponse(c.decoder, c.RestoreResponseBody)
	)
	return func(ctx context.Context, v interface{}) (interface{}, error) {
		req, err := c.BuildShowRequest(v)
		if err != nil {
			return nil, err
		}
		err = encodeRequest(req, v)
		if err != nil {
			return nil, err
		}

		resp, err := c.ShowDoer.Do(req)

		if err != nil {
			return nil, goahttp.ErrRequestError("dummy", "show", err)
		}
		return decodeResponse(resp)
	}
}
```
ctx isn't used. Seems like there should be a req = req.WithContext(ctx,req) in there somewheres.

nitinmohan87:
Fixes #1539

### DIFF
--- a/examples/calc/gen/http/calc/client/client.go
+++ b/examples/calc/gen/http/calc/client/client.go
@@ -49,14 +49,14 @@ func NewClient(
 	}
 }
 
-// Add returns a endpoint that makes HTTP requests to the calc service add
+// Add returns an endpoint that makes HTTP requests to the calc service add
 // server.
 func (c *Client) Add() goa.Endpoint {
 	var (
 		decodeResponse = DecodeAddResponse(c.decoder, c.RestoreResponseBody)
 	)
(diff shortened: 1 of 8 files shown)


raphael:
Great, thank you!

PR #1545, opened 2018-02-16
Check if validation format is supported

### LINKED ISSUE #1537: goa v2: panic when a unsupported format supplied on attribute
When supplying an string of an unsupported format type on an attribute. `goa gen` panics instead of retuning a friendly error stating the line that the invalid data was supplied.

Reproducing Design:
```golang
package design

var _ = API("foo", func() {
	Title("Foo Service")
})

var _ = Service("foo", func() {
	Method("version", func() {
		Payload(func() {
			Attribute("a", func() {
				Format("foobar")
			})
			Required("a")
		})
		Result(Int)
		HTTP(func() {
			GET("/foo/{a}")
			Response(StatusOK)
		})
	})
})
```

nitinmohan87:
Fixes #1537

### DIFF
--- a/design/attribute.go
+++ b/design/attribute.go
@@ -511,3 +511,36 @@ func (v *ValidationExpr) Dup() *ValidationExpr {
 		Required:  req,
 	}
 }
+
+// IsSupportedValidationFormat checks if the validation format is supported by goa.
+func (a *AttributeExpr) IsSupportedValidationFormat(vf ValidationFormat) bool {
+	switch vf {
+	case FormatDateTime:
+		return true
+	case FormatUUID:
+		return true
+	case FormatEmail:
+		return true
+	case FormatHostname:
+		return true
+	case FormatIPv4:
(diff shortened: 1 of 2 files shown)


design/attribute.go
```diff
@@ -511,3 +511,36 @@ func (v *ValidationExpr) Dup() *ValidationExpr {
 		Required:  req,
 	}
 }
+
+// IsSupportedValidationFormat checks if the validation format is supported by goa.
+func (a *AttributeExpr) IsSupportedValidationFormat(vf ValidationFormat) bool {
```
raphael: This test seems a bit odd - not sure we need to protect against user force casting to `ValidationFormat`? Maybe a better way to approach this is to make the `ValidationFormat` type private - this way DSL writers would be forced to use the public values defined for the enum.

raphael:
Merging this for now, we can revisit later.

Issue #1423, opened 2017-11-10
goa v2: add missing tests.


raphael:
Go through goa v2 packages and increase test coverage. In particular:

* Add tests to design packages
* Add tests to DSL packages
* Add tests to goa package

tchssk:
This is still not over.

tchssk:
These functions and methods are very simple. It is suitable for first-time contributors.

* [ ] `design.APIExpr.EvalName()`
* [ ] `design.APIExpr.Hash()`
* [ ] `design.ContactExpr.EvalName()`
* [ ] `design.DocsExpr.EvalName()`
* [ ] `design.LicenseExpr.EvalName()`
* [ ] `design.ServerExpr.Attribute()`
* [ ] `design.ServerExpr.EvalName()`
* [ ] `design.ServerParamExpr.EvalName()`

raphael:
Thank you for identifying these! Would be great to have individual issues for each labeled with `v2` and `good first issue`. Maybe the issue description could point to an example (one of the tests you've written).

PR #1546, opened 2018-02-17
Fix the condition to get schemes in APIExpr

tchssk:
It should get only schemes which can be parsed and not empty.


raphael:
whoops, thank you! Could I convince you to also add a test please?

tchssk:
https://github.com/goadesign/goa/pull/1546/commits/685a4bb3a7f44e5d778c430a772dca0ffea6fd61 adds tests for `design.APIExpr.Schemes()`. https://github.com/goadesign/goa/issues/1423

raphael:
Perfect, thank you!

PR #1548, opened 2018-02-19
Remove the unused valiable

ikawaha:
Go version 1.10,  `go vet` command strictly detects not used variables in tests as follows:

```
$ go vet ./app/test/
# cellar/app/test
app/test/bottle_testing.go:36:3: resp declared but not used
```
```go
func ShowBottleNotFound(t goatest.TInterface, ctx context.Context, service *goa.Service, ctrl app.BottleController, bottleID interface{}) http.ResponseWriter {
        // Setup service
        var (
                logBuf bytes.Buffer
                resp   interface{} // ← ★

                respSetter goatest.ResponseSetterFunc = func(r interface{}) { resp = r }
```

This patch fix it to generate it only when the `resp` is used.


raphael:
Great! thank you.

raphael:
Can you please port that fix to the `v1` branch as well? Thank you!

PR #1558, opened 2018-03-01
Add tests for design.AttributeExpr.AllRequired()

tchssk:
#1423

### DIFF
--- a/design/attribute_test.go
+++ b/design/attribute_test.go
@@ -145,3 +145,50 @@ func TestAttributeExprValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestAttributeExprAllRequired(t *testing.T) {
+	cases := map[string]struct {
+		typ        DataType
+		validation *ValidationExpr
+		expected   []string
+	}{
+		"user type": {
+			typ: &UserTypeExpr{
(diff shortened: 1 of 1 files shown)


design/attribute_test.go
```diff
@@ -145,3 +145,50 @@ func TestAttributeExprValidate(t *testing.T) {
...
+			typ: &UserTypeExpr{
+				AttributeExpr: &AttributeExpr{
+					Validation: &ValidationExpr{
+						Required: []string{"foo", "bar"},
+					},
+				},
+			},
+			expected: []string{"foo", "bar"},
+		},
+		"not user type": {
```
raphael: I'm wondering if we need to test for this case? it works but having required attributes on a type that is not an object does not make a lot of sense.

tchssk: It makes sence. I remove the case and rename `user type` to `some required`.

tchssk: Fixed it. https://github.com/goadesign/goa/pull/1558/commits/581b2014316b876a5d61797b5bfc8bf0b7f8c7aa

raphael: Thank you!

PR #1560, opened 2018-03-03
Fix codegen imports on Windows

douglaswth:
When running `goa example` on Windows, the import paths had `\` instead
of `/` since `filepath.Join` was being used instead of `path.Join`. For
example:

```golang
package main

import (
    "context"
    ...
    "github.com\douglaswth\blah"
)

...
```

This will not compile since `\` is used for escape sequences, etc.

### DIFF
--- a/http/codegen/example_server.go
+++ b/http/codegen/example_server.go
@@ -2,7 +2,7 @@ package codegen
 
 import (
 	"os"
-	gopath "path"
+	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


raphael:
Nice catch, thank you! Could I convince you to rename the local variable `path` to something else so there's no need to rename the package in the import?

douglaswth:
Done.

raphael:
Thank you!

PR #1565, opened 2018-03-05
Add multipart request support

nitinmohan87:


examples/cellar/gen/http/storage/client/client.go
```diff
@@ -43,6 +49,10 @@ type Client struct {
 	decoder func(*http.Response) goahttp.Decoder
 }
 
+// StorageMultiAddEncoderFunc is the type to encode multipart request for the
+// "storage" service "multi_add" endpoint.
+type StorageMultiAddEncoderFunc func(*multipart.Writer, *[]*storage.Bottle) error
```
raphael: Shouldn't the type used by `[]*storage.Bottle` and not `*[]*storage.Bottle`?

examples/cellar/gen/http/storage/server/encode_decode.go
```diff
@@ -193,6 +193,49 @@ func DecodeRateRequest(mux goahttp.Muxer, decoder func(*http.Request) goahttp.De
...
+// NewStorageMultiAddDecoder returns a decoder to decode the multipart request
+// for the "storage" service "multi_add" endpoint.
+func NewStorageMultiAddDecoder(storageMultiAddDecoderFn StorageMultiAddDecoderFunc) func(r *http.Request) goahttp.Decoder {
+	return func(r *http.Request) goahttp.Decoder {
+		return goahttp.EncodingFunc(func(v interface{}) error {
+			mr, err := r.MultipartReader()
+			if err != nil {
+				return err
+			}
+			p := v.(*[]*storage.Bottle)
```
raphael: Same as above

http/codegen/client.go
```diff
@@ -466,3 +491,27 @@ const singleResponseT = `case {{ .StatusCode }}:
...
+`
+
+// input: multipartData
+const multipartRequestEncoderT = `{{ printf "%s returns an encoder to encode the multipart request for the %q service %q endpoint." .InitName .ServiceName .MethodName | comment }}
+func {{ .InitName }}(encoderFn {{ .FuncName }}) func(r *http.Request) goahttp.Encoder {
+	return func(r *http.Request) goahttp.Encoder {
+		body := &bytes.Buffer{}
+		mw := multipart.NewWriter(body)
+		return goahttp.EncodingFunc(func(v interface{}) error {
+			p := v.(*{{ .PayloadRef }})
```
raphael: That hard-coded `*` doesn't seem right...

http/codegen/server.go
```diff
@@ -964,3 +1000,24 @@ const responseT = `{{ define "response" -}}
 	{{- end }}
 {{- end -}}
 `
+
+// input: multipartData
+const multipartRequestDecoderTypeT = `{{ printf "%s is the type to decode multipart request for the %q service %q endpoint." .FuncName .ServiceName .MethodName | comment }}
+type {{ .FuncName }} func(*multipart.Reader, *{{ .PayloadRef }}) error
```
raphael: ditto

http/codegen/server.go
```diff
@@ -964,3 +1000,24 @@ const responseT = `{{ define "response" -}}
...
+// input: multipartData
+const multipartRequestDecoderT = `{{ printf "%s returns a decoder to decode the multipart request for the %q service %q endpoint." .InitName .ServiceName .MethodName | comment }}
+func {{ .InitName }}({{ .VarName }} {{ .FuncName }}) func(r *http.Request) goahttp.Decoder {
+	return func(r *http.Request) goahttp.Decoder {
+		return goahttp.EncodingFunc(func(v interface{}) error {
+			mr, err := r.MultipartReader()
+			if err != nil {
+				return err
+			}
+			p := v.(*{{ .PayloadRef }})
```
raphael: another one

examples/cellar/gen/http/storage/client/client.go
nitinmohan87: Fixed

examples/cellar/gen/http/storage/server/encode_decode.go
nitinmohan87: I think here we have to pass a pointer so that the underlying decoder func can populate this variable.

http/codegen/client.go
nitinmohan87: Fixed

http/codegen/server.go
nitinmohan87: We need to have a pointer here.

http/codegen/server.go
nitinmohan87: Same here. We must pass a pointer here.

http/dsl/http.go
```diff
@@ -537,6 +537,34 @@ func MapParams(args ...interface{}) {
 	e.MapQueryParams = &mapName
 }
 
+// MultipartRequest defines the HTTP request for the endpoint to be a
```
raphael: Maybe this is a bit clearer?
```
MultipartRequest indicates that HTTP requests made to the method use MIME multipart encoding as defined in RFC 2046.
```

http/dsl/http.go
```diff
@@ -537,6 +537,34 @@ func MapParams(args ...interface{}) {
...
+// multipart content type.
+//
+// MultipartRequest must appear in a HTTP endpoint expression.
+//
+// When an endpoint is defined to have multipart request, the code generator
```
raphael: Maybe something like the following is clearer?

>goa generates a custom encoder that writes the payload for requests made to HTTP endpoints that use MultiPartRequest. The generated encoder accept a user provided function that does the actual mapping of the payload to the multipart content. The user provided function accepts a multipart writer and a reference to the payload and is responsible for encoding the payload. goa also generates a custom decoder that reads back the multipart content into the payload struct. The generated decoder also accepts a user provided function that takes a multipart reader and a reference to the payload struct as parameter. The user provided decoder is responsible for decoding the multipart content into the payload. The example command generates a default implementation for the user decoder and encoder.

raphael:
Thank you!

PR #1570, opened 2018-03-07
Check if attribute also exists in Bases (via Extend)

### LINKED ISSUE #1567: Attributes from Extend not available in View
## Expected Behavior

When using `Extend` in a type definition, I would expect to be able to use attributes from the extended type in a `View` declaration.

## Actual Behavior

Instead, it seems that `View` only accepts attributes that are defined directly on the type.

Given the following file, `goa gen ...` gives an error:
```golang
package public

import (
    . "goa.design/goa/http/dsl"
)

var Thingy = Type("Thingy", func() {
    Attribute("name", String, "The name of the thingy")
})

var CoolThingy = ResultType("application/vnd.example.coolthingy", func() {
    Extend(Thingy)
    Attributes(func() {
        Attribute("coolness", UInt, "How cool this thingy is")
        Required("name", "coolness")
    })
    View("default", func() {
        Attribute("name")
        Attribute("coolness")
    })
})
```
The error message is:
```
[design\public\types.go:18] unknown attribute "name" in attribute
```
Line `18` is the beginning of the `View` declaration.

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Windows

nitinmohan87:
Fixes #1567

### DIFF
--- a/design/attribute.go
+++ b/design/attribute.go
@@ -161,7 +161,7 @@ func (a *AttributeExpr) Validate(ctx string, parent eval.Expression) *eval.Valid
 	verr.Merge(a.validateEnumDefault(ctx, parent))
 	if o := AsObject(a.Type); o != nil {
 		for _, n := range a.AllRequired() {
-			if a.FindRecursive(n) == nil {
+			if a.Find(n) == nil {
 				verr.Add(parent, `%srequired field %q does not exist`, ctx, n)
 			}
 		}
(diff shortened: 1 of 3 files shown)


raphael:
Thank you! I'm wondering whether the issue could pop up in other places. It seems the fact that there are bases should be more transparent. Is there a reason we couldn't make `Attribute` itself recursive? Maybe we can copy the attributes in the `Finalize` or is that too late? Another potential issue I see is that the algorithm is not protected against infinite recursion. It's possible for a child attribute to point to a parent (e.g. recursive type definitions).

raphael:
Thank you!

PR #1571, opened 2018-03-07
Fix codegen.AppendHelpers to not add already existing helper

### LINKED ISSUE #1568: Conflicting marshal methods for imported attribute types
## Expected Behavior

Should be able to import the same type for multiple attributes.

## Actual Behavior

When importing the same type for multiple attributes within a design type, it generates conflicting methods with identical names.

Something like this...

```go
var result = ResultType(func() {
   Attributes(func() {
       Attribute(“created_by”, imported.User)
       Attribute(“updated_by”, imported.User)
   })\n
})
```

... generates two methods `marshalUserToUser` in .../encode_decode.go that conflict.

## Specifications

  - goa Version: V2
  - Platform (Linux/Windows etc.): OSX

nitinmohan87:
Fixes #1568


raphael:
Awesome thank you!

PR #1580, opened 2018-03-09
Fix new go vet cancel function context leak issue

douglaswth:
- With Go 1.10 there is now a `go vet` error on the generated and
  example server commands: `the cancel function returned by
  context.WithTimeout should be called, not discarded, to avoid a context
  leak`


raphael:
Thank you!

Issue #1270, opened 2017-06-16
Support For Posting MultiPart File Form


EarlCoder:
Is there any plans for adding file upload support to the design spec that would generate proper operand/swagger?

`Param("file", MultipartFile, "File to upload")`

raphael:
I'm not planning on adding this as I'm currently focusing on v2. But I would be happy to help with putting together a PR.

PR #1582, opened 2018-03-11
Add a DSL MultipartForm

tchssk:
Related to https://github.com/goadesign/goa/issues/1270 (and #1565 for v2).

This pull request adds brand new DSLs `Multipart` and `OptionalMultipart`. It makes it possible to handle `multipart/form-data` using goa. I'v already done the test with https://github.com/goadesign/examples/tree/master/form_encoding but you might find some issues. I would appreciate any comments or suggestions.

### DIFF
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -485,7 +485,7 @@ func payload(isOptional, isMultipart bool, p interface{}, dsls ...func()) {
 // the HTTP request body should be encoded using multipart form data as described in
 // https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2.
 //
-func MultipartForm(p interface{}, dsls ...func()) {
+func MultipartForm() {
 	if a, ok := actionDefinition(); ok {
 		a.PayloadMultipart = true
 	}
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -403,7 +403,7 @@ func Params(dsl func()) {
 //	})
 //
 func Payload(p interface{}, dsls ...func()) {
-	payload(false, false, p, dsls...)
+	payload(false, p, dsls...)
 }
 
 // OptionalPayload can be used in: Action
@@ -415,10 +415,10 @@ func Payload(p interface{}, dsls ...func()) {
 //	OptionalPayload(BottlePayload)		// Request payload is described by the BottlePayload type and is optional
 //
 func OptionalPayload(p interface{}, dsls ...func()) {
-	payload(true, false, p, dsls...)
+	payload(true, p, dsls...)
 }
 
-func payload(isOptional, isMultipart bool, p interface{}, dsls ...func()) {
+func payload(isOptional bool, p interface{}, dsls ...func()) {
 	if len(dsls) > 1 {
 		dslengine.ReportError("too many arguments given to Payload")
 		return
@@ -437,7 +437,6 @@ func payload(isOptional, isMultipart bool, p interface{}, dsls ...func()) {
 			if len(dsls) == 0 {
 				a.Payload = actual
 				a.PayloadOptional = isOptional
-				a.PayloadMultipart = isMultipart
 				return
 			}
 			att = design.DupAtt(actual.Definition())
@@ -475,7 +474,6 @@ func payload(isOptional, isMultipart bool, p interface{}, dsls ...func()) {
 			TypeName:            fmt.Sprintf("%s%sPayload", an, rn),
 		}
 		a.PayloadOptional = isOptional
-		a.PayloadMultipart = isMultipart
 	}
 }
 
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -403,7 +403,7 @@ func Params(dsl func()) {
 //	})
 //
 func Payload(p interface{}, dsls ...func()) {
-	payload(false, p, dsls...)
+	payload(false, false, p, dsls...)
 }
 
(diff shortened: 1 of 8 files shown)


raphael:
Thank you for the pull request! 

If I understand correctly, this PR adds the ability to use form encoding. Form encoding can be done using multipart or `application/x-www-form-urlencoded` MIME  encoding (see section 13.4 in https://www.w3.org/TR/html401/interact/forms.html). It would seem that "FormEncoded" might be a better term than "Multipart"?

I am a little hesitant to put this as part of the core goa code generators. As you've seen goa now supports multipart requests with #1565 which is more generic. There is value though in generating the form encoding and decoding algorithms though (like you have done in the PR here). Maybe this could be a "HTTP Form Encoding" plugin? The plugin would add a "FormEncoded()" DSL that does the same thing "Multipart()" does in this PR.

tchssk:
Yes. This PR adds the ability to use multipart form encoding.

goa has already support `application/x-www-form-urlencoded` (using DSLs `Comsumes` and `Package`) so this PR is focusing on `multipart/form-data`.

At first, I tried to implement this as an additional encoder at `github.com/goadesign/goa/encoding/multipart` but decoding `multipart/form-data` needs the boundary in each request header. It was impossible to implement on the interface `goa.Docoder`.

What I want to do is:

* Make it possible to generate [`in: formData`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterIn) in `gen_swagger`.
* Make it possible to map each part in multipart to corresponding payload struct field in `gen_app`.
* Make it possible to send a request as multipart in `gen_client`.
* Add a DSL to indicate the request is multipart.

Implementing as plugin may be a good idea but I'm not sure whether it's going to implement that.

raphael:
Thank you for the clarification - it makes sense. I completely missed the fact that this was for v1 and not v2... 

Maybe we could call the DSL functions `MultipartForm` and `OptionalMultipartForm` to be explicit? I'll review the comment as well in the PR.

tchssk:
I think so too. MultipartForm is more explicit.

As another issue, there is another idea about how to use the DSL.

Now this PR contains following:

```go
Action("show", func() {
	MultipartForm(FooPayload)
})```

and here is another idea:

```go
Action("show", func() {
	Payload(FooPayload)
	MultipartForm()
});

**Comment by raphael** [comment]
I like the second idea, i.e. keeping `Payload` but then qualifying with `MultipartForm()`

**Comment by tchssk** [comment]
https://github.com/goadesign/goa/pull/1582/commits/a26d9b771f63251fe7339e4e934631cff36c12c0 adds a DSL `MultipartForm` and remove `Multipart` and `OptionalMultipart`.

File: design/apidsl/action.go [hunk 1]
```diff
@@ -474,6 +475,19 @@ func payload(isOptional bool, p interface{}, dsls ...func()) {
...
+	}
+}
+
+// MultipartForm can be used in: Action
+//
+// MultipartForm implements the action multipart form DSL. An action multipart form indicates that
+// the HTTP request body should be encoded using multipart form data as described in
+// https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2.
+//
+func MultipartForm(p interface{}, dsls ...func()) {
```
raphael: Do we really need arguments to `MultipartForm`?

tchssk: It's my mistake. https://github.com/goadesign/goa/pull/1582/commits/d4f73d7acc8a2388ab1b402c69667c43ae50c6a0 removes aruments of `MultipartForm`.

design/apidsl/action.go
```diff
@@ -403,7 +403,7 @@ func Params(dsl func()) {
 //	})
 //
 func Payload(p interface{}, dsls ...func()) {
-	payload(false, p, dsls...)
+	payload(false, false, p, dsls...)
```
raphael: The second argument to `payload`can be removed now I think...

design/apidsl/action.go
```diff
@@ -415,10 +415,10 @@ func Payload(p interface{}, dsls ...func()) {
 //	OptionalPayload(BottlePayload)		// Request payload is described by the BottlePayload type and is optional
 //
 func OptionalPayload(p interface{}, dsls ...func()) {
-	payload(true, p, dsls...)
+	payload(true, false, p, dsls...)
 }
 
-func payload(isOptional bool, p interface{}, dsls ...func()) {
+func payload(isOptional, isMultipart bool, p interface{}, dsls ...func()) {
```
raphael: See above, seems we don't need the `isMultipart` parameter now that we use a different DSL.

design/apidsl/action.go
tchssk: Sorry, fixed it. https://github.com/goadesign/goa/pull/1582/commits/f7d765af7f76651b72c190e2c094f017f2ad25af

design/apidsl/action.go
tchssk: Sorry, fixed it. https://github.com/goadesign/goa/pull/1582/commits/f7d765af7f76651b72c190e2c094f017f2ad25af

raphael:
Looks great! let me know if you think it's ready to merge, it looks good to me.

tchssk:
Thank you. I think it's ready to merge.

tchssk:
Thank you! I'm planning to add a example to https://github.com/goadesign/examples.

raphael:
sounds great and thank YOU :)

PR #1583, opened 2018-03-12
validate different wildcards only if request method is the same

kovetskiy:
Hi. I noticed that goa design doesn't allow to use different wildcards at the same position for different methods. It's reasonable only if we check wildcard positions in URLs with the same HTTP method. But it's not reasonable to make this check if http methods are different.
My error was:
```
resource "decisions" action "list": route "/api/v1/decisions/:entity/:entityID/" conflicts with route "/api/v1/decisions/:decisionID" of decisions action delete. Make sure wildcards at the same positions have the same name. Conflicting wildcards are "entity" from route GET "/:entity/:entityID/" of resource "decisions" action "list" and "decisionID" from route DELETE "/:decisionID" of resource "decisions" action "delete".
```
So `DELETE "/:decisionID" ` was conflicting with `GET "/:entity/:entityID/"` but there are no really possible conflicts because http methods are different and mux can handle both without any problem.

I made a patch and tested it, works just fine.

### DIFF
--- a/design/validation.go
+++ b/design/validation.go
@@ -121,6 +121,9 @@ func (a *APIDefinition) Validate() error {
 			if route == other {
 				continue
 			}
+			if route.Route.Verb != other.Route.Verb {
+				continue
+			}
 			if strings.HasPrefix(route.Key, other.Key) {
 				diffs := route.DifferentWildcards(other)
 				if len(diffs) > 0 {
--- a/design/validation.go
+++ b/design/validation.go
@@ -116,8 +116,39 @@ func (a *APIDefinition) Validate() error {
 		})
 		return nil
 	})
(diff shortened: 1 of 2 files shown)


kovetskiy:
https://github.com/goadesign/goa/pull/1583/files?-w with `-w` looks more clear

raphael:
This is great, thank you. Could I convince you to add a test that would have failed before the fix but doesn't with the PR applied? Thank you!

kovetskiy:
@raphael I have added test that fails without my fixes and passes with my fixes.

raphael:
One last thing I promise :) could you backport the fix to the `v1` branch? Simply cherry pick the merge commit and create a new PR targeting the `v1` branch. Thank you!

Issue #1033, opened 2017-02-02
How I put JSON in Any Type


perigee:
I defined a MediaType with

```
var StatusMedia = MediaType("application/vnd.errormedia+json", func() {
	Description("Status of service")
	TypeName("StatusMedia ")
	ContentType("application/json")
	Attributes(func() {
		Attribute("request_id", UUID, "requst ID")
		Attribute("message", String, "Response Message")
		Attribute("data", Any, "Data body")
		Required("request_id", "message")
	})
})
```

in such case, I defined attribute `data` as `Any` Type, since I will just embed a json inside data, but how can I assign a JSON in `data` ? Thank you in advance.

raphael:
Do you mean a JSON string? is so data could just be a string. If you mean any value then what you have should work just fine. The generated code will produce a field of type `interface{}` and you should be able to assign anything to it that can be JSON marshaled.

perigee:
I tried to use:

```
type StatusData struct {
  Status string `json:"status"`
}

... 

res := &app.StatusMedia {
     RequestID: uuid.NewV4(),
     Message:   "ok",
     Data:         &StatusData{ Status: "status ok"},
}
```

which is failed to compile

perigee:
I got the idea, when the `Attribute`  `data` is not required, it will generate ` * interface {}` instead of `interface{}`, so i failed to compile.

PR #1584, opened 2018-03-12
Prevent Any fields from pointing to *interface{}

xoob:
We fix an issue where non-required fields of kind Any were generated as
a pointer to *interface{}. This broke the Go compiler which failed with
the error "type *interface {} is pointer to interface, not interface".

To prevent this, we teach AttributeDefinition.IsPrimitivePointer that
AnyKind should never be converted to a pointer, since it already is an
interface.

The change also adds a test case.

## Example of the problem

```go
var NullAny = Type("NullAny", func() {
	Attribute("value", Any)
})
```

```go
// nullAny user type.
type nullAny struct {
	Value *interface{} `form:"value,omitempty" json:"value,omitempty" xml:"value,omitempty" bson:"value,omitempty"`
}

// Publicize creates NullAny from nullAny
func (ut *nullAny) Publicize() *NullAny {
	var pub NullAny
	if ut.Value != nil {
		pub.Value = ut.Value
	}
	return &pub
}

// NullAny user type.
type NullAny struct {
	Value *interface{} `form:"value,omitempty" json:"value,omitempty" xml:"value,omitempty" bson:"value,omitempty"`
}
```

### DIFF
--- a/design/definitions.go
+++ b/design/definitions.go
@@ -1033,9 +1033,23 @@ func (a *AttributeDefinition) IsPrimitivePointer(attName string) bool {
 		return false
 	}
 	if att.Type.IsPrimitive() {
-		return !a.IsRequired(attName) && !a.HasDefaultValue(attName) && !a.IsNonZero(attName)
+		return !a.IsRequired(attName) && !a.HasDefaultValue(attName) && !a.IsNonZero(attName) && !a.IsInterface(attName)
 	}
 	return false
+}
+
+// IsInterface returns true if the field generated for the given attribute has
+// an interface type that should not be referenced as a "*interface{}" pointer.
+// The target attribute must be an object.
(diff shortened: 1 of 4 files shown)


xoob:
Fixes #1033

raphael:
Thanks for the detailed report and great PR. Do you think you could backport the fix to the `v1` branch? (cherry pick the merge commit). Thank you!

xoob:
@raphael Yes I can. Just so I get this right, what’s on the master branch right now? Is it v2?

raphael:
Master branch is still v1 at this point, it contains a few breaking changes from v1 so can't be completely merged into it but uses the v1 DSL and code generators. The cherry-pick should be clean. v2 is in its own branch (`v2`).

PR #1590, opened 2018-03-14
fix index out of bounds error

avinashbhashyam-rs:
This is a minor fix to address a bug introduced in https://github.com/goadesign/goa/pull/1585. The old logger's log definition was `func (a *adapter) log(lvl string, keyvals ...interface{})`. so we initialized the vals to be of size n+1 and set the lvl at 0th index https://github.com/goadesign/goa/pull/1585/files#diff-c569e732b7a9e4f2486a8c37f7b06c98L42.


raphael:
Thank you!

PR #1591, opened 2018-03-14
Backport to v1: Prevent "Any" fields from pointing to *interface{}

xoob:
This is a backport of the bugfix from `master` #1584 (86448e9) to the `v1` stable branch, as requested by @raphael.

-----

We fix an issue where non-required fields of kind "Any" were generated
as a pointer to *interface{}. This broke the Go compiler which failed
with the error "type *interface {} is pointer to interface, not
interface".

To prevent this, we teach AttributeDefinition.IsPrimitivePointer that
AnyKind should never be converted to a pointer, since it already is an
interface. The change also adds a test case.

This is a backport of #1584 to the "v1" stable branch.

### DIFF
--- a/design/definitions.go
+++ b/design/definitions.go
@@ -1033,9 +1033,23 @@ func (a *AttributeDefinition) IsPrimitivePointer(attName string) bool {
 		return false
 	}
 	if att.Type.IsPrimitive() {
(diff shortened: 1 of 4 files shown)


raphael:
Thank you!

PR #1592, opened 2018-03-14
Removed [INFO]/[ERROR] from the log

avinashbhashyam-rs:
Missed this change in the previous pr https://github.com/goadesign/goa/pull/1590


raphael:
Thanks

PR #1595, opened 2018-03-15
Rename the DSL MultipartForm to MultipartRequest

tchssk:
Related to #1582.

I had forgotten about the DSL affect to only requests. If we need a DSL that affect to responses, we will name it MultipartResponse. The names should be symmetric.

### DIFF
--- a/design/apidsl/action.go
+++ b/design/apidsl/action.go
@@ -477,13 +477,13 @@ func payload(isOptional bool, p interface{}, dsls ...func()) {
 	}
 }
 
-// MultipartForm can be used in: Action
+// MultipartRequest can be used in: Action
 //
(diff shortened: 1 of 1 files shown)


raphael:
I don't think this applies? Multipart form only applies to requests.

tchssk:
Actually I've never seen a response as multipart/form-data but is it the specification? I read some RFC but I can't have confidence in that. I found some articles about multipart response when I google it.

tchssk:
Maybe there are very few persons who want to use multipart response so we can decide to not support that in goa.

raphael:
You can do multipart responses but not multipart FORM. A form is only for requests.

tchssk:
OK. I'll close this.

PR #1598, opened 2018-03-15
openapi fixes

nitinmohan87:
Required by https://github.com/goadesign/plugins/pull/15

### DIFF
--- a/http/codegen/openapi/openapi_v2_builder.go
+++ b/http/codegen/openapi/openapi_v2_builder.go
@@ -51,7 +51,7 @@ func NewV2(root *httpdesign.RootExpr) (*V2, error) {
 			Contact:        root.Design.API.Contact,
 			License:        root.Design.API.License,
 			Version:        root.Design.API.Version,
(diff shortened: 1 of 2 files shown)


http/codegen/openapi/openapi_v2_builder.go
```diff
@@ -110,6 +110,35 @@ func NewV2(root *httpdesign.RootExpr) (*V2, error) {
 	return s, nil
 }
 
+// ExtensionsFromExpr generates swagger extensions from the given metadata
+// expression.
+func ExtensionsFromExpr(mdata design.MetadataExpr) map[string]interface{} {
```
raphael: why does this have to be a public method?

nitinmohan87: It is used in the security plugin

nitinmohan87: https://github.com/goadesign/plugins/blob/security_middleware/security/generate.go#L163

raphael: got it, thanks!

PR #1602, opened 2018-03-16
Handle interface type in convert

### LINKED ISSUE #1586: ConvertTo and CreateFrom do not seem to handle maps to any
## Expected Behavior

Given an external model type that has a `map[string]interface{}` field and a Goa type definition that has a corresponding attribute with the type `MapOf(String, Any)`, the `ConvertTo` and `CreateFrom` functions should generate converter functions when running `goa gen ...`.

## Actual Behavior

Instead, I get an error about mismatched types.

Given the following file, `goa gen ...` gives an error:

```golang
package thingy

import (
    . "goa.design/goa/http/dsl"
)

type thingy struct {
    Hats map[string]interface{} `bson:"hats"`
}

var Thingy = Type("Thingy", func(){
    ConvertTo(thingy{})
    CreateFrom(thingy{})
    Attribute("hats", MapOf(String, Any))
})

var _ = API("Thingy", func() {
})

var _ = Service("Thingy", func() {
    Method("show", func() {
        Result(Thingy)
    })
})
```

The error that it generates is:

```
exit status 1
"thingy": types don't match: type of <value>.Hats.value is  but type of corresponding attribute is any
```

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Windows

nitinmohan87:
Fixes #1586

### DIFF
--- a/codegen/service/convert.go
+++ b/codegen/service/convert.go
@@ -455,6 +455,8 @@ func isPrimitive(t reflect.Type) bool {
 		fallthrough
 	case reflect.Float64:
 		fallthrough
+	case reflect.Interface:
+		fallthrough
 	case reflect.String:
 		return true
 	case reflect.Slice:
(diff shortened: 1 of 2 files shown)


raphael:
Nice! thanks.

PR #1604, opened 2018-03-16
Use AttributeExpr.Find to find an attribute during Validate

### LINKED ISSUE #1603: Goa gen v2 Payload Extend has no effect
## Expected Behavior
Payload is extended with attributes from the Payload specified in `Extend()`
## Actual Behavior
Payload is not extended, compilation fails if expecting those items in the payload

`goa gen testing/design`

`design.go`:
```go
package design

// API describes the global properties of the API server.
var _ = API("collections", func() {
	Title("collections Service")
	Description("HTTP service for the collections resource")
})

// PayloadA is the payload to be used to extend PayloadB
var PayloadA = Type("PayloadA", func() {
	Attribute("a", String)
})

// PayloadB should be extended by the attributes of PayloadA
// If you comment out the "a" attribute in PayloadB, `goa gen` does not work
var PayloadB = Type("PayloadB", func() {
	Attribute("b", String)
	// Attribute("a", String)
	Extend(PayloadA)
})

	Method("do", func() {
		Payload(PayloadB)
		Result(Int)
		HTTP(func() {
			GET("/service/{a}/{b}")
			Response(StatusOK)
		})
	})
})

```

## Specifications

  - goa Version: v2, latest branch (specifically 8ae8493dac8f54248c5d540a15ea694944ca7144)
  - Platform (Linux/Windows etc.): Mac
  - go version go1.10 darwin/amd64

nitinmohan87:
Fixes #1603

### DIFF
--- a/http/design/endpoint.go
+++ b/http/design/endpoint.go
@@ -416,31 +416,20 @@ func (e *EndpointExpr) Validate() error {
 			}
 		}
 
-		if pObj := design.AsObject(e.MethodExpr.Payload.Type); pObj != nil {
+		if design.IsObject(e.MethodExpr.Payload.Type) {
 			if e.MapQueryParams != nil {
 				if pAttr := *e.MapQueryParams; pAttr == "" {
 					verr.Add(e, "MapParams is set to map entire payload but payload is an object. Payload must be a map.")
-				} else {
-					found := false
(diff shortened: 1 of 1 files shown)


raphael:
Nice, thank you!

PR #1610, opened 2018-03-18
Prevent pointer to '*interface{}' in user types

xoob:
A previous patch in [#1591][1] addressed an issue where optional fields
    of kind "Any" would result in a pointer to interface. We fix an
    additional occurence of this problem in private user types, which follow
    slightly different logic than the public media types previously fixed
    and add a test case for private defs.

    [1]: https://github.com/goadesign/goa/pull/1591


raphael:
Thank you, good catch!

PR #1612, opened 2018-03-19
Add a primitive type File

tchssk:
This pull request adds a new primitive type FIle.

It represents the `file` type of the Swagger specification.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types

> An additional primitive data type "file" is used by the Parameter Object and the Response Object to set the parameter type or the response as being a file.

It can be used with `MultipartForm` and It makes it possible to create a service that can be uploaded a file.

### DIFF
--- a/design/definitions.go
+++ b/design/definitions.go
@@ -1035,7 +1035,7 @@ func (a *AttributeDefinition) IsPrimitivePointer(attName string) bool {
 		return false
 	}
 	if att.Type.IsPrimitive() {
-		return !a.IsRequired(attName) && !a.HasDefaultValue(attName) && !a.IsNonZero(attName) && !a.IsInterface(attName)
+		return (!a.IsRequired(attName) && !a.HasDefaultValue(attName) && !a.IsNonZero(attName) && !a.IsInterface(attName)) || a.IsFile(attName)
 	}
 	return false
 }
@@ -1052,6 +1052,20 @@ func (a *AttributeDefinition) IsInterface(attName string) bool {
 		return false
 	}
 	return att.Type.Kind() == AnyKind
+}
+
+// IsFile returns true if the field generated for the given attribute has
+// an file type that should not be referenced as a "*multipart.File" pointer.
+// The target attribute must be an object.
+func (a *AttributeDefinition) IsFile(attName string) bool {
+	if !a.Type.IsObject() {
+		panic("checking pointer field on non-object") // bug
+	}
+	att := a.Type.ToObject()[attName]
+	if att == nil {
+		return false
+	}
+	return att.Type.Kind() == FileKind
 }
 
 // SetExample sets the custom example. SetExample also handles the case when the user doesn't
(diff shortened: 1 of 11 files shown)


raphael:
Thanks! can you explain a little bit what happens when this type is used in other places (e.g. a request path parameter). Seems like there would need to be a lot of error handling for these cases? Also can you explain a little bit more about the valid use cases for this new type? Thank you.

tchssk:
As you say, it seems we have to add some error handling. I'll improve it.

tchssk:
Here's a valid use case. https://github.com/tchssk/examples/commit/824d42c0c930daadbc804a1e6fb5928da992eae8

raphael:
Also what does mean for gRPC? should this be a HTTP specific DSL?

tchssk:
I forgot to say. This is for v1.

raphael:
Ooh whoops I keep making that mistake - sorry. OK that makes a lot more sense now.

tchssk:
Yes. I'm adding validations for the File type. It seems we should add them in `design/validation.go`

tchssk:
@raphael I added a validation for `File`. Please take a look. https://github.com/goadesign/goa/pull/1612/commits/c1042929f25384f6429e4dabaa167ec2b1b5e661

design/definitions.go
```diff
@@ -1054,6 +1054,20 @@ func (a *AttributeDefinition) IsInterface(attName string) bool {
 	return att.Type.Kind() == AnyKind
 }
 
+// IsFile returns true if the field generated for the given attribute has
```
raphael: Ideally the DSL and design functions should not be defined in terms of generated Go code since the design describes the API "in the absolute". So maybe something like:
```
IsFile returns true if the attribute is of type File or if any its children attributes (if any) is.
```

design/types.go
```diff
@@ -188,6 +190,9 @@ const (
 
 	// Any is the type for an arbitrary JSON value (interface{} in Go).
 	Any = Primitive(AnyKind)
+
+	// File is the type for a file.
```
raphael: Maybe add a note explaining that this type can only be used in a multipart definition.

design/definitions.go
tchssk: I fixed it. https://github.com/goadesign/goa/pull/1612/commits/9cf3d77c29560a7519072228ead52a91b1655960

design/types.go
tchssk: I fixed it. https://github.com/goadesign/goa/pull/1612/commits/50a3a0d56661868b1ed798a1bcdcb80f2ea47850

raphael:
Awesome, thank you!

raphael:
Could you backport this to `v1` as well? thank you!

tchssk:
@raphael I tried to apply this to `v1`, but I got a merge conflict. It seems to be caused by #1583 that has not been backported to v1.

PR #1626, opened 2018-03-23
Fix MuxHandler initialization race

rwool:
Fix race condition that can occur when calling handlers returned from
Controller.MuxHandler where the Service's slice of middleware functions
is written to in a way that is not thread-safe.

The race will only occur when the capacity of the Service's middleware
slice is greater than its length and when the number of Controller
middlewares is more than 0. The functions returned by MuxHandler (that
are all from the same Controller object), if called in parallel, could
cause the race to occur.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ cyclo:
 	fi
 
 test:
-	@ginkgo -r --randomizeAllSpecs --failOnPending --randomizeSuites -skipPackage vendor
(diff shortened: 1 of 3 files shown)


raphael:
Makes sense, thank you for the great PR and good catch! Could you please backport the fix to the `v1` branch as well? (simply cherry-pick the merge commit). Thanks again!

rwool:
Thanks! Made PR for v1 branch at #1627.

PR #1634, opened 2018-03-27
Fix transforming nested maps

### LINKED ISSUE #1619: Using nested Map in design generates code with compilation issues
## Expected Behavior
Generated code to compile successfully

## Actual Behavior
Generated code fails with a compilation error

The following simply snippet can be used to reproduce the issue
```golang
package design

import (
	. "goa.design/goa/http/dsl"
)

var _ = API("dummy", func() {
	Title(`Dummy API`)
})

var SomeRequest = Type("ShowItem", func() {
	Attribute("id", UInt, "id of something.", func() {
		Example(20)
	})
	Attribute("nested_map", MapOf(String, ArrayOf(MapOf(String, Any))), "The nested map")
	Required("id")
})

var _ = Service("dummy", func() {
	Description("Foo")
	HTTP(func() {
		Path("/api/foo")
	})

	Method("Show", func() {
		Description("Show.")
		Payload(SomeRequest)

		HTTP(func() {
			GET("/")
			Response(StatusOK)
		})
	})

})
```

Generate the code and build the generated server type
```
go build gen/http/dummy/server/types.go
```

and this produces the following compilation error
```
gen/http/dummy/server/types.go:39:8: invalid operation: tv[i] (type interface {} does not support indexing)
```

Here is the relevant snippet from the generated code:

```golang
// NewShowShowItem builds a dummy service Show endpoint payload.
func NewShowShowItem(body *ShowRequestBody) *dummysvc.ShowItem {
	v := &dummysvc.ShowItem{
		ID: *body.ID,
	}
	if body.NestedMap != nil {
		v.NestedMap = make(map[string][]map[string]interface{}, len(body.NestedMap))
		for key, val := range body.NestedMap {
			tk := key
			tv := make([]map[string]interface{}, len(val))
			for i, val := range val {
				tv[i] = make(map[string]interface{}, len(val))
				for key, val := range val {
					tk := key
					tv := val
					tv[i][tk] = tv
				}
			}
			v.NestedMap[tk] = tv
		}
	}
	return v
}
```

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Mac

nitinmohan87:
Fixes #1619

### DIFF
--- a/codegen/transform.go
+++ b/codegen/transform.go
@@ -322,12 +322,30 @@ func transformMap(source, target *design.Map, newVar bool, a targs) (string, err
 		"TargetPkg":   a.targetPkg,
 		"Unmarshal":   a.unmarshal,
 		"Scope":       a.scope,
+		"KeyLoopVar":  "",
+		"ValLoopVar":  "",
+	}
+	depth := 0
+	if mapDepth(target.KeyType.Type, &depth); depth > 0 {
+		data["KeyLoopVar"] = string(105 + depth)
+	}
+	depth = 0
+	if mapDepth(target.ElemType.Type, &depth); depth > 0 {
+		data["ValLoopVar"] = string(105 + depth)
 	}
 	var buf bytes.Buffer
 	if err := transformMapT.Execute(&buf, data); err != nil {
 		panic(err) // bug
 	}
 	return buf.String(), nil
+}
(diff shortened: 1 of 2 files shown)


raphael:
Thank you!

PR #1639, opened 2018-03-29
validate different wildcards only if request method is the same

tchssk:
This is backport of #1583 to v1.


raphael:
Thank you for doing this!

PR #1641, opened 2018-03-30
Honor plugin dependencies during code generation

nitinmohan87:
Fixes https://github.com/goadesign/plugins/issues/32

### DIFF
--- a/codegen/plugin.go
+++ b/codegen/plugin.go
@@ -33,17 +33,17 @@ var plugins []*plugin
 // RegisterPlugin adds the plugin to the list of plugins to be invoked with the
 // given command.
 func RegisterPlugin(name string, cmd string, p GenerateFunc) {
-	newP := &plugin{name: name, GenerateFunc: p, cmd: cmd}
+	np := &plugin{name: name, GenerateFunc: p, cmd: cmd}
 	var inserted bool
 	for i, plgn := range plugins {
-		if plgn.last || (!plgn.first && newP.name < plgn.name) {
-			insertPlugin(newP, i)
+		if plgn.last || (!plgn.first && np.name < plgn.name) {
+			plugins = append(plugins[:i], append([]*plugin{np}, plugins[i:]...)...)
 			inserted = true
 			break
 		}
 	}
 	if !inserted {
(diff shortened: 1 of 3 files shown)


codegen/plugin.go
```diff
@@ -13,18 +13,77 @@ type (
...
+// plugins keeps track of the registered plugins sorted by their first/last bools,
+// names, or registration order.
 var plugins []*plugin
 
 // RegisterPlugin adds the plugin to the list of plugins to be invoked with the
 // given command.
-func RegisterPlugin(cmd string, p GenerateFunc) {
-	plugins = append(plugins, &plugin{p, cmd})
+func RegisterPlugin(name string, cmd string, p GenerateFunc) {
+	newP := &plugin{name: name, GenerateFunc: p, cmd: cmd}
```
raphael: style: please use short local variable names (one or two letters), e.g. `p` or `np`. See https://talks.golang.org/2014/names.slide.

codegen/plugin_test.go
```diff
@@ -0,0 +1,110 @@
+package codegen
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestRegisterPlugin(t *testing.T) {
+	var (
+		abcP = &plugin{name: "abc"}
```
raphael: See comment above about variable names.

codegen/plugin.go
```diff
@@ -42,3 +101,10 @@ func RunPlugins(cmd, genpkg string, roots []eval.Root, genfiles []*File) ([]*Fil
 	}
 	return genfiles, nil
 }
+
+// insertPlugin inserts the plugin into the list of plugins at the given index.
+func insertPlugin(p *plugin, index int) {
+	plugins = append(plugins, &plugin{})
```
raphael: Do we really need a function here? seems it should be possible to do:
```go
plugins = append(plugins[:i], append(p, plugins[i:]...)...)
```

nitinmohan87: Ah nice.. Just now found that in https://github.com/golang/go/wiki/SliceTricks

PR #1648, opened 2018-04-04
Fix the encode error

arangamani:
Don't proceed to encoding the response if there was an error and
encoding the error was successful.


raphael:
Thank you!

PR #1658, opened 2018-04-05
Set token in header by default

nitinmohan87:
Changes required by https://github.com/goadesign/plugins/pull/38


raphael:
Thank you!

PR #1668, opened 2018-04-13
Fix variable shadowing issue.

psschroeter:
Debug info was getting written to inner w (http.ResponseWriter) instead
of passed in writer.


raphael:
Whoops, thank you!

Issue #1673, opened 2018-04-17
v2: clientErrors struct issues


psschroeter:
So for the goa.v2 http clients, In addition to the goa.ServiceErrors from the design, it can also return these clientError structs in a number of cases: https://github.com/goadesign/goa/blob/v2/http/client.go#L138

Right now theses a few issues with these errors:
* clientError is a private struct -- It probably should be a publicly exported struct so application code can typecast to it. Or at least have some more interface methods so I can switch on the type like "ErrInvalidType" or "ErrEncodingErr" without reflection.
* No Timeout() and Temporary() functions -- need these so I can return a bad gateway vs. other types of errors.

PR #1674, opened 2018-04-17
Export clientError as ClientError so it can be type cast (#1673)

psschroeter:


http/client.go
```diff
@@ -33,13 +33,14 @@ type (
...
-	// clientError is an error returned by a HTTP service client.
-	clientError struct {
-		message   string
-		service   string
-		method    string
-		temporary bool
-		timeout   bool
+	// ClientError is an error returned by a HTTP service client.
+	ClientError struct {
+		Name        string
```
raphael: Please add a comment for each of the struct fields.

http/client.go
```diff
@@ -129,36 +130,46 @@ func (dd *debugDoer) Fprint(w io.Writer) {
...
 
 // Error builds an error message.
-func (c *clientError) Error() string {
-	return fmt.Sprintf("[%s %s]: %s", c.service, c.method, c.message)
+func (c *ClientError) Error() string {
+	return fmt.Sprintf("[%s %s]: %s", c.Service, c.Method, c.Message)
 }
 
-// ErrInvalidType is the error returned when the wrong type is given to an
+// Temporary implements net.Error#Temporary
```
raphael: Do we really need this? we already have public fields (which should be named `Temporary` and `Timeout`)

raphael:
Thank you!

PR #1675, opened 2018-04-17
Add in validation error type for consistency

psschroeter:
Not necessarily a bug but every other error returned from the http client seems to be a goahttp.ErrXXX error so this helps keeps things consistent.


raphael:
Makes sense, thanks!

PR #1678, opened 2018-04-19
Integrate security

raphael:
This PR moves the security plugin into the core goa design and codegen packages. 

Due to its nature the security plugin source code was closely coupled to the goa code generators which made for brittle and complex algorithms. Having the security DSL built into goa makes the code an order of magnitude simpler and given that security is something any non-trivial API is going to require it seemed like the right trade-off.


raphael:
To update to this version, make sure that:

* The design is updated to remove all imports of the `goa.design/plugins/security` package.
* The service code imports `goa.design/goa/security` instead of `goa.design/plugins/security`.

Issue #1628, opened 2018-03-23
Attribute used for Tag shows up even though its not in View


douglaswth:
## Expected Behavior

I would like to be able to use a `Tag` on an `Attribute` to determine if I want to return `304 Not Modified` based on an Etag; however, that `Attribute` isn't in my view so I don't want it to show up in the actual response.

## Actual Behavior

The `Attribute` shows up in the response anyway.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

TBD

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Windows

PR #1679, opened 2018-04-20
[BREAKING] Fix views result type

nitinmohan87:
[BREAKING]
* The method signature will now return a view string along with the result type and an error if the result type defines more than one view.
* Views package is generated per service to render result types based on view.
* Constructors on service package can be used to convert result type to and from viewed result type
* Server package response types use pointers for all fields.
* Fixes https://github.com/goadesign/goa/issues/1628


codegen/service/service_data.go
```diff
@@ -597,3 +743,242 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+		return false
+	}
+	rt, ok := dt.(*design.ResultTypeExpr)
+	return ok && len(rt.Views) > 1
+}
+
+// buildViewedType builds the type data to generate views and the constructor
+// code to transform result type to its viewed counterpart (and vice versa).
+func buildViewedType(vAtt, at *design.AttributeExpr, scope *codegen.NameScope, viewsPkg string) *ViewedTypeData {
+	resultVar := scope.GoTypeName(at)
```
raphael: `rvar` would be fine here.

codegen/service/service_data.go
```diff
@@ -597,3 +743,242 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+// more than one view and not a collection.
+func isProjectable(dt design.DataType) bool {
+	if design.IsArray(dt) {
+		return false
+	}
+	rt, ok := dt.(*design.ResultTypeExpr)
+	return ok && len(rt.Views) > 1
+}
+
+// buildViewedType builds the type data to generate views and the constructor
```
raphael: .. used to generate ... also should the "project" nomenclature be used here?

codegen/service/service_data.go
```diff
@@ -195,6 +203,73 @@ type (
...
+		ToResult *InitData
+		// ToViewed is the constructor code to convert a result type to a
+		// viewed result type.
+		ToViewed *InitData
+		// Validate is the validations run on the type which is not a
+		// viewed type (i.e., ViewType is not set)
+		Validate string
+	}
+
+	// ViewData contains data about a result type view.
```
raphael: Would be good to use the `project` nomenclature like is used in the existing code. So there is a result type which can be projected to multiple views. If I understand correctly `ViewData` could then be named `ProjectedData` and `Conversion` should be changed to use `Project`. `ToViewed` might also just be `Project` or `ProjectCode` or `Projection` ... you get the idea :) - Maybe I'm misunderstanding the use case though.

codegen/service/service_data.go
```diff
@@ -597,3 +743,242 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+	}
+	vt := design.Dup(ut).(design.UserType)
+	vt.Rename(ut.Name() + "View")
+	vAtt.Type = vt
+	ref := scope.GoTypeRef(vAtt)
+	varName := codegen.Goify(vt.Name(), true)
+	return &ViewedTypeData{
+		ViewType: &UserTypeData{
+			Name:        vt.Name(),
+			Description: fmt.Sprintf("%s is the transformed type of %s type.", varName, resultVar),
```
raphael: Not sure "transformed" is going to make sense to users. Probably need more explanation here.

codegen/service/service_data.go
```diff
@@ -597,3 +743,242 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+		})
+	}
+	return views
+}
+
+// convertResult converts the given result type to/from a viewed result type.
+// If toViewed is true it converts the method result type to viewed
+// result type.
+func convertResult(src, tgt *design.AttributeExpr, srcVar, tgtVar, srcPkg, tgtPkg string, toViewed bool, scope *codegen.NameScope) *InitData {
+	srcRT, ok := src.Type.(*design.ResultTypeExpr)
```
raphael: no mix case in local variable names please.

docs/FAQ.md
```diff
@@ -44,3 +44,37 @@ or client side to unmarshal a response) the default value is used to set the
 value of missing fields. Note that if the attribute is required then the
 generated code returns an error if the corresponding field is missing. So this
 only applies for non required attributes with default values.
+
+# How are views for a result type computed?
+
+Views can be defined on a result type. If a method returns a result type with
```
raphael: Would be good to start with the user facing consequences of defining multiple views, that is the fact that the service method returns an extra value in this case which corresponds to the name of the view that should be rendered in the response. Also would be good to have some examples of design and corresponding generated code.

docs/FAQ.md
```diff
@@ -44,3 +44,37 @@ or client side to unmarshal a response) the default value is used to set the
...
+multiple views then a views package is generated at the service level which
+projects the result type as per the attributes defined in the view.
+Constructors are generated in the service package to convert a result type
+to a viewed result type and vice versa.
+
+The views package defines a type with name as the result type name suffixed
```
raphael: For each result type with multiple views the `views` package defines a type with identical field names and types but using pointers for all fields so that view specific validation logic may be generated.

docs/FAQ.md
```diff
@@ -44,3 +44,37 @@ or client side to unmarshal a response) the default value is used to set the
...
+with "View". This type has all the fields defined on the result type as
+pointers. The views package also has a type with same name as the result type.
+This type contains the viewed type mentioned above along with a required "view"
+attribute holding the view name which is used to validate that the viewed
+result type contains all the attributes as defined in the view. The type also
+defines "As<View>" functions which projects the viewed result type with only
+the attributes defined in the view.
+
+The signature for the method that returns a result type with multiple views
```
raphael: Use the present tense: `will return` -> `returns`

docs/FAQ.md
```diff
@@ -44,3 +44,37 @@ or client side to unmarshal a response) the default value is used to set the
...
+will return the result type, the view to render the result type, and an error.
+The generated endpoint function will use this view to create a validated
+viewed result type with only the attributes defined in the view.
+
+The server side response marshaling code will marshal the viewed result type
+returned by the endpoint into a will server type omitting any nil attributes.
```
raphael: typo also use present tense.

http/codegen/server.go
```diff
@@ -894,9 +895,14 @@ const responseEncoderT = `{{ printf "%s returns an encoder for responses returne
 func {{ .ResponseEncoder }}(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
 	return func(ctx context.Context, w http.ResponseWriter, v interface{}) error {
 
-	{{- if .Result.Ref }}
+	{{- if and .Result.Ref .ResponseBodyExists }}
+		{{- if .Method.ViewedResult }}
+		vRes := v.({{ .Method.ViewedResult.FullRef }})
```
raphael: `vres`

http/codegen/client.go
```diff
@@ -468,7 +471,25 @@ const singleResponseT = ` {{- if .ClientBody }}
 				return nil, goahttp.ErrValidationError("{{ $.ServiceName }}", "{{ $.Method.Name }}", err)
 			}
 		{{- end }}
-	{{ end }}
+	{{- else if and $.Method.ViewedResult (not .IsError) }}
+		var (
+			vRes {{ $.Method.ViewedResult.FullRef }}
```
raphael: `vres`

codegen/service/service.go
```diff
@@ -152,7 +152,7 @@ type Service interface {
 			{{- end }}
 		{{- end }}
 	{{- end }}
-	{{ .VarName }}(context.Context{{ if .Payload }}, {{ .PayloadRef }}{{ end }}) {{ if .Result }}({{ .ResultRef }}, {{ if .ViewedResult }}string, {{ end }}error){{ else }}error{{ end }}
+	{{ .VarName }}(context.Context{{ if .Payload }}, {{ .PayloadRef }}{{ end }}) ({{ if .Result }}res {{ .ResultRef }}, {{ if .ViewedResult }}view string, {{ end }}{{ end }}err error)
```
raphael: Wonder why named return values for the interface methods?

http/codegen/service_data.go
```diff
@@ -523,15 +523,19 @@ func (svc *ServiceData) Endpoint(name string) *EndpointData {
 	return nil
 }
 
-// ResponseBodyExists returns true if at least one of the responses has a body.
-func (e *EndpointData) ResponseBodyExists() bool {
-	if e.Result != nil {
-		return true
+// NeedServerResponse returns true if server response has a body or a header.
+// It is used in initializing the result in the server response encoding.
```
raphael: It is used when

codegen/service/endpoint.go
```diff
@@ -258,7 +256,25 @@ func New{{ .VarName }}Endpoint(s {{ .ServiceVarName}}{{ range .Schemes }}, auth{
 			return nil, err
 		}
 {{- end }}
-{{- if .ResultRef }}
+{{- if .ViewedResult }}
+		res, view, err := s.{{ .VarName }}(ctx{{ if .PayloadRef }}, p{{ end }})
+		if err != nil {
+			return nil, err
+		}
+		vres := {{ .ViewedResult.ConvertToViewed.VarName }}(res)
```
raphael: Looking at this now, it seems strange that we would create the viewed type to then project it right away. Why wouldn't we have view specific constructors instead so that we don't have to initialize fields that we nullify in the next call? Could we replace the constructor above with one constructor per view (and adjust the comment explaining how that works).

nitinmohan87: Sure.. I will change that as well.

codegen/service/service_data.go
```diff
@@ -195,6 +203,69 @@ type (
 		// In indicates the request element that holds the credential.
 		In string
 	}
+
+	// ProjectedTypeData contains the data to generate a user type that can
+	// be projected based on a view. The generated type uses pointers for all
+	// fields so that view specific validation logic may be implemented.
+	// A projected type will be generated for every user type found in a
```
raphael: is generated

raphael: is generated

codegen/service/service_data.go
```diff
@@ -288,6 +364,10 @@ func (d ServicesData) analyze(service *design.ServiceExpr) *Data {
 				ratt = ut.Attribute()
 			}
 			types = append(types, collectTypes(ratt, seen, scope)...)
+			if isProjectable(m.Result.Type) {
+				_, pts := collectProjectedTypes(m.Result, seenProj, scope, viewspkg)
+				projTypes = append(projTypes, pts...)
```
raphael: Could that produce duplicates in `projTypes`? Does it matter?

codegen/service/service_data.go
```diff
@@ -597,3 +683,252 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+				Ref:         scope.GoTypeRef(projected),
+				Type:        pt,
+			},
+			Validate: codegen.RecursiveValidationCode(valattr, false, true, false, "result"),
+		}
+		types = append(types, p)
+		seen[pt.Name()] = p
+		return types
+	}
+	// pt is a result type with multiple views. Generate two types in the
```
raphael: unfinished comment

codegen/service/service_data.go
```diff
@@ -597,3 +683,252 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+		if v, ok := n.Attribute.Metadata["view"]; ok {
+			if attr.Metadata == nil {
+				attr.Metadata = design.MetadataExpr{}
+			}
+			attr.Metadata["view"] = v
+		}
+		obj.Set(n.Name, attr)
+	}
+	code, helpers := buildConstructorCode(res, tgt, "res", "t", "", viewspkg, true, scope)
+	code += fmt.Sprintf("\nreturn &%s.%s{\n%s: t,\nView: %q,\n}", viewspkg, rname, vresrt.TypeName, view.Name)
```
raphael: Would be nice to use templates instead of building the code inline here and below. Would make it a lot easier to grok.

codegen/service/testdata/endpoint_code.go
```diff
@@ -0,0 +1,148 @@
...
+			return nil, err
+		}
+		var vres *withresultmultipleviewsviews.Viewtype
+		switch view {
+		case "default":
+			vres = NewViewtypeDefault(res)
+		case "tiny":
+			vres = NewViewtypeTiny(res)
+		default:
+			return nil, fmt.Errorf("unknown view %s", view)
```
raphael: use `%q`

codegen/service/testdata/views_code.go
```diff
@@ -0,0 +1,239 @@
...
+	A *string
+	B *string
+}
+
+// ResultType is the viewed result type that projects ResultTypeView based on a
+// view.
+type ResultType struct {
+	*ResultTypeView
+	// View to render
+	View string
```
raphael: I wonder if we should not use a named field to hold `ResultTypeView`. if that type has ` View` field then this won't work.

docs/FAQ.md
```diff
@@ -44,3 +44,29 @@ or client side to unmarshal a response) the default value is used to set the
...
+The server side response marshaling code marshals the viewed result type
+returned by the endpoint into a server type omitting any nil attributes.
+The view used to render the result type is passed to the client in "Goa-View"
+header.
+
+The client side response unmarshaling code unmarshals the response into the
+viewed result type and sets the view attribute from the Goa-View header.
+It validates the attributes in the viewed result type as defined by the view
+and converts the viewed result type into the service result type using the
+appropriate constructor.
```
raphael: Great explanation!

codegen/service/service_data.go
nitinmohan87: it won't produce duplicates since `seenProj` will stop it from adding it to the return value.

codegen/service/service_data.go
```diff
@@ -597,3 +699,301 @@ func requirements(m *design.MethodExpr) []*design.SecurityExpr {
...
+	}
+	data["Attributes"] = attrs
+	var buf bytes.Buffer
+	if err = initResultTypeCodeTmpl.Execute(&buf, data); err != nil {
+		panic(err) // bug
+	}
+	return buf.String(), helpers
+}
+
+const initResultTypeCodeT = `{{- if not .View }}
```
raphael: I don't follow the code below, looks like we initialize `{{ .TargetVar }}` but then use `{{ .Target }}`?

codegen/service/testdata/service_code.go
```diff
@@ -0,0 +1,381 @@
...
+		Message: err.Error(),
+	}
+}
+`
+
+const MultipleMethodsResultMultipleViews = `
+// Service is the MultipleMethodsResultMultipleViews service interface.
+type Service interface {
+	// A implements A.
+	// A must return one of the following views
```
raphael: Maybe the comment could be:
```
The return value "view" must have one of the following values:
```
Also would be nice to quote the view names:
```
* "tiny"
* "default"
```

codegen/service/testdata/views_code.go
```diff
@@ -0,0 +1,303 @@
...
+type RT3 struct {
+	// Type to project
+	Projected *RT3View
+	// View to render
+	View string
+}
+
+// NewRT initializes RT viewed result type from RTView projected type and a
+// view.
+func NewRT(p *RTView, view string) *RT {
```
raphael: I am not sure these are really needed? couldn't the decode code simply create the struct inline?

http/codegen/service_data.go
```diff
@@ -503,6 +497,23 @@ func (svc *ServiceData) Endpoint(name string) *EndpointData {
 	return nil
 }
 
+// NeedServerResponse returns true if server response has a body or a header.
+// It is used in initializing the result in the server response encoding.
```
raphael: It is used when

http/codegen/service_data.go
```diff
@@ -1105,11 +1139,24 @@ func buildResultData(svc *service.Data, s *httpdesign.ServiceExpr, e *httpdesign
 				}
 				notag = i
 			}
+			// Project response body if body type is a result type and if the
+			// endpoint is not a result type with multiple views
```
raphael: ... otherwise the response body is used to initialize a viewed result type that is then validated and transformed into a result type (??)

http/codegen/service_data.go
```diff
@@ -1444,7 +1495,7 @@ func buildBodyType(svc *service.Data, s *httpdesign.ServiceExpr, e *httpdesign.E
 		// Otherwise when generating types that are used to unmarshal
 		// use pointer fields regardless of whether the attribute is
 		// required or has a default value.
-		marshaled = !svr && req || svr && !req
+		marshaled = !svr && req
```
raphael: Is `marshaled` still useful like this? would it be better to leave it (would be used on line 1538) and then add a new `needValidation` that would be used on line 1521?

http/codegen/service_data.go
```diff
@@ -1467,7 +1518,7 @@ func buildBodyType(svc *service.Data, s *httpdesign.ServiceExpr, e *httpdesign.E
 			}
 			desc = fmt.Sprintf("%s is the type of the %q service %q endpoint HTTP %s body.",
 				varname, svc.Name, e.Name(), ctx)
-			if !marshaled {
+			if svr && req || (!svr && !req && md.ViewedResult == nil) {
```
raphael: What happens to validations other than required? are they still being checked somewhere?

http/codegen/service_data.go
```diff
@@ -1502,20 +1553,31 @@ func buildBodyType(svc *service.Data, s *httpdesign.ServiceExpr, e *httpdesign.E
...
-			name, ctx, rctx, e.Name(), s.Name())
+			name, ctx, rctx, e.Name(), svc.Name)
 
-		// If design uses Body("name") syntax then need to use payload
-		// attribute to transform.
 		srcType := att.Type
+		tgtType := body.Type
+		if svr && !req {
+			// In server response type all the attributes are pointers. So we need to
+			// remove the required validations from the body to transform.
```
raphael: Could we do it so that `body.Type` is correctly initialized in that case? or would that break something else? if so please explain in comment.

http/codegen/service_data.go
```diff
@@ -1639,21 +1701,32 @@ func extractQueryParams(a *design.MappedAttributeExpr, serviceType *design.Attri
...
 		fieldName := codegen.Goify(name, true)
 		if !design.IsObject(serviceType.Type) {
 			fieldName = "" // result is initialized directly from header
 		}
+		if viewed {
+			// In a viewed result type the attributes are always pointers.
+			required = false
+			// No need to generate validation code since the validation is performed
+			// by the viewed result type.
+			vcode = ""
```
raphael: Are validations other than required run somewhere else?

http/codegen/service_data.go
```diff
@@ -1710,7 +1783,7 @@ func collectUserTypes(dt design.DataType, cb func(design.UserType), seen ...map[
 	}
 }
 
-func attributeTypeData(ut design.UserType, req, ptr, server bool, scope *codegen.NameScope, rd *ServiceData) *TypeData {
+func attributeTypeData(ut design.UserType, req, ptr, server, mustValidate bool, scope *codegen.NameScope, rd *ServiceData) *TypeData {
```
raphael: Could we make sure the type has no required validation so we wouldn't need that extra argument? or do we need it for other reasons?

http/codegen/testdata/result_encode_functions.go
```diff
@@ -625,6 +625,40 @@ func EncodeMethodBodyUserResponse(encoder func(context.Context, http.ResponseWri
 }
 `
 
+var ResultBodyMultipleViewsEncodeCode = `// EncodeMethodBodyMultipleViewResponse returns an encoder for responses
+// returned by the ServiceBodyMultipleView MethodBodyMultipleView endpoint.
+func EncodeMethodBodyMultipleViewResponse(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
+	return func(ctx context.Context, w http.ResponseWriter, v interface{}) error {
+		vres := v.(*servicebodymultipleviewviews.Resulttypemultipleviews)
```
raphael: Wondering why the generated type name isn't properly camelized?

http/codegen/service_data.go
nitinmohan87: The validations generated in the views package will validate everything along with required.

http/codegen/service_data.go
nitinmohan87: Yup.. the validation function in the views package runs all validations.

codegen/service/service_data.go
nitinmohan87: `TargetVar` refers to struct fields of type result type. `Target` is the final result type/viewed result type returned.

http/codegen/service_data.go
nitinmohan87: Well, the validation code can generate validation for other things in the client and server types. But for an endpoint using a result type with multiple views, we don't use the validation code in the server and client packages, but use the validation logic generated in the views package. So this `bool` helps in not generating the validation code for them in the server and client packages.

http/codegen/testdata/result_encode_functions.go
nitinmohan87: That's because I set the result like this `ResultType("ResultTypeMultipleViews", func() {` . So the result type identifier (all lower-case after calling mime.ParseMediaType) is being used as the type name.

codegen/service/service_data.go
nitinmohan87: Modified the code to have better clarity.

codegen/service/testdata/views_code.go
nitinmohan87: Now the projected type is held by the field named "Projected".

codegen/service/service.go
```diff
@@ -143,12 +143,12 @@ type Service interface {
 {{- range .Methods }}
 	{{ comment .Description }}
 	{{- if .ViewedResult }}
-		{{ printf "%s must return one of the following views" .VarName | comment }}
+		{{ comment "The return value must have one of the following views" }}
```
raphael: There is more than one return value and it may be good to be explicit so maybe something like:
```
The "view" return value must have one of the following values:
```

raphael:
Looks good! thank you for all the work.

PR #1682, opened 2018-04-21
added ability to suppress sensitive headers from log. closes #1676

### LINKED ISSUE #1676: [enhancement] sensitive headers are being logged in middleware/log_request.go
First, thanks for the awesome work. Long-time user. Introduced it to our team a few months ago. Whole team loves it and we pretty much moved all of our api's over to goa.

Just have a quick suggestion. Wanted to pass it by the you / the community. Would be happy to open a PR.

We use `service.Use(middleware.LogRequest(true))` and we noticed some sensitive headers were being logged (basic auth, access tokens, etc).

A simple solution (but not one that's backwards compatible) could be to change the function definition to
` func LogRequest(verbose bool, suppressHeadersLower map[string]bool) goa.Middleware { `.  Then a minor change starting on line 65:

```
if len(r.Header) > 0 {
    logCtx := make([]interface{}, 2*len(r.Header))
    i := 0
    for k, v := range r.Header {
        logCtx[i] = k

        if suppressHeadersLower[strings.ToLower(k)] {
            logCtx[i+1] = "<secret>"
        } else {
            logCtx[i+1] = interface{}(strings.Join(v, ", "))
        }

        i = i + 2
    }
    goa.LogInfo(ctx, "headers", logCtx...)
}
```

adam-hanna:
Minor change to `middleware/log_request.go` that allows the user to suppress sensitive headers from the log. Closes #1676.

### DIFF
--- a/middleware/log_request.go
+++ b/middleware/log_request.go
@@ -19,7 +19,15 @@ import (
 // This middleware is aware of the RequestID middleware and if registered after it leverages the
 // request ID for logging.
 // If verbose is true then the middlware logs the request and response bodies.
-func LogRequest(verbose bool) goa.Middleware {
+func LogRequest(verbose bool, sensitiveHeaders ...string) goa.Middleware {
+	var suppressed map[string]struct{}
+	if len(sensitiveHeaders) > 0 {
+		suppressed = make(map[string]struct{}, len(sensitiveHeaders))
+		for _, sh := range sensitiveHeaders {
+			suppressed[strings.ToLower(sh)] = struct{}{}
+		}
+	}
+
 	return func(h goa.Handler) goa.Handler {
 		return func(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
 			reqID := ctx.Value(reqIDKey)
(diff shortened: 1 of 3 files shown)


middleware/log_request.go
```diff
@@ -33,11 +33,22 @@ func LogRequest(verbose bool) goa.Middleware {
 				"ctrl", goa.ContextController(ctx), "action", goa.ContextAction(ctx))
 			if verbose {
 				if len(r.Header) > 0 {
+					var suppressed map[string]struct{}
```
raphael: Please move this block outside of the goa handler definition so that the map isn't recreated on every request. Concurrent map access is safe.

raphael:
Looks great! left a review comment, I'll merge once it's addressed. Thank you.

adam-hanna:
Whoops, how embarrassing! Good catch!

raphael:
Great! do you think you could backport this PR to the `v1` branch? simply cherry-pick the merge commit and create a new PR targeting `v1`. Thank you!

adam-hanna:
Will do. Thanks!

raphael:
I'm not sure why but the merge broke the build, would appreciate if you could take a quick look, thanks!
https://travis-ci.org/goadesign/goa/builds/370258454

adam-hanna:
Very strange. looking into it now.

Looks like it broke, here: `make: mdc: Command not found`

raphael:
Sorry, there is another issue with the `golint` import being broken (https://github.com/golang/lint/issues/397) so I added a workaround (https://github.com/goadesign/goa/commit/6e802dbfdc5452e8fc560638a15cc87ae63929a5) which re-triggered a build hiding the initial error (it was one of tests added in this PR failing). I'll keep an eye on it and will ping here if it ever reproduces. Thanks for checking.

PR #1700, opened 2018-05-02
Add Scan and Value methods for UUID

mattrco:
These are included in satori/go.uuid and must be added here due
to restrictions on adding new methods to non-local types.

The other alternative is for the user to declare a type alias,
but this will solve the issue for all goa users.

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -5,7 +5,12 @@
 
 package uuid
 
-import "github.com/satori/go.uuid"
+import (
+	"database/sql/driver"
+	"fmt"
+
+	"github.com/satori/go.uuid"
+)
 
 // FromString Wrapper around the real FromString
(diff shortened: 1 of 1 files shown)


raphael:
Thanks for the PR! Can you explain the use case a bit more, when are these methods needed? Adding a dependency on the `sql` package is not very attractive...

mattrco:
Sure - we're using `gorma` to generate storage models, so types that we use in our design must be able to be mapped to database columns. The `Scanner` and `Value` methods are necessary for this.

My understanding is that we've only wrapped `satori/go.uuid` for API compatibility / gopher.js stuff, but unfortunately that means we can't use the original package's database mapping.

mattrco:
Thanks! Will submit the additional PR.

PR #1723, opened 2018-05-23
use_x_request_id

smessier:
added options for RequestID middleware
same default behavior but with options to use incoming "X-Request-Id" header
added unit tests


raphael:
This looks great thank you. Would you mind putting a couple of examples for usage of the options in the middleware function header comment?

raphael:
Thanks!

PR #1724, opened 2018-05-23
Escape '%' to avoid "%A(MISSING)" strings in log

stooke:
URLs with query strings often have those queries escaped with '%'.
The underlying logger implementation eventually calls log.Info(msg), which will parse the message as a format string, and then not find any values to format.


stooke:
I agree, and apparently it's some combination of Goa and other third party packages that causes the issue.  I'm closing this PR.  Sorry to waste your time.

raphael:
Thanks for your understanding!

Issue #1716, opened 2018-05-19
result type containing array of sub-type generated as uncompilable type mismatch


smessier:
The following design generates code that does not compile:
`$GOPATH/foo/design.go`
```package design

import (
	. "goa.design/goa/http/design"
	. "goa.design/goa/http/dsl"
)

var _ = API("foo", func() {
	HTTP(func() {
		Path("/foo")
	})
})

var Point = Type("Point", func() {
	Attribute("x", Int)
	Attribute("y", Int)
	Required("x", "y")
})

var Bar = ResultType("application/vnd.foo.bar+json", func() {
	TypeName("Bar")

	Attributes(func() {
		Attribute("id", String, "Unique identifier.")
		Attribute("href", String, "Self-referential hyperlink reference (HRef).")
		Attribute("source", Point, "The targets")
		Attribute("targets", ArrayOf(Point), "The targets")
		Required("id", "href", "source", "targets")
	})
	View("default", func() {
		Attribute("id")
		Attribute("href")
		Attribute("source")
		Attribute("targets")
	})
})

var _ = Service("bar", func() {
	Method("index", func() {
		NoSecurity()
		Error("not_found")
		Payload(Empty)
		Result(CollectionOf(Bar))
		HTTP(func() {
			GET("/bars")
			Response(StatusOK)
		})
	})
})
```
this results in 'foo/gen/http/bar/server/types.go` which on line 40-61 has the following:
```
func NewIndexResponseBody(res bar.BarCollection) IndexResponseBody {
	body := make([]*BarResponseBody, len(res))
	for i, val := range res {
		body[i] = &BarResponseBody{
			ID:   &val.ID,
			Href: &val.Href,
		}
		if val.Source != nil {
			body[i].Source = marshalPointToPointResponseBody(val.Source)
		}
		if val.Targets != nil {
			body[i].Targets = make([]*PointResponseBody, len(val.Targets))
			for j, val := range val.Targets {
				body[i].Targets[j] = &PointResponseBody{
					X: val.X,  // X is a non-pointer being assigned to pointer X:
					Y: val.Y,  // so is Y
				}
			}
		}
	}
	return body
}
```

PR #1725, opened 2018-05-23
Simplify views

nitinmohan87:
* Viewed result type is generated only for method result types with multiple views. Other result types which a method result type may contain are treated as projected types.
* Validate functions are defined on projected types in the views package. If the projected type is a result type with multiple views, then a validate function is defined for each view.
* Remove the need for unsetting validations just to generate a pointer type.
* Fixes https://github.com/goadesign/goa/issues/1716


codegen/scope.go
```diff
@@ -67,22 +67,22 @@ done:
 
 // GoTypeDef returns the Go code that defines a Go type which matches the data
 // structure definition (the part that comes after `type foo`).
-func (s *NameScope) GoTypeDef(att *design.AttributeExpr, useDefault bool) string {
+func (s *NameScope) GoTypeDef(att *design.AttributeExpr, useDefault bool, pointer bool) string {
```
raphael: Please document the `pointer` argument in the method header comment.

raphael: Would it make things easier if there was two methods `GoTypeDef` and `GoTypeDefPointer` instead of an argument? Maybe it would have helped but now that the change is made it doesn't help anymore.

nitinmohan87: Well.. I am ok having it either way. Right now, adding the `pointer` parameter is not complicating the existing code too much. I don't see us adding much more to this code.

Issue #1718, opened 2018-05-20
Type with Any throws error


kumarsiva07:
![image](https://user-images.githubusercontent.com/7725174/40277153-2d93679a-5c38-11e8-86a8-0abb02401313.png)
I just added one attribute with Any
`Attribute("any", Any, "Winery that produces wine")`

```
kumarsiva:cellar sivakumarm$ go run cmd/cellarsvc/main.go 
# goa.design/goa/examples/cellar/gen/storage
gen/storage/service.go:187:13: invalid indirect of vres.Projected.Any (type interface {})
# goa.design/goa/examples/cellar/gen/http/sommelier/server
gen/http/sommelier/server/types.go:91:7: cannot use val.Any (type interface {}) as type *interface {} in field value:
	*interface {} is pointer to interface, not interface
```

kumarsiva07:
`Attribute("winerys", ArrayOf(Winery), "Winery that produces wine")`

```
kumarsiva:cellar sivakumarm$ go run cmd/cellarsvc/main.go 
# goa.design/goa/examples/cellar/gen/storage
gen/storage/service.go:201:18: val.Name undefined (type *views.Winery has no field or method Name)
gen/storage/service.go:202:18: val.Region undefined (type *views.Winery has no field or method Region)
gen/storage/service.go:203:18: val.Country undefined (type *views.Winery has no field or method Country)
gen/storage/service.go:204:17: val.URL undefined (type *views.Winery has no field or method URL)
gen/storage/service.go:250:9: unknown field 'Name' in struct literal of type views.Winery
gen/storage/service.go:251:11: unknown field 'Region' in struct literal of type views.Winery
gen/storage/service.go:252:12: unknown field 'Country' in struct literal of type views.Winery
gen/storage/service.go:253:8: unknown field 'URL' in struct literal of type views.Winery
# goa.design/goa/examples/cellar/gen/http/sommelier/server
gen/http/sommelier/server/types.go:108:10: cannot use val.Name (type string) as type *string in field value
gen/http/sommelier/server/types.go:109:12: cannot use val.Region (type string) as type *string in field value
gen/http/sommelier/server/types.go:110:13: cannot use val.Country (type string) as type *string in field value
```
Getting error for ArrayOf data type also. @nitinmohan87

kumarsiva07:
Still I am getting the below error.
`app/gen/vtable/views/view.go:149:16: v.Validate undefined (type *StoredVColumnView has no field or method Validate)
`

Design-
[...]

Issue #1712, opened 2018-05-10
Methods don't inherit service headers correctly


andrewalonis:
Headers defined at service level don't trickle down to methods appropriately.

Defining the header at the service level doesn't work. POST requests (anything requiring a payload?) expect/require the payload attribute to appear in the body, and GET requests (no payload) don't expect/require it at all.
```var _ = Service(“Front” func() {
  HTTP(func() {
    Path("/")
    Header(“api_version:Api-Version)
  }
  Method(“show”, func() {
    Payload(func() {
      Attribute(“api_version", String)
      Required("api_version")
    }
    HTTP(func() {}
      GET("/")
    }
  }
}```
However, moving it into each method definition works as intended.
```var _ = Service(“Front” func() {
  HTTP(func() {
    Path("/")
  }
  Method(“show”, func() {
    Payload(func() {
      Attribute(“api_version", String)
      Required("api_version")
    }
    HTTP(func() {}
      GET("/")
      Header(“api_version:Api-Version)
    }
  }
}```

PR #1729, opened 2018-05-25
Headers params service level

nitinmohan87:
* Header attribute in HTTP endpoint and response expression are no longer initialized in Finalize. We extract the headers correctly in the codegen phase.
* Header and params defined in the service level are added to the endpoint level headers and params. (Fixes https://github.com/goadesign/goa/issues/1712)
* Create unique local variable names in codegen.
* Add IsPointer field to AttributeExpr to force pointer generation.
* Fixes https://github.com/goadesign/goa/issues/1718


codegen/scope.go
```diff
@@ -65,28 +65,50 @@ done:
 	return name
 }
 
+// Unique returns a unique name for the given name. If given name not unique
```
raphael: Might be better to specify in `HashedUnique` the fact that requesting `HashedUnique` multiple times for keys returning the same hash return the same value. And not mention anything here about it since the user of the code doesn't know what the "named and count hash" is.

http/design/body_types.go
```diff
@@ -187,6 +189,36 @@ func buildResponseBody(name string, attr *design.AttributeExpr, resp *HTTPRespon
 	return &design.AttributeExpr{Type: nmt, Validation: userType.Validation}
 }
 
+func setForcePointer(att *design.AttributeExpr, seen ...map[string]struct{}) {
```
raphael: Is it really necessary to set the flag on all attributes? seems like it's only needed if the attribute has a primitive type? (so an additional case for primitive types that sets the flag might be enough?)

codegen/service/service_data.go
```diff
@@ -738,6 +738,7 @@ func collectProjectedTypesR(projected, att *design.AttributeExpr, seen map[strin
 	collect := func(projected, att *design.AttributeExpr) []*ProjectedTypeData {
 		return collectProjectedTypesR(projected, att, seen, scope, viewspkg)
 	}
+	projected.ForcePointer = true
```
raphael: We don't need to recurse here? ^

nitinmohan87: This is already in a recursive logic which goes through all attributes to collect projected types.

http/design/body_types.go
nitinmohan87: Fixed.

Issue #1722, opened 2018-05-21
query params do not support default values


smessier:
The following design attempts to use a default value for one of the query params:
```
package design

import (
	. "goa.design/goa/http/design"
	. "goa.design/goa/http/dsl"
)

var _ = API("foo", func() {
	HTTP(func() {
		Path("/foo")
	})
})

var _ = Service("bar", func() {
	Method("puts", func() {
		NoSecurity()
		Payload(func() {
			Attribute("required", String)
			Attribute("optional", String, func() {
				Default("default value")
			})
			Required("required")
		})
		Result(Empty)
		HTTP(func() {
			PUT("/puts")
			Param("required")
			Param("optional")
			Response(StatusNoContent)
		})
	})
})
```
this results in 'foo/gen/http/bar/server/encode_decode.go` which on line 29-51 has the following:
```
func DecodePutsRequest(mux goahttp.Muxer, decoder func(*http.Request) goahttp.Decoder) func(*http.Request) (interface{}, error) {
	return func(r *http.Request) (interface{}, error) {
		var (
			required string
			optional string
			err      error
		)
		required = r.URL.Query().Get("required")
		if required == "" {
			err = goa.MergeErrors(err, goa.MissingFieldError("required", "query string"))
		}
		optionalRaw := r.URL.Query().Get("optional")
		if optionalRaw != "" {
			optional = optionalRaw
		}  // <<< else missing logic for the default value?
		if err != nil {
			return nil, err
		}
		payload := NewPutsPutsPayload(required, optional)

		return payload, nil
	}
}
```

PR #1730, opened 2018-05-25
Set default values during server request decode

### LINKED ISSUE #1661: Attribute with Default value generates incorrect CLI build payload
Attribute with `Default` value set generates incorrect BuildXXPayload code in HTTP CLI which causes client build to fail.
```
$ go build ./cmd/dummycli/
gen/http/dummy/client/cli.go:20:7: cannot use &dummyShowID (type *string) as type string in assignment
```

## Expected Behavior
Code generates and compiles.

## Actual Behavior
Code generates but compilation fails with error above.

```
var _ = Service("dummy", func() {
  Method("Show", func() {
    Payload(func() {
      Attribute("id", String, func() {
        Default("")
      })
    })

    HTTP(func() {
      GET("/")
      Param("id")
      Response(StatusOK)
    })
  })
})
```

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Linux

nitinmohan87:
Fixes #1661 and #1722 
* Path params are treated same as query parameters (no special logic to determine if path params should be pointers)

### DIFF
--- a/examples/cellar/gen/http/storage/server/encode_decode.go
+++ b/examples/cellar/gen/http/storage/server/encode_decode.go
@@ -56,7 +56,9 @@ func DecodeShowRequest(mux goahttp.Muxer, decoder func(*http.Request) goahttp.De
 
 			params = mux.Vars(r)
 		)
-		id = params["id"]
+		{
+			id = params["id"]
+		}
 		viewRaw := r.URL.Query().Get("view")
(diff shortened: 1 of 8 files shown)


raphael:
The generate type fields corresponding to path parameters should never use pointers as they can never be nil. It would be good to tweak these changes so that stays the case. Thank you!

PR #1727, opened 2018-05-25
add a yaml tag to generated types fields

ThibaultLaurens:

PR #1733, opened 2018-05-28
add a yaml tag to generated types fields (#1727)

ThibaultLaurens:
* add a yaml tag to generated types fields

* yaml tag: add missing args

* update unit tests

* yaml tags, yaml tags everywhere

* unit tests: remove white space


raphael:
Thank you!

PR #1738, opened 2018-06-07
Validate viewed result only if client body exists

nitinmohan87:
Fixes an issue in client response decode where it tries to validate the viewed result type even if the response body is empty.


raphael:
Thanks!

PR #1742, opened 2018-06-11
Fix for v1 multipart/form-data Swagger semantic error

MarkSonghurst:
Fixes #1741 in accordance with https://swagger.io/docs/specification/2-0/file-upload/

PR #1744, opened 2018-06-11
Fix for v1 multipart/form-data Swagger semantic error (#1742)

MarkSonghurst:
* Added Action specific Consumes support for multipart

* Added unit test for multipart


raphael:
Thank you!

PR #1754, opened 2018-06-13
Implement ID for user type

nitinmohan87:
* Implements ID function for user type used in breaking recursion
* Removed unnecessary attribute duping in few places
* All user types in views package will be suffixed with View to avoid collision in helper method generation in http package.
* Result type with single view will also generate views package.


docs/FAQ.md
```diff
@@ -56,17 +56,15 @@ only applies for non required attributes with default values.
 
 # How are views for a result type computed?
 
-Views can be defined on a result type. If a method returns a result type with
-multiple views then
-* the service method returns an extra view along with the result and error
+Views can be defined on a result type. If a method returns a result type
```
raphael: missing a "then"

docs/FAQ.md
```diff
@@ -56,17 +56,15 @@ only applies for non required attributes with default values.
...
+* the service method returns an extra view along with the result and error if
```
raphael: Maybe reword as
> * if the result type has multiple views the service method returns the view used to render the response along with the result and error.

docs/FAQ.md
```diff
@@ -56,17 +56,15 @@ only applies for non required attributes with default values.
...
+  this view to create a viewed result type.
 * a views package is generated at the service level which defines a viewed
-  result type for each result with multiple views. This viewed result type
-  has identical field names and types but uses pointers for all fields so that
-  view specific validation logic may be generated. Constructors are generated
-  in the service package to convert a result type to a viewed result type and
-  vice versa.
-* the generated endpoint function uses the view returned by the service method
-  to create a viewed result type.
+  result type for each method result. This viewed result type has identical
```
raphael: Do users have to deal with viewed result types directly? If not maybe I would simplify this entire section with something like:
> The view is used to validate the content of the result type client side.

docs/FAQ.md
```diff
@@ -80,3 +78,6 @@ the attributes in the viewed result type as defined by the view and converts
 the viewed result type into the service result type using the appropriate
 constructor.
 
+NOTE: If a result type is defined without any views, a "default" view is added
```
raphael: Should specify that the default view contains the same attributes as the result type (no need to go into the details of collections at this level).

kumarsiva07:
`Response(StatusOK, func() {
				Header("content_type:Content-Type")
				Body("response_string")
			})`
After this commit I am could not use Body in response.

`tracker/gen/http/form/server/encode_decode.go:199:14: res.ResponseString undefined (type *views.FormPage has no field or method ResponseString)
`

Issue #1740, opened 2018-06-07
Can't serve websocket using basic auth


slaterx:
I'm designing an API to serve both http and ws with basic auth. HTTP code works fine but WebSocket code has no generated authentication code. 

Attempting to do the same thing without basic_auth works fine too. 

## Expected Behavior
Actions can be served using websocket protocol and basic authentication. 

## Actual Behavior
Actions do not have the basic authentication boilerplate code if Scheme is ws/wss.
CLI tests return 401 unauthorized on server and bad status on CLI.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

design.go:

```golang
package design

import (
        . "github.com/goadesign/goa/design"
        . "github.com/goadesign/goa/design/apidsl"
)

var _ = API("WebSocketBug", func() {
        Scheme("http")
        Origin("*", func() {
                Methods("GET", "POST", "PUT", "PATCH", "DELETE", "OPTIONS")
                Headers("Accept", "Content-Type", "Authorization")
                Expose("Content-Type", "Origin")
                MaxAge(600)
                Credentials()
        })
        Host("localhost:8080")

        Consumes("application/json")
        Produces("application/json")

})

var BasicAuth = BasicAuthSecurity("basic_auth")

var _ = Resource("events", func() {
        BasePath("/events")

        Security(BasicAuth)

        Action("watch", func() {
                Description("Watch events via WebSocket connection")
                Routing(GET("/"))
                Scheme("ws")
                Security(BasicAuth)
                Params(func() {
                        Param("fooID", String, "foo ID")
                })
                Response(SwitchingProtocols)
                Response(BadRequest, ErrorMedia)
                Response(NotFound)
        })

        Action("get", func() {
                Description("Get foo by id")
                Routing(GET("/:fooID"))
                Security(BasicAuth)
                Params(func() {
                        Param("fooID", String, "foo ID")
                })
                Response(OK)
                Response(BadRequest, ErrorMedia)
                Response(NotFound)
        })

})
```

That generates the following boilerplate code:

client/events.go
```golang
// Code generated by goagen v1.3.1, DO NOT EDIT.
//
// API "WebSocketBug": events Resource Client
//
// Command:
// $ goagen
// --design=bug/design
// --out=$(GOPATH)/src/bug
// --version=v1.3.1

package client

import (
        "context"
        "fmt"
        "golang.org/x/net/websocket"
        "net/http"
        "net/url"
)

// GetEventsPath computes a request path to the get action of events.
[...]

PR #1755, opened 2018-06-14
Allow basic auth to attach Authorization header for a websocket (#1740)

davidplumpton:
Currently if the cli client is making a websocket request then any supplied user/pass values will not be added as a basic auth header. This change causes the generated client code to add the Authorization header.

Developed together with @slaterx

### DIFF
--- a/client/signers.go
+++ b/client/signers.go
@@ -1,8 +1,11 @@
 package client
 
 import (
+	"encoding/base64"
 	"fmt"
 	"net/http"
+
+	"golang.org/x/net/websocket"
 )
 
 type (
@@ -10,6 +13,10 @@ type (
 	Signer interface {
 		// Sign adds required headers, cookies etc.
 		Sign(*http.Request) error
+	}
+
+	WebSocketSigner interface {
+		SignWebSocket(*websocket.Conn) error
 	}
 
 	// BasicSigner implements basic auth.
@@ -85,6 +92,13 @@ type (
 func (s *BasicSigner) Sign(req *http.Request) error {
 	if s.Username != "" && s.Password != "" {
 		req.SetBasicAuth(s.Username, s.Password)
+	}
+	return nil
+}
+
+func (s *BasicSigner) SignWebSocket(cfg *websocket.Config) error {
+	if s.Username != "" && s.Password != "" {
+		cfg.Header.Set("Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(s.Username+":"+s.Password)))
 	}
 	return nil
 }
--- a/client/signers.go
+++ b/client/signers.go
@@ -8,16 +8,22 @@ import (
 	"golang.org/x/net/websocket"
 )
 
+const (
+	// AuthHeader is the standard Authorization header name.
+	AuthHeader = "Authorization"
+)
+
 type (
-	// Signer is the common interface implemented by all signers.
+	// Signer is the common interface implemented by all http signers.
 	Signer interface {
 		// Sign adds required headers, cookies etc.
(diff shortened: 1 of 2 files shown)


client/signers.go
```diff
@@ -12,6 +15,11 @@ type (
 		Sign(*http.Request) error
 	}
 
+	// WebSocketSigner is the interface implemented by all signers using websocket.
+	WebSocketSigner interface {
+		SignWebSocket(*websocket.Conn) error
```
raphael: Please call the method `Sign` to avoid stuttering :)

client/signers.go
```diff
@@ -89,6 +97,14 @@ func (s *BasicSigner) Sign(req *http.Request) error {
 	return nil
 }
 
+// SignWebSocket adds the basic auth header to the WebSocket dial config
+func (s *WebSocketSigner) SignWebSocket(cfg *websocket.Config) error {
```
raphael: That does not work, you need to have one struct per security scheme each implementing the `WebSocketSigner` interface similarly to how it's done with the existing `Signer` interface.

goagen/gen_client/generator.go
```diff
@@ -1011,6 +1014,13 @@ func (c *Client) {{ $funcName }}(ctx context.Context, path string{{ if .Params }
 	if err != nil {
 		return nil, err
 	}
+	if c.BasicAuthSigner != nil {
+		wsSigner, ok := c.BasicAuthSigner.(*goaclient.BasicSigner)
```
raphael: It would be better to have explicit fields on the client struct for each possible web socket signer type. So something like:
```go
if c.BasicAuthWSSigner != nil {
    c.BasicAuthWSSigner.Sign(cfg)
}
```
Also the code should check for the proper security scheme and use the corresponding field (`BasicAuthWSSigner` or `APIKeyWSSigner` or ...) like is done in the existing signer case.

client/signers.go
```diff
@@ -63,6 +69,8 @@ type (
 	Token interface {
 		// SetAuthHeader sets the Authorization header to r.
 		SetAuthHeader(r *http.Request)
+		// SetWsAuthHeader sets the Authorization header to cfg.
```
raphael: Ah we need to think a bit more about this as it would be a breaking change... We may need to introduce a new `WSToken` interface as well. This may also mean a `WSTokenSource` interface :( I'm starting to wonder if a better approach wouldn't be to leave all of this the way it is and instead change the generated code so that it configures the websocket using the result of  signing a dummy request (i.e. extract the auth header and use it to initialize the websocket config).

client/signers.go
```diff
@@ -8,16 +8,22 @@ import (
...
+	// Signer is the common interface implemented by all http signers.
 	Signer interface {
 		// Sign adds required headers, cookies etc.
 		Sign(*http.Request) error
 	}
 
-	// WebSocketSigner is the interface implemented by all signers using websocket.
-	WebSocketSigner interface {
-		SignWebSocket(*websocket.Conn) error
+	// WsSigner is the common interface for websocket signers
```
raphael: Please use `WSSigner` everywhere, see [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

client/signers.go
```diff
@@ -8,16 +8,22 @@ import (
...
+	WsSigner interface {
+		// WsSign adds required headers, cookies, etc. to websocket config
```
raphael: Maybe `SignWS` would be a better name?

raphael:
Looks like we hit a dead-end, thanks for working through this.

PR #1762, opened 2018-06-20
Add support for readonly attributes

dikhan:
This pull request adds support for ReadOnly attributes to goa DSL. Payload and MediaType attributes can be set as readonly and the corresponding swagger file generated will render the 'readOnly' as per [swagger 'readOnly' property documentation](https://swagger.io/specification/#fixed-fields-20).

This PR addresses issue #1736

Looking forward to your feedback.

All tests seemed to pass successfully fine with the new change (and additional tests added). An output is attached for the reference:

````
make test
[1529450979] GenClient Suite - 33/33 specs ••••••••••••••••••••••••••••••••• SUCCESS! 10.373164744s PASS
[1529450979] Json Encoding Suite - 3/3 specs ••• SUCCESS! 1.105313ms PASS
[1529450979] Design Suite - 98/98 specs •••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• SUCCESS! 40.898348ms PASS
[1529450979] Test Suite - 1/1 specs • SUCCESS! 797.221µs PASS
[1529450979] GenController Suite - 6/6 specs •••••• SUCCESS! 67.86682ms PASS
[1529450979] GenMain Suite - 6/6 specs •••••• SUCCESS! 1.739633777s PASS
[1529450979] JWT Security Middleware - 25/25 specs ••••••••••••••••••••••••• SUCCESS! 5.999741ms PASS
[1529450979] Codegen Suite - 79/79 specs ••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• SUCCESS! 81.418711ms PASS
[1529450979] Client Suite - 3/3 specs ••• SUCCESS! 1.021765ms PASS
[1529450979] GenJS Suite - 3/3 specs ••• SUCCESS! 18.69709ms PASS
[1529450979] Middleware Suite - 24/24 specs •••••••••••••••••••••••• SUCCESS! 9.606494ms PASS
=== RUN   TestMatchOrigin
--- PASS: TestMatchOrigin (0.00s)
=== RUN   TestMatchOriginRegexp
--- PASS: TestMatchOriginRegexp (0.00s)
PASS
[1529450979] Gzip Suite - 18/18 specs •••••••••••••••••• SUCCESS! 33.652479ms PASS
[1529450979] GenSchema Suite - 4/4 specs •••• SUCCESS! 6.262237ms PASS
[...]

PR #1776, opened 2018-06-24
Backport to v1: Add support for readonly attributes (#1762)

dikhan:
* Add support for readonly attributes

* run go fmt

* use metadata to set readonly flag instead

# Conflicts:
#	design/apidsl/attribute.go


raphael:
Looks great, thank you!

PR #1769, opened 2018-06-22
Fix minItems/maxItems for array type in a generated JSON schema (#1768).

k-lepa:
Hello,

I fixed #1768.

Thanks

PR #1779, opened 2018-06-25
Backport to v2: Fix minItems/maxItems for Array in a generated JSON schema

k-lepa:
Hello, 

It's a backport of #1769.

Thanks


http/codegen/openapi/json_schema.go
```diff
@@ -271,7 +274,7 @@ func GenerateTypeDefinition(api *design.APIExpr, ut *design.UserTypeExpr) {
 }
 
 // TypeSchema produces the JSON schema corresponding to the given data type.
-func TypeSchema(api *design.APIExpr, t design.DataType) *Schema {
+func TypeSchema(api *design.APIExpr, t design.DataType, val *design.ValidationExpr) *Schema {
```
nitinmohan87: In goa v2, validations are tied to an `AttributeExpr` and not a `DataType`. So moving the validation logic from `buildAttributeSchema` to here doesn't sound right. Would it be possible to explain why you moved the logic?

http/codegen/openapi/json_schema.go
```diff
@@ -434,34 +459,14 @@ func (s *Schema) Dup() *Schema {
...
-	s.Merge(TypeSchema(api, at.Type))
+	s.Merge(TypeSchema(api, at.Type, at.Validation))
 	if s.Ref != "" {
 		// Ref is exclusive with other fields
 		return s
 	}
 	s.DefaultValue = toStringMap(at.DefaultValue)
 	s.Description = at.Description
 	s.Example = at.Example(api.Random())
-	val := at.Validation
```
nitinmohan87: See comment above. This seems to be the correct place to go through the validations on the AttributeExpr.

nitinmohan87:
@k-lepa Thank you so much for the PR! It looks great. Left one question which needs clarification.

http/codegen/openapi/json_schema.go
k-lepa: @nitinmohan87 Look at this:

```
http/codegen/openapi/openapi_v2_builder.go
427:            schema = TypeSchema(root.Design.API, r.Body.Type, r.Body.Validation)
491:                    schema := TypeSchema(root.Design.API, design.ErrorResult, nil)
569:                            Schema:      TypeSchema(root.Design.API, endpoint.Body.Type, endpoint.Body.Validation),

http/codegen/openapi/json_schema.go
171:                    requestSchema = TypeSchema(api, a.MethodExpr.Payload.Type, a.MethodExpr.Payload.Validation)
196:                                    targetSchema = TypeSchema(api, mt, val)
200:                                    targetSchema.AnyOf = []*Schema{firstSchema, TypeSchema(api, mt, val)}
202:                                    targetSchema.AnyOf = append(targetSchema.AnyOf, TypeSchema(api, mt, val))
462:    s.Merge(TypeSchema(api, at.Type, at.Validation))

```

In all that places I need to copy the validations. If someone adds a TypeSchema call, he will have to add the validations too. 

If you know a better option, I'm ready to change it.

nitinmohan87: I think we can probably export this `buildAttributeSchema` to `AttributeSchema` and call this wherever `TypeSchema` is called? The validation logic shall remain in the `AttributeSchema` since validations are tied to an AttributeExpr.

Probably these lines can still call `TypeSchema`
```
491:                    schema := TypeSchema(root.Design.API, design.ErrorResult, nil)
```
Makes sense?

k-lepa: OK, I'll refactor the ```buildAttributeSchema```.

http/codegen/openapi/json_schema.go
```diff
@@ -456,12 +458,26 @@ func buildAttributeSchema(api *design.APIExpr, s *Schema, at *design.AttributeEx
...
+			s.MaxItems = val.MaxLength
+		} else {
+			s.MaxLength = val.MaxLength
+		}
 	}
 	s.Required = val.Required
+}
+
+// AttributeTypeSchema produces the JSON schema corresponding to the given attribute.
+func AttributeTypeSchema(api *design.APIExpr, at *design.AttributeExpr) *Schema {
```
nitinmohan87: Is there a reason not to change `buildAttributeSchema` to `AttributeSchema` instead of introducing a new function? Or probably call `buildAttributeSchema` inside `AttributeSchema`? From what I see, both these functions do pretty much the same thing. `buildAttributeSchema` adds more fields like `Example`, `DefaultValue`, `Description`. Are they not desired?

nitinmohan87: Sounds good! Thanks.

PR #1780, opened 2018-06-25
Add RFC3339 date format

k-lepa:
Hello,

I added support of RFC3339 date format. This format exists in [Swagger 2.0 Spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types).

Thanks

### DIFF
--- a/design/example.go
+++ b/design/example.go
@@ -6,7 +6,6 @@ import (
 	"regexp"
 	"time"
 
-	"github.com/goadesign/goa"
 	regen "github.com/zach-klippenstein/goregen"
 )
 
@@ -153,8 +152,8 @@ func (eg *exampleGenerator) generateFormatExample() interface{} {
 	if res, ok := map[string]interface{}{
 		"email":     eg.r.faker.Email(),
 		"hostname":  eg.r.faker.DomainName() + "." + eg.r.faker.DomainSuffix(),
-		"date":      time.Unix(int64(eg.r.Int())%1454957045, 0).Format(goa.LayoutRFC3339Date), // to obtain a "fixed" rand
-		"date-time": time.Unix(int64(eg.r.Int())%1454957045, 0).Format(time.RFC3339),          // to obtain a "fixed" rand
+		"date":      time.Unix(int64(eg.r.Int())%1454957045, 0).Format("2006-01-02"), // to obtain a "fixed" rand
+		"date-time": time.Unix(int64(eg.r.Int())%1454957045, 0).Format(time.RFC3339), // to obtain a "fixed" rand
 		"ipv4":      eg.r.faker.IPv4Address().String(),
 		"ipv6":      eg.r.faker.IPv6Address().String(),
 		"ip":        eg.r.faker.IPv4Address().String(),
--- a/design/example.go
+++ b/design/example.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/goadesign/goa"
 	regen "github.com/zach-klippenstein/goregen"
 )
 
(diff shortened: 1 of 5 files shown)


raphael:
Thanks, can you elaborate a little bit on why the current `FormatDateTime` didn't work?

k-lepa:
RFC 3339 defines "date-time" in 5.6 as:
```
full-date       = date-fullyear "-" date-month "-" date-mday
full-time       = partial-time time-offset
date-time       = full-date "T" full-time
```

So, without the `FormatDate` I can send dates with `full-date "T" full-time` format only.

k-lepa:
@raphael, I understand now that this PR is not complete =( For full support I need to change JSON marshalling. I'm not sure that it's necessary to do it.

What do you think about it?

raphael:
Have you tried using a RFC3339 date value for an attribute with the `DateTime` format and does that not work? I would have expected the Go stdlib to be OK with that.

k-lepa:
Look at [this](https://play.golang.org/p/UMIhCpcTvXL):
```
package main

import (
	"encoding/json"
	"fmt"
	"os"
	"time"
)

const layoutRFC3339Date = "2006-01-02"

func main() {
	//
	// The parse problem of date as RFC3339 
	//
	_, err := time.Parse(time.RFC3339, "2015-03-15")
	if err != nil {
		fmt.Fprintf(os.Stderr, "the first parse is not ok: %s\n", err)
	} else {
		fmt.Println("the first is ok")
	}
	// Output: the first is not ok: parsing time "2015-03-15" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "T"

	dt, err := time.Parse(layoutRFC3339Date, "2015-03-15")
	if err != nil {
		fmt.Fprintf(os.Stderr, "the second parse is not ok: %s\n", err)
	} else {
		fmt.Println("the second is ok")
	}
	// Output: the second is ok


	//
	// This code demonstrates the problem of standard marshalling
	//
	res, _ := json.Marshal(dt)
	fmt.Printf("the first attempt: %q\n", res)
	// Output: the first attempt: "\"2015-03-15T00:00:00Z\""

	res, _ = json.Marshal(rfc3339FullDate(dt))
	fmt.Printf("the second attempt: %q\n", res)
	// Output: the second attempt: "\"2015-03-15\""
}

type rfc3339FullDate time.Time

func (d rfc3339FullDate) MarshalJSON() ([]byte, error) {
	return []byte(fmt.Sprintf(`"%s"`, time.Time(d).Format(layoutRFC3339Date))), nil
}
```

raphael:
Thanks for all the details. This is the flow I had in mind:
1. You start with a date as a time.Time value
2. You "send" that date value either in the client as an attribute of the payload or as a server as an attribute of the result.
3. The other side "receives" that date value as a time.Time value.

This would all work, the value would get marshaled as a `RFC3339` format which 0s for the hour, minutes etc. So I'm curious what the use case is here. Is it that you need to send a RFC3339 Date to your client which is not a goa client?

k-lepa:
We don't use a goa client because our clients can use Swagger API only. In our API we want to emphasize that some dates aren't in the date-time format.

raphael:
OK make sense, you'll also need to add it to the DSL: https://github.com/goadesign/goa/blob/2c35cf19edddb51151921551b3f10d75281de473/design/apidsl/attribute.go#L424

design/example.go
```diff
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/goadesign/goa"
```
raphael: This import is not OK, the design package shouldn't rely on the goa package - it should be the opposite, thanks!

validation.go
```diff
@@ -51,6 +54,9 @@ const (
 
 	// FormatRFC1123 defines RFC1123 date time values.
 	FormatRFC1123 = "rfc1123"
+
+	// LayoutRFC3339Date is a layout of RFC3339 date for use in time.Parse.
+	LayoutRFC3339Date = "2006-01-02"
```
raphael: Do we really need a documented constant here? wondering if just an inline string in two places isn't simpler - this is not likely to change ever...

raphael:
Thanks - left a couple of comments in the PR.

validation.go
k-lepa: I agree that an inline string is better for this use case.

design/example.go
k-lepa: OK, I'll fix it.

raphael:
Thank you for making these changes, looks great. Would you mind backporting this to v1 as well? thanks again.

PR #1781, opened 2018-06-25
Fix example generation

### LINKED ISSUE #1770: Example service does not match generated interface
running `goa example` produces a service which does not satisfy the generated service interface.

The DSL at the bottom of this issue will generate the following code:

./gen/service_body_multiple_view/service.go:

`// Service is the ServiceBodyMultipleView service interface.
type Service interface {
	// MethodBodyMultipleView implements MethodBodyMultipleView.
	MethodBodyMultipleView(context.Context, *MethodBodyMultipleViewPayload) (res *Resulttypemultipleviews, err error)
}`

./service_body_multiple_view.go:

`func (s *serviceBodyMultipleViewSvc) MethodBodyMultipleView(ctx context.Context) (res *servicebodymultipleview.Resulttypemultipleviews, view string, err error) {
     ...
}`

Please fill the sections below:

## Expected Behavior
`goa example` produces functioning code
## Actual Behavior
`goa example` produces code incompatible with `goa gen`
`goa gen` then `goa example` and observe compile errors
## Specifications
  - goa Version: v2.0.0-wip
  - Platform OSX 10.13.2

Example DSL:

```
package design

import (
	. "goa.design/goa/http/dsl"
)

var _ = API("goa-bug-repros", func() {
	Title("GOA-Bug-Repros")
	Description("For Debugging")
	Server("http:/localhost:8080")
})

var RT = ResultType("ResultTypeMultipleViews", func() {

	Attribute("a", String)
	Attribute("b", String)
	Attribute("c", String)

	View("default", func() {
		Attribute("a")
		Attribute("b")
		Attribute("c")
	})

	View("tiny", func() {
		Attribute("c")
	})
})

var _ = Service("ServiceBodyMultipleView", func() {
	Method("MethodBodyMultipleView", func() {

		Result(RT, func() {
			View("tiny")
		})

		HTTP(func() {
			POST("/")
			Response(StatusOK, func() {
				Header("c:Location")
			})
		})
	})
})
```

nitinmohan87:
* Fixes #1770 
* auth.go is generated only if security schemes are defined in DSL
* Remove "DO NOT EDIT" comment from auth.go

### DIFF
--- a/codegen/service/auth_funcs.go
+++ b/codegen/service/auth_funcs.go
@@ -25,6 +25,7 @@ func AuthFuncsFile(genpkg string, root *design.RootExpr) *codegen.File {
 
 	var (
 		sections []*codegen.SectionTemplate
+		generate bool
 	)
 	{
 		specs := []*codegen.ImportSpec{
@@ -46,6 +47,7 @@ func AuthFuncsFile(genpkg string, root *design.RootExpr) *codegen.File {
 		for _, s := range root.Services {
 			svc := Services.Get(s.Name)
 			if len(svc.Schemes) > 0 {
+				generate = true
 				sections = append(sections, &codegen.SectionTemplate{
 					Name:   "security-authfuncs",
 					Source: dummyAuthFuncsT,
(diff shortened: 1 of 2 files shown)


raphael:
Thank you! I was hoping we could keep the `Header` function of the `codegen` package only for generated code i.e. code that cannot be edited. For example code the header is generated as part of the example itself, can we do the same thing here?

Issue #1574, opened 2018-03-08
Support for force-generation of types using metadata tag


arangamani:
This is a feature request.

I have a type I define for my project in a top level public design package that is used by many microservices. I want to be able to use this type in the business logic of those microservices. Some parts of the type are going to be used to build the microservice API request and result types but not all of them.

I want to be able to have a way to provide a metadata tag which will force-generate the type in the service package.

## Specifications

  - goa Version: v2
  - Platform (Linux/Windows etc.): Mac

PR #1782, opened 2018-06-25
Force generate types from different package

nitinmohan87:
* Add "type:generate:force" in Metadata expression of a UserType to force its generation on the services even though the type is not used in the design. (#1574)

### DIFF
--- a/dsl/metadata.go
+++ b/dsl/metadata.go
@@ -12,7 +12,32 @@ import (
 // services and API definitions.
 //
 // While keys can have any value the following names are handled explicitly by
-// goa when set on attributes.
+// goa when set on attributes or types.
+//
+// type:generate:force specified on a Type will generate a Go struct in the
+// service packages specified in the value. If no services are specified, it
+// generates the type in all the services in the design package. The Type
+// definition to force generate can exist in an external package and must be
+// imported in the API design.
+//
+//        package my_package
+//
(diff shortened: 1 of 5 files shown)


dsl/metadata.go
```diff
@@ -12,7 +12,32 @@ import (
 // services and API definitions.
 //
 // While keys can have any value the following names are handled explicitly by
-// goa when set on attributes.
+// goa when set on attributes or types.
+//
+// type:generate:force specified on a Type will generate a Go struct in the
```
raphael: Maybe something like:
```
type:generate:force forces the code generation for the type it is defined on. By default
goa only generates types that are used explicitly by the service methods. This metadata
key makes it possible to override this behavior and forces goa to generate the corresponding
struct. The value is a slice of strings that lists the names of the services for which to generate
the struct. If left empty then the struct is generated for all services.
```

dsl/metadata.go
```diff
@@ -12,7 +12,32 @@ import (
...
+// service packages specified in the value. If no services are specified, it
+// generates the type in all the services in the design package. The Type
+// definition to force generate can exist in an external package and must be
+// imported in the API design.
+//
+//        package my_package
```
raphael: I don't think we need to use this case as an example. The fact is you can import design packages at any time and this has no bearing on this feature. I would just use a simple example here.

PR #1785, opened 2018-06-27
added import aliases to external packages in convert.go

ccrossley:
convert.go external imports where being omitted if the package name didn't match the file name.

### DIFF
--- a/codegen/service/convert.go
+++ b/codegen/service/convert.go
@@ -85,19 +85,27 @@ func ConvertFile(root *design.RootExpr, service *design.ServiceExpr) (*codegen.F
 	}
 
 	// Retrieve external packages info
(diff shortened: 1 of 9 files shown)


raphael:
Thank you for the great PR, do you think you could write a few tests as well? thanks!

raphael:
Looks great, thanks for all the hard work! Just one question: did you mean to call `SectionCodeFromInputAndMethods` actually `SectionCodeFromImportsAndMethods`?

ccrossley:
Whoops!  Yea I somehow borked the refactor.  I have a commit in the pipe to fix it, hopefully it will show up here soon!

raphael:
Great, thank you!

PR #1792, opened 2018-06-29
Add support for JWT Token authorization via query parameter

jiekang:
This adds support for extracting the JWT token from the query parameter. How does it look?

### DIFF
--- a/middleware/security/jwt/jwt.go
+++ b/middleware/security/jwt/jwt.go
@@ -18,10 +18,10 @@ import (
 // scopes claim in the JWT and ensures goa-defined Security DSLs are properly validated.
 //
 // The steps taken by the middleware are:
-//
-//     1. Validate the "Bearer" token present in the "Authorization" header against the key(s)
+//     1. Extract the "Bearer" token from the Authorization header or query parameter
+//     2. Validate the "Bearer" token against the key(s)
 //        given to New
-//     2. If scopes are defined in the design for the action, validate them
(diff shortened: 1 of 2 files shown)


middleware/security/jwt/jwt.go
```diff
@@ -134,6 +137,30 @@ func New(resolver KeyResolver, validationFunc goa.Middleware, scheme *goa.JWTSec
...
+
+	incomingToken := strings.Split(val, " ")[1]
+
+	return incomingToken, nil
+}
+
+func extractTokenFromQueryParam(schemeName string, req *http.Request) (string, error) {
+	incomingToken := req.URL.Query().Get(schemeName)
+	if incomingToken == "" {
+		return "", ErrJWTError(fmt.Sprintf("missing parameter %q %+v", schemeName, req))
```
raphael: Do we need to write the complete request object in the error? wondering if just writing the URL would be enough? what does this look like?

raphael:
Thank you for the PR, looks great! I left one minor comment in the code.

middleware/security/jwt/jwt.go
jiekang: Thank you for the catch! This printing of request object was not supposed to be there; I added it while I was trying to verify my tests.

raphael:
Perfect, thank you! do you think you could backport this to `v1` as well? simply cherry-pick the commit in a branch created off of the `v1` branch and send a new PR. Thanks!

PR #1793, opened 2018-06-29
Fix pointer bug in validate when default and enum set

### LINKED ISSUE #1778: Default is not working
```
Attribute("page_number", Int, "Search string of object", func() {
        Default(1)
    })
```
I didn't added this attribute in required but i am getting that in params.

```
HTTP(func() {
			GET("/")
			Params(func() {		
				Param("page_number")
			})
			Response(StatusOK)
		})
```
I am not getting the default value if the param not passed from request.

nitinmohan87:
Fixes #1778


raphael:
Looks great thank you!

PR #1800, opened 2018-07-04
Add support for JWT Token authorization via query parameter

jiekang:
This is a backport of https://github.com/goadesign/goa/pull/1792 against `v1`

The cherry-pick via git was not clean so I chose to manually backport it myself. How does it look?

### DIFF
--- a/middleware/security/jwt/jwt.go
+++ b/middleware/security/jwt/jwt.go
@@ -69,13 +69,11 @@ func New(validationKeys interface{}, validationFunc goa.Middleware, scheme *goa.
 			)
 
 			if scheme.In == goa.LocHeader {
-				incomingToken, err = extractTokenFromHeader(scheme.Name, req)
-				if err != nil {
+				if incomingToken, err = extractTokenFromHeader(scheme.Name, req); err != nil {
 					return err
 				}
 			} else if scheme.In == goa.LocQuery {
-				incomingToken, err = extractTokenFromQueryParam(scheme.Name, req)
-				if err != nil {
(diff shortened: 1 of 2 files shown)


middleware/security/jwt/jwt.go
```diff
@@ -62,24 +63,27 @@ func New(validationKeys interface{}, validationFunc goa.Middleware, scheme *goa.
...
-			}
+			var (
+				incomingToken string
+				err           error
+			)
 
-			if !strings.HasPrefix(strings.ToLower(val), "bearer ") {
-				return ErrJWTError(fmt.Sprintf("invalid or malformed %q header, expected 'Authorization: Bearer JWT-token...'", val))
+			if scheme.In == goa.LocHeader {
+				incomingToken, err = extractTokenFromHeader(scheme.Name, req)
```
raphael: This could be written as:
```go
if incomingToken, err = extractTokenFromHeader(scheme.Name, req); err != nil {
    return err
}
```

middleware/security/jwt/jwt.go
```diff
@@ -62,24 +63,27 @@ func New(validationKeys interface{}, validationFunc goa.Middleware, scheme *goa.
...
+				if err != nil {
+					return err
+				}
+			} else if scheme.In == goa.LocQuery {
+				incomingToken, err = extractTokenFromQueryParam(scheme.Name, req)
```
raphael: ditto

jiekang:
I've addressed the comments as suggested. Thanks for the review!

raphael:
Thank you!

PR #1823, opened 2018-07-26
Added service name to generated swagger method definitions

ccrossley:


raphael:
This is great! thanks for the fix!

PR #1825, opened 2018-07-30
Remove redundant variables declaration

lfaoro:
Checking contribution ease-fullness


raphael:
Nice catch, thank you! (super easy)

PR #1827, opened 2018-07-31
Fix "err redefined" error using multipart request

nitinmohan87:
Fixes code build when using design from https://github.com/goadesign/plugins/issues/52


raphael:
Thank you!

Issue #1828, opened 2018-07-31
index variables in generated code for iterating over nested ArrayOf


buth:
Hey all; I ran into an issue with the generated code for nested `ArrayOf` values. The details are bellow. Thank you!

## Expected Behavior

The generated code should use different sequential index variable names to iterate over nested array values.

## Actual Behavior

The generated code uses the same index variable name to iterate over nested array values.

```go
// NewPostNestedArrayOf builds a arrays service post endpoint payload.
func NewPostNestedArrayOf(body *PostRequestBody) *arrays.NestedArrayOf {
	v := &arrays.NestedArrayOf{}
	if body.Attribute != nil {
		v.Attribute = make([][]string, len(body.Attribute))
		for j, val := range body.Attribute {
			v.Attribute[j] = make([]string, len(val))
			for j, val := range val {
				v.Attribute[j][j] = val
			}
		}
	}
	return v
}
```

## Steps to Reproduce the Problem (including complete and simple design if relevant)

This design will generate the code above under `http/arrays/server`.

```go
package design

import . "goa.design/goa/http/design"
import . "goa.design/goa/http/dsl"

var _ = API("api", func() {
	Title("NestedArrayOf Bug")
})

var NestedArrayOf = Type("NestedArrayOf", func() {
	Attribute("attribute", ArrayOf(ArrayOf(String)))
})

var _ = Service("arrays", func() {
	Method("post", func() {
		Payload(NestedArrayOf)
		Result(Empty)
		HTTP(func() {
			POST("/arrays")
			Response(StatusNoContent)
		})
	})
})
```

## Specifications

  - goa Version: [v2](https://github.com/goadesign/goa/commit/a72f5dd8c6996ecfc45a10f03d5a0f6f6b0ef984) 
  - Platform (Linux/Windows etc.): macOS

nitinmohan87:
Thanks for the PR!

PR #1829, opened 2018-08-01
Fix nested array index variables (#1828)

buth:
* Count target "[" instances rather than "." to determine array depth

* Updated test code strings to expect "i" as the first array index variable

* Regenerated examples


raphael:
Thank you!

PR #1833, opened 2018-08-02
Fix err undefined when MultipartRequest is used with Param

nitinmohan87:
If `MultipartRequest` DSL is used with `Param` or `Header` DSL in an endpoint, code build fails with "err undefined" error. This issue surfaced after merging https://github.com/goadesign/goa/pull/1827


raphael:
Thank you!

PR #1837, opened 2018-08-07
Fix Project for seen user types

nitinmohan87:
When a seen result type is projected again using a different view, the code [inadvertently modifies the type of the original seen result type](https://github.com/goadesign/goa/blob/3b65d562232cb3c28c971119e27ac3b1075dc191/design/result_type.go#L302) instead of modifying a copy.


raphael:
Thank you!

Issue #1799, opened 2018-07-04
Add tests for design.SecurityExpr.EvalName()


tchssk:
Reference implementation: #1702

tchssk:
Already finished.

Issue #1798, opened 2018-07-04
Add tests for design.SchemeExpr.EvalName()


tchssk:
Reference implementation: #1702

tchssk:
Already finished.

Issue #1797, opened 2018-07-04
Add tests for design.FlowExpr.EvalName()


tchssk:
Reference implementation: #1702

tchssk:
Already finished.

PR #1849, opened 2018-08-25
Added test code for three issues

k-kurikuri:
- [#1797](https://github.com/goadesign/goa/issues/1797)
- [#1798](https://github.com/goadesign/goa/issues/1798)
- [#1799](https://github.com/goadesign/goa/issues/1799)

### DIFF
--- a/design/service_test.go
+++ b/design/service_test.go
@@ -132,3 +132,36 @@ func TestFlowExpr_EvalName(t *testing.T) {
 		}
 	}
 }
+
+func TestSchemeExpr_EvalName(t *testing.T) {
+	cases := map[string]struct {
+		kind     SchemeKind
+		expected string
+	}{
+		"OAuth2Kind":    {kind: OAuth2Kind, expected: "OAuth2Security"},
+		"BasicAuthKind": {kind: BasicAuthKind, expected: "BasicAuthSecurity"},
+		"APIKeyKind":    {kind: APIKeyKind, expected: "APIKeySecurity"},
+		"JWTKind":       {kind: JWTKind, expected: "JWTSecurity"},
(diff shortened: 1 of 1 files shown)


design/service_test.go
```diff
@@ -111,3 +111,73 @@ func TestErrorExprValidate(t *testing.T) {
...
+
+	// panic recover
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("should have panicked!")
+		}
+	}()
+
+	for k, tc := range cases {
+		se := &SchemeExpr{Kind: tc.kind}
```
raphael: It would be good to count how many tests do not panic to make sure there were 4. Otherwise the test could succeed even if one of the first 4 tests panicked.

design/service_test.go
```diff
@@ -111,3 +111,73 @@ func TestErrorExprValidate(t *testing.T) {
...
+
+func TestSecurityExpr_EvalName(t *testing.T) {
+	scheme1 := &SchemeExpr{SchemeName: "A"}
+	scheme2 := &SchemeExpr{SchemeName: ""}
+
+	cases := map[string]struct {
+		schemes  []*SchemeExpr
+		expected string
+	}{
+		"security with suffix":           {schemes: []*SchemeExpr{scheme1}, expected: fmt.Sprintf("Securityscheme A")},
```
raphael: Do we really need `fmt.Sprintf` on this line and the next?

design/service_test.go
k-kurikuri: You're right.  I'll fix it.

design/service_test.go
k-kurikuri: `fmt.Sprintf` was unnecessary. I'll fix it.

raphael:
Looks great, thank you!

PR #1851, opened 2018-08-30
Headers in swagger

psschroeter:
Headers were missing from swagger definition for input parameters. Returned early from the function which generated them. Assume this was an oversight?

Have the Authorization header showing up as an input parameter currently, but I think this may be wrong and we should maybe filter that header out it's defined also in the security section.


raphael:
Awesome, thank you!

PR #1863, opened 2018-09-19
bugfix type error in multipart-form payload

aya-eiya:
bugfix for #1862

type errors occur when set MultipartForm flag

Issue #1862, opened 2018-09-18
some attribute-type in Type cause compile errors in generated code when use as Payload


aya-eiya:
I found some unexpected behave in some goa generated codes.

## Expected Behavior

goagen generates no defect codes.

## Actual Behavior

the codes generated with goagen has some odd lines as golang code and those can't be compiled.

1. ArrayOf for Payload causes defined but not be used error
2. non-required Integer Type for Payload causes strconv.Itoa(int) type mismatch error

this seems to arise only if set multipart-form.

```
# github.com/aya-eiya/goa-test/apps/app
apps/app/controllers.go:72:2: rawArrs declared and not used
# github.com/aya-eiya/goa-test/apps/client
apps/client/my_resorce.go:63:28: cannot use payload.IntVal (type *int) as type int in argument to strconv.Itoa
```

## Steps to Reproduce the Problem (including complete and simple design if relevant)

I made the error reproducing project here.

https://github.com/aya-eiya/goa-test

## Specifications

  - goa Version:1.4.0
  - Platform:MacOSX

aya-eiya:
I made a PR( #1863 ) for fix this.
I confirmed the fix for error 2 above firstly.

aya-eiya:
I added a fix for error 1 to PR( #1863 ).

Both fix are not perfect to avoid all error cases like I noticed but I think it is somewhat better than current behavior.

raphael:
Thanks! I will review the PR soon.

aya-eiya:
PR #1865 and #1863 were merged, so I close this.

raphael:
Thank you for the hard work on this!

PR #1865, opened 2018-09-28
bugfix type error in multipart-form payload

aya-eiya:
bugfix for issue #1862

this PR is for master branch. previous PR for v1 is #1863

* bugfix type error in multipart-form payload

* bugfix Array attribute error in multipartform

* fix wrong comment line and add comment to new function

* fix fromString func to do not cause panic and support some other types for array element


raphael:
Good work! thank you.

PR #1702, opened 2018-05-03
Add tests for design.APIExpr.EvalName()

tchssk:
#1423

PR #1870, opened 2018-10-07
add simple tests for ServerExpr and HostExpr

### LINKED ISSUE #1707: Add tests for expr.ServerExpr.Attribute()
Reference implementation: #1702

shihanng:
This should close #1702 and close #1707 (which is for ServerExpr but it seems that the Attribute() method is now implemented by HostExpr).

### DIFF
--- a/design/server_test.go
+++ b/design/server_test.go
@@ -1,0 +1,67 @@
+package design
+
+import (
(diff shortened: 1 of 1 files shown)


raphael:
This is great, thank you!

Issue #1872, opened 2018-10-08
V2: Using expression "Security" in a service


nikolaas:
Hello to All!

The documentation for the expression `Security` says:

> Security must appear in a Service or Method expression.

But I got the error when I tried to define `Security` in a `Service` expression. If `Security` is defined in a `Method` expression the error disappears and `Security` expression works as expected.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

1. Create `design.go`:
```golang
package design

import (
	. "goa.design/goa/http/dsl"
)

var _ = API("myapi", func() {
	Title("This is my incredible api")
})

var OAuth2 = OAuth2Security("OAuth2", func() {
	ImplicitFlow("https://oauth2.mydomain.org")
	Scope("openid", "an OIDC required scope")
})

var _ = Service("myservice", func() {
	// Security throws the error when it defined here
	Security(OAuth2)

	Method("foo", func() {
		// Security works as expected when it defined here
		//Security(OAuth2)

		Payload(func() {
			Attribute("name", String)
		})
		Result(Int)

		HTTP(func() {
			POST("/foo")
		})
	})
})
```
2. Run code generation:
```bash
$ goa gen mydomain.ru/go/myapi/design
```

## Actual Behavior

```bash
$ goa gen mydomain.ru/go/myapi/design
exit status 2
panic:  is not the name of a child of the mapped attribute

goroutine 1 [running]:
goa.design/goa/design.(*MappedAttributeExpr).Map(0xc42030fb80, 0x156c4ce, 0xd, 0x0, 0x0)
	/Users/nstupak/go/src/goa.design/goa/design/mapped_attribute.go:100 +0x1b3
goa.design/goa/http/design.(*EndpointExpr).Finalize(0xc420311b80)
	/Users/nstupak/go/src/goa.design/goa/http/design/endpoint.go:464 +0x25d
goa.design/goa/eval.finalizeSet(0xc42028b700, 0xe, 0x10, 0x0, 0x0)
	/Users/nstupak/go/src/goa.design/goa/eval/eval.go:124 +0x99
goa.design/goa/http/design.(*RootExpr).WalkSets(0x18dc440, 0x15d6658)
	/Users/nstupak/go/src/goa.design/goa/http/design/root.go:95 +0x490
goa.design/goa/eval.RunDSL(0x7ffeefbff08b, 0xc4200246b0)
	/Users/nstupak/go/src/goa.design/goa/eval/eval.go:55 +0x40d
main.main()
[...]

PR #1873, opened 2018-10-08
Fix Security for using in a Service expression (#1872)

nikolaas:
Proposed fix for issue #1872


raphael:
Great PR, makes sense thank you!

PR #1876, opened 2018-10-10
Bring back tests for api

shihanng:
In #1864, `api_test.go` was removed but I think some tests can be kept? This PR is to bring back some of the test cases. Please feel free to ignore/close this PR if the removal of àpi_test.go` was intentional.

### DIFF
--- a/design/api_test.go
+++ b/design/api_test.go
@@ -1,10 +1,7 @@
 package design
 
 import (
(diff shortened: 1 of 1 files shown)


raphael:
Thank you, I had removed the code then added it back and forgot to add back the tests!

PR #1879, opened 2018-10-10
Support HTTP DSL in API for defining common errors across services

nitinmohan87:
Allow HTTP DSL to be defined at API level to define common errors across services.


raphael:
Thank you!

PR #1881, opened 2018-10-11
fix some linter's  warnings

andriikushch:
This pr contains style fixes like import sorting, removing redundant parentheses etc.

### DIFF
--- a/cmd/aliaser/main.go
+++ b/cmd/aliaser/main.go
@@ -162,7 +162,7 @@ func main() {
 		}
 
 		designImports := map[string]*PackageDecl{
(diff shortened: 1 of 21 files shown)


raphael:
Great, thank you! As a note I'm not convinced that having to specify the field name in literal struct values is always required - for example in tests it makes the declaration uglier for no real gain. But this PR has a lot of good stuff in it!

andriikushch:
You are welcome!

Completely agree that explicitly setting field's name in general is a question of personal codding's style and not always brings some benefit. 

Personally for me only main reason to have it, it is a case when people are reading code without IDE, for example directly from github, then it helps.

For example: `&Human{75}` it is not obvious is "75" is age or weight.

Issue #1871, opened 2018-10-07
Access to logger from UserBasicAuth


difengo:
Hi,

I created a small service with basic authentication enabled. As expected the 
UserBasicAuth function was generated. 

However, how can I access to the logger from that method as I can do it from the endpoints ? I could of course create a global variable but this would make the s.logger useless in endpoints funcs no?

raphael:
I believe this is not an issue anymore thanks to your great contributions :) - Please close the issue if that's the case otherwise provide a description of what is still pending. Thank you!

PR #1882, opened 2018-10-11
New logging package and integration of logger in security functions

difengo:
Hi,

As raised few days ago in issue #1871 , the generated security/auth functions did not able to easily use the service logger. To fix that, I created a new dedicated package for logging and modified few files. The passed logger is now using a richer interface that can be implemented through adapters.

I hope that you will see these changes useful.


difengo:
Build failed due to unmodified calc example.

Issue #1884, opened 2018-10-12
V2: The "Body" expression does not work when the argument is the name of an attribute


nikolaas:
Hello!

I tried to use the `Body` expression to map the `Result` props to HTTP response.

The documentation to the `Body` expression says:

>Body accepts one argument which describes the shape of the body, it can be:
>
>- The name of an attribute of the request or response type. In this case the
>  attribute type describes the shape of the body.
>
>- A function listing the body attributes. The attributes inherit the
>  properties (description, type, validations etc.) of the request or
>  response type attributes with identical names.

I specified the name of an attribute of the `Result` type and I got an error.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

1. Create a design file:
```golang
// example.org/app/design/design.go

package design

import (
	. "goa.design/goa/http/design"
	. "goa.design/goa/http/dsl"
)

var _ = API("api", func() {
	Title("api")
})

var OAuth2 = OAuth2Security("OAuth2", func() {
	ImplicitFlow("https://oauth2.example.org", "https://oauth2.example.org")
	Scope("openid", "an OIDC required scope")
})

var Person = Type("Person", func() {
	Attribute("id", String)
	Attribute("name", String)
})

var AddPersonPayload = Type("AddPersonPayload", func() {
	AccessToken("token")
	Attribute("person", Person)
	Required("person")
})

var _ = Service("person", func() {
	Security(OAuth2)
	
	Method("add", func() {
		Payload(AddPersonPayload)

		HTTP(func() {
			POST("/persons")
			Header("token:Authorization", String)
			Body("person")
			Response(StatusOK)
		})
	})
})
```
The method "add" of the "person" service is designed to send an authorization token in the Authorization header and send a person data in the body of a HTTP request.

2. Generate the code:
```bash
$ goa gen example.org/app/design
```

## Actual Behavior
```bash
$ goa gen example.org/app/design
exit status 1
service "person" HTTP endpoint "add": Body "id" is not found in Payload.
service "person" HTTP endpoint "add": Body "name" is not found in Payload.
```

## Expected Behavior

The code should be generated successfully.

## Specifications

  - goa Version: v2 (commit 2dbea94bd0223f71b91a14065ac7edd3ba2647bb)
  - Platform (Linux/Windows etc.): macOS 10.13.6

nikolaas:
I found one more bug after I fixed the first bug.

goagen generates a wrong code when I specify the `Body` expression as the name of an attribute.

[...]

PR #1885, opened 2018-10-12
Fix using the Body expression with the name of an attribute as an argument of the expression

nikolaas:
Proposed fix for issue #1884


raphael:
Do you think you could create a separate PR with only the golint fix? Thank you!

raphael:
Thank you - sorry it took so long to fix the unrelated build failures.

PR #1886, opened 2018-10-12
New logging package

difengo:
Hi,

Following to the #1871 issue raised few days ago, I created a new logging package and modified the security/auth scheme file to allow logging within these functions. I took this opportunity to create a richer interface for the logger and an adapter for the standard log package.

As these modifications had an impact on already generated components, I then regenerated the examples and updated them.

I also created a pull request in the plugins repository for the CORS example plugin to reflect these changes.

I hope you will find these changes helpful and please feel free to do the necessary adjustments.


raphael:
Wow that's a lot of work - thank you for doing that. One issue I see here is that many people have different opinions and preferences when it comes to logging. The interface you came up with is probably good in most cases but it does force certain implementations and mappings. Would it be possible to keep this as an example or external package rather than make goa opinionated in terms of logging?

difengo:
Hi Raphael,

It's indeed probably too many changes on the existing code base.
I'm going to think about how this could be turned into a plugin, but whatever maybe, my opinion is that the existing interface is far too limited and that the one I defined is the strict minimum to have something operational.

On 23 Oct 2018, at 05:12, Raphaël Simon <notifications@github.com<mailto:notifications@github.com>> wrote:


Wow that's a lot of work - thank you for doing that. One issue I see here is that many people have different opinions and preferences when it comes to logging. The interface you came up with is probably good in most cases but it does force certain implementations and mappings. Would it be possible to keep this as an example or external package rather than make goa opinionated in terms of logging?

raphael:
One thing to keep in mind is that the interface defined by goa is the interface *it* needs to log. It is not meant to be the interface your application uses. The idea is that you inject a logger into goa that implements that interface so that the goa generated code can call into it. The logger you inject can be anything and its interface is not limited by the interface defined by goa. Hope this makes sense...

difengo:
Yes it's clear. In that case the logger needs either to be recasted in the funcs or "generated" to expand the basic interface. The second option is definitively more elegant and could meet the plugin approach. 

However, this will not help with the auth functions issue. The auth functions defined in scheme.go unfortunately don't give access to the logger. 

This will be the same problem if an OpenTracing tracer needs to be used there as well...

These elements could be passed in the context of course but this would be less elegant than the way the logger is accessed in the service generated funcs today.

So, it would be nice to have the auth funcs signature modified to have the logger passed (and potentially a tracer) or to make the auth functions as methods of the service. The second option would be more elegant but would require more changes.

difengo:
I've just pushed a first version of a ZapLogger plugin. This should answer the first part of the problem. However, the auth funcs issue cannot be fixed without some changes as described in my previous comment.

difengo:
I finally found a way to make the logger (and potentially any other service feature) accessible from the auth funcs. Please have a look to pull request #1906 and let me know.

raphael:
Thank you for coming up with that!

Issue #1896, opened 2018-10-24
V2: goa gen doesn't write swagger extensions to openapi.yaml


nikolaas:
Hello

`goa gen` doesn't write swagger extensions into the `openapi.yaml` file, but the writing into the `openapi.json` file is correct.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

1. Create a design file:
```golang
// example.org/app/design/design.go

package design

import (
	. "goa.design/goa/http/design"
	. "goa.design/goa/http/dsl"
)

var _ = API("api", func() {
	Title("api")
})

var _ = Service("service", func() {
	Method("foo", func() {
		HTTP(func() {
			GET("/foo")
			Metadata("swagger:extension:x-exampleorg-bar", "baz")
			Response(StatusOK)
		})
	})
})
```
2. Generate the code:
```bash
$ goa gen example.org/app/design/design.go
```

## Actual Behavior

The code of the `example.org/app/gen/http/openapi.json` file:
```yaml
paths:
  /foo:
    get:
      operationId: service#foo
      responses:
        "200":
          description: OK response.
      schemes:
      - http
      - https
      summary: foo service
      tags:
      - service
```

## Expected Behavior

The code of the `example.org/app/gen/http/openapi.json` file:
```diff
paths:
  /foo:
    get:
      operationId: service#foo
      responses:
        "200":
          description: OK response.
      schemes:
      - http
      - https
      summary: foo service
      tags:
      - service
+  x-exampleorg-bar: baz
```

## Specifications

  - goa Version: V2 (commit b8878bb8daccec33839331a85e7e5b5f15c8a228)
  - Platform (Linux/Windows etc.): macOS 10.13.6

raphael:
Thank you for the great report and PR!

PR #1897, opened 2018-10-24
Fix writing of swagger extensions into openapi.yaml (#1896)

nikolaas:
Proposed fix for issue #1896


raphael:
This is great! thank you for doing this.

PR #1898, opened 2018-10-27
Apply ContentType DSL to override Content-Type of the response

k3a:
This change is to make the following design work:

```
		Result(String)
		HTTP(func() {
			GET("txt")
			Response(StatusOK, func() {
				ContentType("text/plain") 
			})
		})
```

Content-Type is set in the generated encoder like this
```
func EncodeGetTxtResponse(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
		res := v.(string)
		enc := encoder(ctx, w)
		body := res
		w.Header().Set("Content-Type", "text/plain")
		w.WriteHeader(http.StatusOK)
		return enc.Encode(body)
	}
}
```

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -1198,6 +1198,9 @@ const responseT = `{{ define "response" -}}
 	{{- if .ErrorHeader }}
 	w.Header().Set("goa-error", {{ printf "%q" .ErrorHeader }})
 	{{- end }}
+	{{- if .ContentType }}
+	w.Header().Set("Content-Type", "{{ .ContentType }}")
+	{{- end }}
 	w.WriteHeader({{ .StatusCode }})
 {{- end }}
 
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -262,6 +262,8 @@ type (
 	ResponseData struct {
 		// StatusCode is the return code of the response.
 		StatusCode string
+		// Content-Type of the response
+		ContentType string
 		// Description is the response description.
 		Description string
 		// Headers provides information about the headers in the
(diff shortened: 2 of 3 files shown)


k3a:
Finally I found time to understand how tests work and updated testdata. 

For example 
```
var ResultType = ResultType("ResultTypeMultipleViews", func() {
```

now also sets 
```
w.Header().Set("Content-Type", "resulttypemultipleviews")
```

---

and 
```
    var RT = ResultType("ResultTypeCollection", func() {
...
    Service("ServiceBodyCollection", func() {
        Method("MethodBodyCollection", func() {
            Result(CollectionOf(RT))
```

sets
```
w.Header().Set("Content-Type", "resulttypecollection; type=collection")
```

k3a:
Is something preventing merge @raphael ? I really need a way to send "Result(String)" responses as text/html or text/plain over HTML without JSON encoding to implement HTML pages. These simple string responses should work just fine in gRPC or other RPC-encodings - sending HTML document.

raphael:
@k3a finally got a chance to properly review this - thank you for putting this PR together! I don't think the change would work as is because setting the "Content-Type" header of all HTTP responses to the mime type of the result type would break quite a lot of things. Instead I would suggest that using the `ContentType` DSL would set a new flag to override the default behavior of the response encoder. The default response encoder would consider a new context key (e.g. `ContentTypeKey`) and the generated code would set that key when the content type is explicitly set in the DSL.

k3a:
Thanks for the review! I agree, it makes more sense to use the context key. I will update the pull request soon, I have been also busy recently..

nitinmohan87:
Thanks for the contribution again! Closing this PR as it is taken care of in #1942

PR #1899, opened 2018-10-27
Form encoding (application/x-www-form-urlencoded) for requests

k3a:
Allows requests with `Content-Type: application/x-www-form-urlencoded`. 

It is only useful for HTTP encoding but it is a very important feature for some legacy, non-json and non-xml HTTP forms. It shouldn't cause any trouble because it will use this encoder only for requests with explicit `application/x-www-form-urlencoded` Content-Type.

multipart/form-data may need some special treating but it would be good to at least have this application/x-www-form-urlencoded in for now.

### DIFF
--- a/http/encoding.go
+++ b/http/encoding.go
@@ -10,6 +10,8 @@ import (
 	"mime"
 	"net/http"
 	"strings"
+
+	"github.com/ajg/form"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


raphael:
Agree this is important to support in some cases. However it would be good to avoid adding this dependency to goa proper (you'll note that all other encoders are standard library encoders). Instead the idea would be for your code to implement and use this method.

raphael:
That being said this would make a great example!

PR #1904, opened 2018-11-01
Set tag value when decoding response

nitinmohan87:


raphael:
Looks great, thank you!

PR #1910, opened 2018-11-10
Plugins pre-generation function

difengo:
Plugins are today executed after the generation of the files in memory. Then, adding an element dynamically to the design roots has no impact on the already generated sections.

The changes of this PR enable the execution of an optional PreGenerate function to modify the design roots with the following signature:

``` go
// PreGenerateFunc makes it possible to modify the design roots before
// the files being generated by the goa code generators or other plugins.
PreGenerateFunc func(genpkg string, roots []eval.Root) error
```

This function is optional but a func or nil must be passed during plugin registration. So, the registration of existing plugins will have to be changed:

From:
``` go
// Register the plugin Generator functions.
func init() {
	codegen.RegisterPlugin(pluginName, "gen", Generate)
	codegen.RegisterPlugin(pluginName, "example", Example)
}
```

To:
``` go
// Register the plugin Generator functions.
func init() {
	codegen.RegisterPlugin(pluginName, "gen", Generate, PreGenerate)
	codegen.RegisterPlugin(pluginName, "example", Example, nil)
}
```


raphael:
Thanks for the PR, can you provide a use case where this is necessary?

difengo:
I'm working on a Prometheus plugin. This would ease the automatic addition of a "metrics" endpoint to the service. This could also help in creating some custom DSL validations or formatting for instance.

raphael:
Thank you that makes sense. Do you think you could change it so that the name of the function is "Prepare" instead of "PreGenerate" (sorry I know it's a little bit nitpicky but names matter :) ). Also it seems to me that having the "prepare" function be the first in the list given to `RegisterPlugin` would be more intuitive. Thank you!

PR #1912, opened 2018-11-13
Split of example service-main section into smaller ones

difengo:
Generated main example file mainly consists of one section (service-main) at the moment.

This PR splits this section into smaller ones to enable plugins to insert code between 2 sections or modify one of the new sections without impacting the others.

This PR has an impact on the CORS plugin as it's referencing the "service-main" section directly to update the binded endpoints definitions. This can be fixed by replacing "service-main" by "service-main-start" for instance as all the new sections are using the same input object.


raphael:
This is great, thank you!

PR #1913, opened 2018-11-20
[BREAKING] dsl package refactor

nitinmohan87:
This refactor makes the service design file import only the top-level `dsl` package instead of top-level `design` and `dsl` and http `design` and `dsl` packages.

### Summary of changes
* Move functions in `http/dsl` package into top-level `dsl` package.
* Move functions in `design` and `http/design` into `expr` package.
* Rename goa v2 `Metadata` to `Meta` (gRPC has a concept of Metadata which will conflict with goa v2 Metadata. Hence this change.)

### What actions should the users take?
* Rename any `Metadata` DSL in your design package to `Meta`
* Import only the top-level dsl package `goa.design/goa/dsl` in your design file. Example:
```
import . "goa.design/goa/dsl"
```
instead of
```
```
If your design uses any plugins, the plugin DSL functions must be called using the package name. Example:
```
import . "goa.design/goa/dsl"
import "goa.design/plugins/cors"  // Use CORS plugin

  Description("The calc service exposes public endpoints that defines CORS policy.")
  cors.Origin("http://127.0.0.1", func() {
    cors.Headers("X-Shared-Secret")
    cors.Methods("GET", "POST")
    cors.Expose("X-Time")
    cors.MaxAge(600)
    cors.Credentials()
  })
})
```


dsl/api.go
```diff
@@ -63,29 +62,43 @@ func API(name string, fn func()) *design.APIExpr {
...
-	if a, ok := eval.Current().(*design.APIExpr); ok {
-		a.Version = ver
+	if s, ok := eval.Current().(*expr.APIExpr); ok {
+		s.Version = ver
+		return
+	}
+	eval.IncompatibleDSL()
+}
+
+// Contact sets the API contact information.
```
raphael: Should we add header documentation for these?

dsl/api.go
```diff
@@ -108,97 +121,30 @@ func Version(ver string) {
...
-//
-// TermsOfService takes a single argument which is the TOS text or URL.
-//
-// Example:
-//
-//    var _ = API("github", func() {
-//        TermsOfService("https://help.github.com/articles/github-terms-of-API/"
-//    })
-//
+// TermsOfService describes the API terms of services or links to them.
```
raphael: Should we keep the header documentation?

dsl/response.go
```diff
@@ -0,0 +1,194 @@
...
+//    - The result type attributes are all mapped to the HTTP response body or gRPC response message.
+//
+// Example:
+//
+//    Method("create", func() {
+//        Payload(CreatePayload)
+//        Result(CreateResult)
+//        Error("an_error")
+//
+//        HTTP(func() {
```
raphael: Would be cool to add a `GRPC` call as well.

examples/cellar/gen/http/storage/client/types.go
```diff
@@ -285,7 +285,7 @@ func NewShowNotFound(body *ShowNotFoundResponseBody) *storage.NotFound {
 	return v
 }
 
-// Validate runs the validations defined on ShowNotFoundResponseBody
+// Validate runs the validations defined on show_not_foundResponseBody
```
raphael: That doesn't look right.... Is that solved in a upcoming PR? or maybe examples need to be regenerated?

examples/cellar/gen/http/storage/server/types.go
```diff
@@ -293,7 +293,7 @@ func NewMultiUpdatePayload(body *MultiUpdateRequestBody, ids []string) *storage.
 	return v
 }
 
-// Validate runs the validations defined on AddRequestBody
+// Validate runs the validations defined on addRequestBody
```
raphael: That doesn't look right.... Is that solved in an upcoming PR? or maybe examples need to be regenerated?
(skipping other occurrences of this problem)

http/middleware/log.go
```diff
@@ -15,7 +15,7 @@ type (
 	Logger interface {
 		// Log creates a log entry using a sequence of alternating keys
 		// and values.
-		Log(keyvals ...interface{})
+		Log(keyvals ...interface{}) error
```
raphael: Do you understand why this change? seems we return nil everywhere?

dsl/response.go
nitinmohan87: I thought I would add it along with the gRPC PR.

http/middleware/log.go
nitinmohan87: If we use goakit plugin, it will replace the stdlib Logger with gokit's Logger which implements the Logger interface like above. So, we can simply use the gokit logger along with this middleware instead of having a separate adapter for it.

nitinmohan87: Right here - https://github.com/goadesign/plugins/pull/61/files#diff-9aaf94aa7f78e80b8d2568dbc56aef4aR103

dsl/api.go
```diff
@@ -63,29 +62,43 @@ func API(name string, fn func()) *design.APIExpr {
...
-//
-// Version accepts a single string argument.
-//
-// Example:
-//
-//    var _ = API("divider", func() {
-//        Version("1.0")
-//    })
-//
+// Version specifies the API version. One design describes one version.
```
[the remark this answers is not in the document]
nitinmohan87: Fixed

dsl/api.go
nitinmohan87: Fixed

dsl/api.go
nitinmohan87: Fixed

http/middleware/log.go
raphael: Ah yes thank you!

PR #1926, opened 2018-12-06
Refactor example generation to consume server expression

nitinmohan87:
This PR modifies the server and CLI flags:
1. `host` flag to specify one of the hosts defined in Server expression
2. URL variables defined in Server expression are added as flags

On server side example
1. `domain` flag is added to override the domain specified in Server expression
2. `http-port` flag to specify port on which the HTTP port will listen overriding the port specified in design.


codegen/server/example_cli.go
```diff
@@ -0,0 +1,222 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"goa.design/goa/codegen"
+	"goa.design/goa/expr"
+)
+
+// ExampleCLIFiles returns example client tool main implementation for each
+// server expression in the design.
```
raphael: Can you remind me why we decided to generate multiple main - one per server - vs. having multiple command line flags? might be good to put that explanation in the comments too.

codegen/service/example_svc.go
```diff
@@ -0,0 +1,141 @@
...
+	"path/filepath"
+	"strings"
+
+	"goa.design/goa/codegen"
+	"goa.design/goa/expr"
+)
+
+type (
+	// basicEndpointData contains the data needed to render a basic endpoint
+	// implementation in the exanoke service file.
```
raphael: typo

codegen/generator/example.go
```diff
@@ -18,21 +19,37 @@ func Example(genpkg string, roots []eval.Root) ([]*codegen.File, error) {
 			continue // could be a plugin root expression
 		}
 
-		// Auth
-		f := service.AuthFuncsFile(genpkg, r)
-		if f != nil {
+		// example service implementation
+		if fs := service.ExampleServiceFiles(genpkg, r); len(fs) != 0 {
```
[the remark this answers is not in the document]
nitinmohan87: I have changed these `Example..` functions to skip `nil` files. So we are either left with an empty slice or valid example file.

codegen/service/example_svc.go
nitinmohan87: Ha ha.. That's really bad one.

codegen/server/example_cli.go
nitinmohan87: Wasn't that your original intention behind using the server expression in generating example mains? One for each server expression. The hosts and URI variables within each server expression will get their own CLI args.

raphael: Duh yeah you are right. One server generated per server expression :)

difengo:
In the "example_scv.go" file, a specific struct named "basicEndpointData" is used to hold the data. But this struct is private. This creates an issue for plugins that need to read the values from the data before modifying the generated endpoint. Would it be possible to make it public instead ?

PR #1927, opened 2018-12-07
[BREAKING] Refactor validation logic

nitinmohan87:
# Summary of changes
* Add a helper interface in validation logic to determine if an attribute is a pointer. In gRPC, the primitive attribute types are always non-pointers when protobuf code is generated.
* Remove `Validate()` function defined on a user type. Instead have a `Validate<user-type-name>` function. Once we have gRPC support, this change would be useful to define validation on a gRPC protobuf type in an external package.

# Actions users must take
If you don't use the `.Validate()` functions outside of the generated code, then no changes necessary. If you use them outside of the generated code, change them to `Validate<user-type-name>(<arg>)`.

### DIFF
--- a/codegen/validation.go
+++ b/codegen/validation.go
@@ -10,6 +10,48 @@ import (
 	"goa.design/goa/expr"
 )
 
+type (
+	// AttributeHelper provides an interface for figuring out common attribute
+	// characteristics.
+	AttributeHelper interface {
+		IsPointer(att *expr.AttributeExpr, required, pointer, useDefault bool) bool
+	}
+
+	defaultH struct{}
+)
+
+// newGoAttributeHelper returns a new Go attribute helper.
+func newAttributeHelper() AttributeHelper {
+	return &defaultH{}
+}
+
+// IsPointer returns if the given attribute is a pointer.
+//
+// req indicates whether the attribute is required (true) or optional (false) in
+// which case target is a pointer if ptr is false and def is false or def is
+// true and there's no default value.
+//
+// ptr indicates whether the data structure described by att uses pointers to
+// hold all attributes (even required ones) so that they may be properly
+// validated.
+//
+// def indicates whether non required attributes that have a default value should
+// not be pointers.
+//
+//    req | ptr | def | pointer?
+//     T  |  F  |  F  | F
+//     T  |  F  |  T  | F
+//     T  |  T  |  F  | T
+//     T  |  T  |  T  | T
+//     F  |  F  |  F  | T
+//     F  |  F  |  T  | F if has default value, T otherwise
+//     F  |  T  |  F  | T
+//     F  |  T  |  T  | T
+//
+func (g *defaultH) IsPointer(att *expr.AttributeExpr, required, pointer, useDefault bool) bool {
+	return pointer || (!required && (att.DefaultValue == nil || !useDefault))
+}
+
 var (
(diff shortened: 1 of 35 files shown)


codegen/validation.go
```diff
@@ -10,6 +10,48 @@ import (
...
+
+// newGoAttributeHelper returns a new Go attribute helper.
+func newAttributeHelper() AttributeHelper {
+	return &defaultH{}
+}
+
+// IsPointer returns if the given attribute is a pointer.
+//
+// req indicates whether the attribute is required (true) or optional (false) in
+// which case target is a pointer if ptr is false and def is false or def is
```
raphael: Description doesn't match name of arguments.

codegen/validation.go
```diff
@@ -10,6 +10,48 @@ import (
 	"goa.design/goa/expr"
 )
 
+type (
+	// AttributeHelper provides an interface for figuring out common attribute
```
raphael: How about `AttributeAnalyzer` instead as a name? `Helper` is a bit generic...

codegen/validation.go
```diff
@@ -41,67 +83,23 @@ func init() {
...
-	}
-	return false
-}
-
 // ValidationCode produces Go code that runs the validations defined in the
 // given attribute definition if any against the content of the variable named
 // target. The generated code assumes that there is a pre-existing "err"
 // variable of type error. It initializes that variable in case a validation
 // fails.
 //
```
raphael: Please describe the role of the `ah` argument (which should now be `an` I guess).

codegen/validation.go
```diff
@@ -197,34 +195,22 @@ func ValidationCode(att *expr.AttributeExpr, req, ptr, def bool, target, context
...
-				continue // break infinite recursions
-			}
-			seen[catt.Attribute] = struct{}{}
-			if hasValidationsRecurse(catt.Attribute, ignoreRequired, seen) {
-				return true
-			}
-		}
+// the variable named target.
+//
+// ahs is the optional AttributeHelper to use to compute validation. If none
```
raphael: Would it be cleaner to make `ahs` just a normal - non variadic argument and have callers instantiate the default attribute analyzer if non other is given?

codegen/validation.go
```diff
@@ -197,34 +195,22 @@ func ValidationCode(att *expr.AttributeExpr, req, ptr, def bool, target, context
...
+	} else {
+		ah = newAttributeHelper()
 	}
-	return false
+	seen := make(map[string]*bytes.Buffer)
+	return recurseValidationCode(att, req, ptr, def, target, target, seen, ah).String()
 }
 
-func recurseValidationCode(att *expr.AttributeExpr, req, ptr, def bool, target, context string, seen map[string]*bytes.Buffer) *bytes.Buffer {
+func recurseValidationCode(att *expr.AttributeExpr, req, ptr, def bool, target, context string, seen map[string]*bytes.Buffer, ah AttributeHelper) *bytes.Buffer {
```
raphael: Would be good to keep `seen` as the last argument - it's been the convention in the goa source code.

nitinmohan87: Done

codegen/validation.go
nitinmohan87: Fixed

raphael:
Looks good, thank you!

PR #1928, opened 2018-12-07
Refactor transform to support gRPC protobuf transforms

nitinmohan87:


codegen/attribute_analyzer.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"goa.design/goa/expr"
+)
+
+type (
+	// AttributeAnalyzer analyzes an attribute and its sub-attributes.
+	AttributeAnalyzer interface {
+		// IsPointer returns true if the attribute being analyzed is a pointer.
+		IsPointer() bool
+		// IsRequired returns true if the Required property is set to true.
```
raphael: I don't understand that comment as the Required field is a list of names in the parent attribute? which "Required" property is this referring to?

codegen/attribute_analyzer.go
```diff
@@ -0,0 +1,191 @@
+package codegen
+
+import (
+	"goa.design/goa/expr"
+)
+
+type (
+	// AttributeAnalyzer analyzes an attribute and its sub-attributes.
```
raphael: This is a big interface which is not idiomatic. I wonder if we couldn't just use structs instead but need to understand better the interactions.

codegen/attribute_analyzer.go
```diff
@@ -0,0 +1,191 @@
...
+		// SetAttribute sets the attribute to analyze.
+		SetAttribute(att *expr.AttributeExpr)
+		// SetProperties sets the attribute properties in the analyzer.
+		SetProperties(required, pointer, defPtr, useDefault bool)
+	}
+
+	// AttributeProperties contains properties that affect how the attribute
+	// is stored (i.e. pointer or non-pointer).
+	AttributeProperties struct {
+		// Pointer if true indicates that the attribute is a pointer even if
```
raphael: Pointer indicates whether the attribute is a pointer ...

Same with comments below.

codegen/attribute_analyzer.go
```diff
@@ -0,0 +1,191 @@
...
+//
+//    Pointer | DefaultPointer | Required | IsPointer
+//       T    |       T        |     T    |     T
+//       T    |       F        |     F    |     T
+//       T    |       T        |     F    |     T
+//       T    |       F        |     T    |     T
+//       F    |       T        |     F    |     T
+//       F    |       T        |     T    |     F
+//       F    |       F        |     T    |     F
+//       F    |       F        |     F    |     F if default value exists and UseDefault is T, else T otherwise
```
raphael: remove "else"

codegen/go_transform.go
```diff
@@ -0,0 +1,652 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+	"text/template"
+
+	"goa.design/goa/expr"
+)
+
+var (
+	// TransformGoArrayT is the template to generate Go array transformation
```
raphael: Do these need to be public? can we get rid of the global variables altogether?

codegen/go_transform.go
```diff
@@ -0,0 +1,652 @@
...
+		"loopVar":            arrayLoopVar,
+	}).Parse(transformGoArrayTmpl))
+	TransformGoMapT = template.Must(template.New("transformGoMap").Funcs(template.FuncMap{
+		"transformAttribute": transformAttributeHelper,
+		"loopVar":            mapLoopVar,
+	}).Parse(transformGoMapTmpl))
+}
+
+type (
+	// GoTransformer defines the fields to generate Go code when transforming
```
raphael: GoTransformer is a Transformer that generates Go code for converting a struct represented as an attribute expression into a different struct also represented as a an attribute expression.

(please explain what the rules for the conversions are at a high level and give an example of whether this is used).

codegen/go_transform.go
```diff
@@ -0,0 +1,652 @@
...
+	// a source attribute to a target attribute. It implements the Transformer
+	// interface.
+	GoTransformer struct {
+		*AttributeTransformer
+	}
+)
+
+// NewGoTransformer returns a new transformer that generates Go code during
```
raphael: NewGoTransformer creates a GoTransfomer that uses prefix to .... (no need to redefine what a transformer is here)

codegen/scope.go
```diff
@@ -93,22 +93,30 @@ done:
 
 // GoTypeDef returns the Go code that defines a Go type which matches the data
 // structure definition (the part that comes after `type foo`).
-func (s *NameScope) GoTypeDef(att *expr.AttributeExpr, useDefault bool) string {
+//
+// ptr if true indicates that the attributes (even the required ones) are
```
raphael: Fix comment

codegen/service/testdata/convert_functions.go
```diff
@@ -109,26 +107,22 @@ func (t *ArrayStringType) ConvertToArrayStringT() *testdata.ArrayStringT {
 var ConvertObjectCode = `// ConvertToObjectT creates an instance of ObjectT initialized from t.
 func (t *ObjectType) ConvertToObjectT() *testdata.ObjectT {
 	v := &testdata.ObjectT{}
-	if t.Object != nil {
```
raphael: Is this test not needed anymore?

codegen/service/testdata/convert_functions.go
```diff
@@ -197,19 +191,15 @@ func marshalObjectFieldToObjectFieldTExt(v *ObjectField) *testdata.ObjectFieldT
...
+		res.Array[i] = val
 	}
-	if v.Map != nil {
-		res.Map = make(map[string]bool, len(v.Map))
-		for key, val := range v.Map {
-			tk := key
-			tv := val
-			res.Map[tk] = tv
-		}
+	res.Map = make(map[string]bool, len(v.Map))
```
raphael: This code (same above) will create a non-nil attribute instead of keeping it nil if the original is nil.

codegen/transformer.go
```diff
@@ -0,0 +1,195 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"goa.design/goa/expr"
+)
+
+type (
+	// Transformer transforms a source attribute to a target attribute.
```
raphael: What is a transformation? also fix the English :)
Please explain carefully what this interface represents, how it's used etc. as it's a pretty fundamental one. Also it's a big interface which isn't great and a bit of a code smell - would be good to review and understand the need for it at a higher level.

codegen/transformer.go
```diff
@@ -0,0 +1,195 @@
...
+	TransformAttrs struct {
+		// SourceVar and TargetVar are the source and target variable names used
+		// in the transformation code.
+		SourceVar, TargetVar string
+		// NewVar is used to determine the assignment operator to initialize
+		// TargetVar.
+		NewVar bool
+	}
+
+	// AttributeTransformer defines the fields to transform a source attribute
```
raphael: I don't understand the comment.

codegen/transformer.go
```diff
@@ -0,0 +1,195 @@
...
+	// Transforming this type requires generating an intermediary function:
+	//
+	//     func recursiveToRecursive(r *Recursive) *service.Recursive {
+	//         var t service.Recursive
+	//         if r.R != nil {
+	//             t.R = recursiveToRecursive(r.R)
+	//         }
+	//    }
+	//
+	TransformFunctionData struct {
```
raphael: Do all these need to be public?

codegen/transformer.go
```diff
@@ -0,0 +1,195 @@
...
+		Name          string
+		ParamTypeRef  string
+		ResultTypeRef string
+		Code          string
+	}
+)
+
+// Transform transforms source attribute to target attribute with the given
```
raphael: Please explain what a transformation is and fix the English.

examples/cellar/gen/http/storage/server/types.go
```diff
@@ -229,7 +229,9 @@ func NewAddBottle(body *AddRequestBody) *storage.Bottle {
 		Description: body.Description,
 		Rating:      body.Rating,
 	}
-	v.Winery = unmarshalWineryRequestBodyToWinery(body.Winery)
+	if body.Winery != nil {
```
raphael: Are these new "if" statement expected? was the code wrong before?

grpc/docs/FAQ.md
```diff
@@ -0,0 +1,51 @@
+# How are default values evaluated in protocol buffers?
+
+Non-nil default values are not supported in protocol buffers
+(see https://developers.google.com/protocol-buffers/docs/proto3#default).
+Hence, there is no way to figure out whether a field was explicitly set to
+the default value or just not set at all. So goa does not initialize such
+fields with their default values.
+
+# How goa deals with nested maps and arrays in protocol buffers?
```
raphael: How does goa deal with ...

grpc/docs/FAQ.md
```diff
@@ -0,0 +1,51 @@
...
+
+proto3 syntax for protocol buffer does not support nested maps and arrays
```
raphael: The proto3 syntax ...

http/codegen/testdata/parse_endpoint_functions.go
```diff
@@ -1002,13 +1002,6 @@ func BuildMethodMapQueryObjectPayload(serviceMapQueryObjectMethodMapQueryObjectB
 		if err != nil {
 			return nil, fmt.Errorf("invalid JSON for c, example of valid JSON:\n%s", "'{\n      \"1484745265794365762\": [\n         \"Similique aspernatur.\",\n         \"Error explicabo.\",\n         \"Minima cumque voluptatem et distinctio aliquam.\",\n         \"Blanditiis ut eaque.\"\n      ],\n      \"4925854623691091547\": [\n         \"Eos aut ipsam.\",\n         \"Aliquam tempora.\"\n      ],\n      \"7174751143827362498\": [\n         \"Facilis minus explicabo nemo eos vel repellat.\",\n         \"Voluptatum magni aperiam qui.\"\n      ]\n   }'")
 		}
-		err = goa.MergeErrors(err, goa.ValidatePattern("c.a", c.A, "patterna"))
```
raphael: Is this not needed anymore?

http/codegen/testdata/payload_decode_functions.go
```diff
@@ -3178,11 +3178,8 @@ var PayloadBodyNestedUserDecodeCode = `// DecodeMethodBodyUserRequest returns a
 func DecodeMethodBodyUserRequest(mux goahttp.Muxer, decoder func(*http.Request) goahttp.Decoder) func(*http.Request) (interface{}, error) {
 	return func(r *http.Request) (interface{}, error) {
 		var (
-			body struct {
-				A *string
-				B *string
-			}
-			err error
+			body NestedType
```
raphael: Wondering whether using `body *NestedType` and passing `body` instead of `&body` would be better. This would follow the convention we've used elsewhere better.

http/codegen/testdata/result_encode_functions.go
```diff
@@ -442,13 +442,9 @@ var ResultHeaderBoolDefaultEncodeCode = `// EncodeMethodHeaderBoolDefaultRespons
...
 	return func(ctx context.Context, w http.ResponseWriter, v interface{}) error {
 		res := v.(*serviceheaderbooldefault.MethodHeaderBoolDefaultResult)
-		if res.H != nil {
-			val := res.H
-			hs := strconv.FormatBool(*val)
-			w.Header().Set("h", hs)
-		} else {
-			w.Header().Set("h", "true")
-		}
+		val := res.H
```
raphael: The test is not needed anymore?

http/codegen/testdata/result_encode_functions.go
```diff
@@ -476,11 +472,7 @@ var ResultHeaderStringDefaultEncodeCode = `// EncodeMethodHeaderStringDefaultRes
 func EncodeMethodHeaderStringDefaultResponse(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
 	return func(ctx context.Context, w http.ResponseWriter, v interface{}) error {
 		res := v.(*serviceheaderstringdefault.MethodHeaderStringDefaultResult)
-		if res.H != nil {
```
raphael: was there a bug before?

http/codegen/service_data.go
```diff
@@ -1205,17 +1208,19 @@ func buildPayloadData(e *expr.HTTPEndpointExpr, sd *ServiceData) *PayloadData {
...
-	if init == nil {
-		if o := expr.AsObject(e.Params.Type); o != nil && len(*o) > 0 {
-			returnValue = codegen.Goify((*o)[0].Name, false)
-		} else if o := expr.AsObject(e.Headers.Type); o != nil && len(*o) > 0 {
-			returnValue = codegen.Goify((*o)[0].Name, false)
-		} else if e.MapQueryParams != nil && *e.MapQueryParams == "" {
-			returnValue = mapQueryParam.Name
+	{
+		if payload.Type != expr.Empty {
+			name = payloadAn.Name(true)
```
raphael: This is an example where I'm not sure the abstraction helps. I'd rather read `svc.Scope.GoFullTypeName` than `payloadAn.Name` as the former is clear while the latter has me go through extra code to understand what's happening.

codegen/go_transform.go
```diff
@@ -38,17 +47,58 @@ func NewGoTransformer(prefix string) Transformer {
...
+// if source and target are not compatible for transformation.
+func (g *GoTransformer) TransformPrimitive(source, target AttributeAnalyzer, ta *TransformAttrs) (string, error) {
+	fmt.Println("go")
+	var code string
+	assign := "="
+	if ta.NewVar {
+		assign = ":="
+	}
+	if _, ok := target.Attribute().Type.(expr.UserType); ok {
+		// Primitive user type, these are used for error results
```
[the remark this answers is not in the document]
nitinmohan87: They are the only ones

http/codegen/testdata/payload_decode_functions.go
nitinmohan87: We initialize non-pointer `body` in other places too.

http/codegen/testdata/result_encode_functions.go
nitinmohan87: We cannot set default values when sending server response because the attributes with default values are stored as non-pointers in the result type. Hence we won't know if the attribute is set by the user or not.

http/codegen/testdata/result_encode_functions.go
nitinmohan87: Yes. See comment above.

examples/cellar/gen/http/storage/server/types.go
nitinmohan87: `body.Winery` is a pointer field. I think it would be safer to check for nil even though Winery is a required attribute.

codegen/service/testdata/convert_functions.go
nitinmohan87: These are required attributes. So they must be non-nil correct?

examples/cellar/gen/http/storage/client/cli.go
```diff
@@ -99,9 +99,7 @@ func BuildAddPayload(storageAddBody string) (*storage.Bottle, error) {
 		Description: body.Description,
 		Rating:      body.Rating,
 	}
-	if body.Winery != nil {
-		v.Winery = marshalWineryRequestBodyToWinery(body.Winery)
-	}
+	v.Winery = marshalWineryRequestBodyToStorageWinery(body.Winery)
```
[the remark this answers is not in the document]
nitinmohan87: Yes. It is required.

codegen/service/testdata/convert_functions.go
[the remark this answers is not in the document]
nitinmohan87: It is a required attribute.

codegen/go_transform.go
```diff
@@ -0,0 +1,693 @@
...
+		"TargetElem":  target,
+		"SourceVar":   ta.SourceVar,
+		"TargetVar":   ta.TargetVar,
+		"NewVar":      ta.NewVar,
+	}
+	return RunGoArrayTemplate(data)
+}
+
+// RunGoArrayTemplate runs the template to generate Go array code.
+func RunGoArrayTemplate(data map[string]interface{}) (string, error) {
```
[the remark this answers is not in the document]
nitinmohan87: Yes. The protobuf transformation code uses this func to generate array code.

codegen/go_transform.go
```diff
@@ -0,0 +1,693 @@
...
+// RunGoMapTemplate runs the template to generate Go map code.
+func RunGoMapTemplate(data map[string]interface{}) (string, error) {
+	var buf bytes.Buffer
+	if err := transformGoMapT.Execute(&buf, data); err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// GoAttribute represents an attribute type that produces Go code.
```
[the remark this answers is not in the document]
nitinmohan87: Yes. Protobuf attribute struct embeds this and overrides a couple of functions.

codegen/service/service_data.go
```diff
@@ -364,6 +363,28 @@ type (
 	}
 )
 
+// TypeContext returns a contextual attribute for service types.
+// Service types are Go types and uses non-pointers to hold attributes
+// having default values.
+func TypeContext(att *expr.AttributeExpr, pkg string, scope *codegen.NameScope) *codegen.ContextualAttribute {
```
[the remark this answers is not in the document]
nitinmohan87: It is used in the http service data to get a contextual attribute for payload and result types.

PR #1929, opened 2018-12-08
Fix body name conflict with multiple services having same method name

nitinmohan87:
In OpenAPI, when you have multiple services having the same method name and raw object Result type, the response body names generated collides. This PR prefixes such response body names with the service name.


raphael:
Would it be possible to add a test that shows how the generated code changes?

raphael:
Thank you!

Issue #1914, opened 2018-11-23
support building code outside of GOPATH for go module support


jtnord:
If I have a project that is using [Go Modules](https://golang.org/doc/go1.11#modules) (as opposed to the old `GOPATH `layout) I would still like to be able to use goagen.  Alas it seems goagen will not work at all unless you are using the legacy GOPATH layout.

## Expected Behavior

`goagen app -d ${PACKAGE_NAME}/${DESIGN_DIR} -o pkg/controller` works with no error.

## Actual Behavior

goagen errors with

`Go source file "C:\workarea\source\github\someorg\some-repo\goagen053437807\design" not in Go workspace, adjust GOPATH C:\PATH\to\go`

## Specifications

  - goa Version: goagen v1.3.1
  - Platform (Linux/Windows etc.): Windows 10

raphael:
Thank you for the report. At this point I'm focusing on v2 and will make sure it supports Go module. I'd be happy to accept a PR that added Go module support to v1 though.

raphael:
Go modules are now supported in v1 (thank you @tchssk!). Please open a new issue if things don't work out as expected.

PR #1930, opened 2018-12-10
Use build.Default.GOPATH instead of env GOPATH

tchssk:
Go v1.8+ set a default GOPATH if the GOPATH env variable is not set. This patch makes it possible to use `goagen` with a default GOPATH.

I found this issue during trying to improve goa to support Go modules (for #1914). I'm plannning to open another pull request for that.

### DIFF
--- a/design/random.go
+++ b/design/random.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/manveru/faker"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 )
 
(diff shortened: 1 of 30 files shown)


tchssk:
Travis says

>/home/travis/gopath/src/github.com/goadesign/goa/metrics.go
/home/travis/gopath/src/github.com/goadesign/goa/metrics_test.go
^ - Repo contains improperly formatted go files

but the files look OK to me...

raphael:
This is great! thank you.

PR #1932, opened 2018-12-15
Use yaml.Unmarshal for swagger.yaml

blp1526:
Converting JSON to YAML, we need yaml.Unmarshal to keep the right number type.

via https://github.com/ghodss/yaml/blob/f5bccc04aa10faf16af8a8c9664a452bf7d80101/yaml.go#L65-L81

An example is here.

```
apidsl.Attribute("id", design.Integer, "Unique bottle ID", func() {
  apidsl.Example(1234567)
})
```

Using json.Unmarshal, goagen swagger generates this swagger.yaml.

```json
id:
  description: Unique bottle ID
  example: 1.234567e+06
  format: int64
  type: integer
```

After using yaml.Unmarshal, goagen swagger generates this swagger.yaml.

```json
id:
  description: Unique bottle ID
  example: 1234567
  format: int64
  type: integer
```

### DIFF
--- a/goagen/gen_swagger/generator.go
+++ b/goagen/gen_swagger/generator.go
@@ -122,7 +122,7 @@ func (g *Generator) Cleanup() {
 
 func jsonToYAML(rawJSON []byte) ([]byte, error) {
 	var yamlSource interface{}
-	if err := json.Unmarshal(rawJSON, &yamlSource); err != nil {
+	if err := yaml.Unmarshal(rawJSON, &yamlSource); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


raphael:
This is great! Do you think you could write a test that would fail without this change and add it to `generator_test.go`? thank you!

blp1526:
@raphael

Thank you for your review.
I add two commits.
e299f7231512b1184a3bdf9bbce79fd61f3c3665 has a test which fails by json.Unmarshal.
d62bf944ac8dba682eb8e14c4a8607dd0af7c235 passes a test by using yaml.Unmarshal.

raphael:
This is great, thank you! One last ask :) - could you backport this fix to the `v1` branch please? (simply cherry-pick the merge commit into a new branch created off of `v1`).

PR #1933, opened 2018-12-16
Support Go 1.11 Modules by goagen

tchssk:
#1914

### DIFF
--- a/goagen/codegen/workspace.go
+++ b/goagen/codegen/workspace.go
@@ -406,7 +406,7 @@ func PackageName(path string) (string, error) {
 	return pkgNames[0], nil
 }
 
-// Copeied from cmd/go/internal/modload/init.go.
+// Copied from cmd/go/internal/modload/init.go.
 var altConfigs = []string{
 	"Gopkg.lock",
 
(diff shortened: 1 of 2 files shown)


goagen/codegen/workspace.go
```diff
@@ -368,6 +406,107 @@ func PackageName(path string) (string, error) {
 	return pkgNames[0], nil
 }
 
+// Copeied from cmd/go/internal/modload/init.go.
```
raphael: typo

tchssk: L425 had a same typo. I fixed both.

raphael:
Thanks!

raphael:
I just realized that the tests don't pass on Windows  - see [appveyor](https://ci.appveyor.com/project/RaphaelSimon/goa-oqtis/branch/master). @tchssk do you think you could take a look? thank you!

raphael:
Also this should probably be ported to v1.

raphael:
Fixed in #1974

tchssk:
Thank you for fixing the tests.

* Has #1930 been applied onto v1? It should be ported too.
* The new version contains some new features. Is it not v1.5.0 but 1.4.1?

raphael:
hmm yes #1930 should be ported, good catch!
The only new feature really is support for Go modules I think? it didn't seem like that justified a minor version upgrade.

raphael:
I have ported #1930 to v1 and retagged v1.4.1, thank you!

tchssk:
OK. I agree to v1.4.1.

tchssk:
BTW you should announce v1.4.1 on Twitter. @raphael

PR #1952, opened 2019-01-07
Check bases and references in validation phase (#1943)

### LINKED ISSUE #1943: v2: cannot build a payload only by using Extend
Using a design that only uses `Extend` in payload without defining any explicit attributes fails to generate.
```
package design

import (
	. "goa.design/goa/dsl"
)

var _ = API("Something", func() {
})

	Token("bar", String)
})

var _ = Service("Foo", func() {
	Error("bad_request")

	HTTP(func() {
		Path("/foo")
		Response("bad_request", StatusBadRequest)
	})

	Method("create", func() {
		Payload(func() {
			Extend(CommonAttributes)
			Extend(MyPayload)
			// Attribute("name", String) <- adding this line makes it work
			Required("name")
		})
		Result(func() {
			Attribute("location", String)
			Required("location")
		})

		HTTP(func() {
			POST("/")
			Response(StatusCreated, func() {
				Header("location:Location")
			})
		})
	})
})

var MyPayload = Type("MyPayload", func() {
	Attribute("name", String)
	Attribute("description", String)
})
```
fails with
```
service "Foo" HTTP endpoint "create": Params are set but Payload is not defined.
```
even if I add a single attribute in the payload, everything works as expected.

nitinmohan87:
Fixes #1943

### DIFF
--- a/expr/http_endpoint.go
+++ b/expr/http_endpoint.go
@@ -329,7 +329,7 @@ func (e *HTTPEndpointExpr) Validate() error {
 	}
 
 	// Validate definitions of params, headers and bodies against definition of payload
-	if e.MethodExpr.Payload.Type == Empty {
+	if isEmpty(e.MethodExpr.Payload) {
 		if e.MapQueryParams != nil {
 			verr.Add(e, "MapParams is set but Payload is not defined")
 		}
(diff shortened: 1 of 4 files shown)


raphael:
Thank you Nitin! do you think you could add a test that fails without this fix but passes with it?

PR #1954, opened 2019-01-09
Add support for gRPC expressions

nitinmohan87:
gRPC support Part 1 of 2.

This PR contains only the dsls and expressions for gRPC. Code generation will be issued in a separate PR.

### DIFF
--- a/dsl/doc.go
+++ b/dsl/doc.go
@@ -77,5 +77,74 @@ The example above can thus be simplified to:
         })
     })
 
+The GRPC DSL adds a "GRPC" function to the DSL constructs that require gRPC
+specific information. These include the API, Service, Method, and Error DSLs.
+
+For example:
+
+    var _ = API("name", func() {
+        Description("Optional description")
+        // gRPC specific properties
+        GRPC(func() {
+        })
+    })
+
+The GRPC function defines the mapping of the data type attributes used in the
+generic DSL to gRPC messages and metadata.
+
+For example:
+
+    var PayloadType = Type("Payload", func() {
+        TypeName("Payload")        // mapped to gRPC message with name "Payload"
+        Field(1, "rq1", String)    // mapped to field in "Payload" message
+        													 // with name "rq1" and tag number 1
+        Field(2, "rq2", String)    // mapped to field in "Payload" message
+        													 // with name "rq2" and tag number 2
+        Attribute("rq3", Int)
+        Attribute("rq4", Int)
+    })
+
+    var ResultType = ResultType("application/vnd.result", func() {
+        TypeName("Result")         // mapped to gRPC message with name "Result"
+        Attributes(func() {
+           Attribute("rp1", Int)
+           Field(1, "rp2", String) // mapped to field in "Result" message
+        													 // with name "rp2" and tag number 1
+        })
+    })
+
+    var _ = Service("name", func() {
+    		Method("name", func() {
+           Payload(PayloadType)
+           Result(ResultType)       // has attributes rp1 and rp2
+           Error("name")
+
+    			 GRPC(func() {
+               Metadata(func() {   // rq3 and rq4 present in gRPC request metadata
+                   Attribute("rq3)
+                   Attribute("rq4")
+               })
+               Message(func() {
+                   Attribute("rq1") // rq1 and rq2 present in gRPC request message
+               })
+               Response(CodeOK, func() {
+                   Metadata(func() {
+                       Attribute("rp1") // rp1 present in gRPC response metadata
+                   })
+                   Message(func() {
+                       Attribute("rp2") // rp2 present in gRPC response message
+                   })
+               })
+               Response("name", CodeInternal)
+           })
+    	  })
+    })
+
+By default the payload and result type attributes define the request and
+response message fields respectively with the exception of security attributes
+in payload which are mapped to request metadata unless specified explicitly.
+The default response status code is CodeOK for success response and
+CodeUnknown for error responses. See google.golang.org/grpc/codes package for
+more information on the status codes.
 */
 package dsl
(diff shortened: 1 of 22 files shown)


dsl/doc.go
```diff
@@ -77,5 +77,74 @@ The example above can thus be simplified to:
...
+        Field(2, "rq2", String)    // mapped to field in "Payload" message
+        													 // with name "rq2" and tag number 2
+        Attribute("rq3", Int)
+        Attribute("rq4", Int)
+    })
+
+    var ResultType = ResultType("application/vnd.result", func() {
+        TypeName("Result")         // mapped to gRPC message with name "Result"
+        Attributes(func() {
+           Attribute("rp1", Int)
```
raphael: Is this legal (attribute with no "tag" metadata)? don't think it should be...

dsl/doc.go
```diff
@@ -77,5 +77,74 @@ The example above can thus be simplified to:
...
+           Field(1, "rp2", String) // mapped to field in "Result" message
+        													 // with name "rp2" and tag number 1
+        })
+    })
+
+    var _ = Service("name", func() {
+    		Method("name", func() {
```
raphael: Indent

dsl/doc.go
```diff
@@ -77,5 +77,74 @@ The example above can thus be simplified to:
...
+               })
+               Response(CodeOK, func() {
+                   Metadata(func() {
+                       Attribute("rp1") // rp1 present in gRPC response metadata
+                   })
+                   Message(func() {
+                       Attribute("rp2") // rp2 present in gRPC response message
+                   })
+               })
+               Response("name", CodeInternal)
```
raphael: Might good to document that this response represents the response message used in case of error for error "name". I forget what we decided about encoding error structs?

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+// the only use of GRPC allowed in the API expression.
+//
+// The functions that appear in GRPC such as Message or Response may take
+// advantage of the request or response types (depending on whether they appear
+// when describing the gRPC request or response). The properties of the message
+// attributes inherit the properties of the attributes with the same names that
+// appear in the request or response types. The functions may also define new
+// attributes or override the existing request or response type attributes.
+//
+// GRPC must appear in a Service or a Method expression.
```
raphael: API as well?

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+		act := res.EndpointFor(actual.Name, actual)
+		act.DSLFunc = fn
+	default:
+		eval.IncompatibleDSL()
+	}
+}
+
+// Message describes a gRPC request or response message.
+//
+// Message must appear in a Method gRPC endpoint expression to define the
```
raphael: Method -> method

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+// attributes that must appear in a request or in a gRPC response expression
```
raphael: fix English

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+// to define the attributes that must appear in a response message.
+// If Message is absent then the request message is built using the method
+// Payload expression and the response message is build using the method
+// Result expression.
+//
+// Message accepts one argument of function type which lists the attributes
+// that must be present in the message. For example, the Message DSL can be
+// defined on the gRPC endpoint expression listing the security attributes
```
raphael: gRPC method expression

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+// Example:
+//
+//     Method("add", func() {
+//         Payload(Operands)
+//         Result(Int)      // method Result is a primitive
+//         GRPC(func() {
+//             Response(CodeOK, func()
+//                 Message(func() {
+//                     Attribute("sum") // Response message has one field with
+//                                      // name "sum" instead of "field"
```
raphael: instead of the default "field".

dsl/grpc.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// Metadata defines a gRPC request metadata.
+//
+// Metadata must appear in a gRPC endpoint expression to describe gRPC request
+// metadata.
+//
+// Security attributes in the method Payload are automatically added to the
+// request metadata unless specified explicitly in request message using
+// Message DSL. All other attributes in method Payload are added to the
```
raphael: using the Message function. I've tried to go away from using the term DSL in favor of just "function" when applicable. Might be good to review all uses of DSL and see if they can be replaced with "function".

dsl/headers.go
```diff
@@ -0,0 +1,100 @@
...
+// to define request headers common to all the service methods. Headers may
+// also appear in a method, response or error HTTP expression to define the
+// HTTP endpoint request and response headers.
+//
+// To define gRPC response header metadata, Headers must appear in a GRPC
+// response expression.
+//
+// Headers accepts one argument: Either a function listing the headers (both
+// HTTP and gRPC) or a user type which must be an object and whose attributes
+// define the headers (only HTTP).
```
raphael: HTTP only

dsl/doc.go
nitinmohan87: Right now, gRPC sends a default error message of this type - https://godoc.org/google.golang.org/grpc/status#Status . We don't really support custom error type yet.

dsl/doc.go
raphael: ok makes sense, thank you.

dsl/doc.go
raphael: So we have a error type to represent that in the design? I seem to recall we said we should simplify the HTTP error type to match that.

raphael:
Thank you, PR looks great!

PR #1962, opened 2019-01-17
add protoc install setting for MacOS

aya-eiya:
`make depend` didn't work on Mac.
So I added darwin case to Install protoc block.


raphael:
Great, thank you!

PR #1963, opened 2019-01-18
Add tests for expr.ServerExpr.Validate()

tchssk:
#1423

### DIFF
--- a/expr/server_test.go
+++ b/expr/server_test.go
@@ -24,6 +24,105 @@ func TestServerExprEvalName(t *testing.T) {
 	}
 }
 
+func TestServerExprValidate(t *testing.T) {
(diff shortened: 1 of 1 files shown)


expr/server_test.go
```diff
@@ -24,6 +24,105 @@ func TestServerExprEvalName(t *testing.T) {
...
+			services: []string{
+				bar,
+			},
+			expected: &eval.ValidationErrors{
+				Errors: []error{
+					errServiceUndefined,
+				},
+			},
+		},
+		"error only both": {
```
raphael: Maybe "error in both" is better?

tchssk: Oops, thank you.

PR #1970, opened 2019-01-24
Generate grpc type constructor for views

nitinmohan87:
When sending a gRPC response that is a result type, the constructor code to transform the result type to the protocol buffer type panics when it dereferences a field that is not in the view. Hence we need to generate per-view constructors to deal with this.


examples/chatter/gen/grpc/cli/chatter/cli.go
```diff
@@ -232,6 +232,6 @@ Returns the chat messages sent to the server.
     -token STRING: 
 
 Example:
-    `+os.Args[0]+` chatter history --view "Et tempora temporibus eligendi." --token "Excepturi quia soluta."
+    `+os.Args[0]+` chatter history --view "Minima voluptatum." --token "Est quibusdam voluptate nulla autem quisquam dolorum."
```
raphael: can we use real view names for the example?

PR #1973, opened 2019-01-28
Use gofrs/uuid instead of satori/go.uuid

raphael:
satori's package isn't being maintained anymore.

### DIFF
--- a/design/random.go
+++ b/design/random.go
@@ -7,8 +7,8 @@ import (
 	"math/rand"
 	"time"
 
+	uuid "github.com/gofrs/uuid"
 	"github.com/manveru/faker"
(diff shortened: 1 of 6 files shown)


raphael:
Does it have a compatible interface?

euller88:
I jumped the gun and misunderstood your actions. I thought that Goa had changed from the old UUID package for an entirely new one, implicating that you modified the project source code. But now, seeing that the new package is a fork of the old one, since it isn't being maintained anymore, your decision makes complete sense.

BTW, the Google package, IMO, is as good as the other, but the interface is not compatible, so you can't change the imports without changing the code.

raphael:
Makes sense, thanks for the pointer though!

PR #1976, opened 2019-01-28
Address backslash problems on Windows

MichaelUrman:
I was attempting to use goa v2 on windows, and ran into a few problems. I've worked around them locally, and wanted to share back, but I'm not entirely sure how to provoke them, as attempting to find minimal steps resulted in slightly different behavior than I encountered.

I am strongly skeptical that my changes are the "right" fix, but determining that will require more familiarity with go and goa than I have, and may result in a much more invasive set of changes. I'm willing to rework my patches, time depending, so please share any feedback. Alternately, I hope this report is enough to isolate what you need. Here are the issues I saw, and further description of what I changed in my commits. 

**Top-level import uses backslash**
go requires forward slashes in import paths. When running on Windows, the use of `filepath.Join` to build what is later used as an import path results in backslashes. I cover for this in 5ac020c with `filepath.ToSlash`, but think a more correct approach may be to use e.g. `path.Join` and avoid introducing backslashes in the first place.

**Example code miscalculates rootPath**
In an odd reversal of the above problem, many of the `*/codegen/example_*.go` files attempt to parse off the base part of a path. In my case, the code was receiving an import path (using forward slashes), but parsing with `os.PathSeparator` (a backslash on Windows). This resulted in a `rootPath` of `"."`, and example code that didn't compile. I worked around this by parsing a copy that I first passed through `filepath.FromSlash`, but I think a more correct approach may be to ensure that the incoming path is always in import form, and use an explicit `"/"` instead of `string(os.PathSeparator)`. These changes are in 72940a0.

**Reproduction steps**
Follow the instructions in the readme for https://github.com/goadesign/goa/tree/v2 through `goa gen calcsvc/design` and `goa example calcsvc/design` on Windows. (I've been testing on Windows 10 1803, but any supported Windows should match.) This results in a calc.go file containing `import ( calcsvc "calcsvc\gen\calc" )` and in a variant I haven't yet isolated how to reproduce, a cmd/calc/http.go file containing `import ( calcsvc "." )` (instead of `calcsvc "calcsvc/gen/calc"`). I had originally tested with another directory layer which reported an error during `goa gen`, and then (after the first fix) an error during `go build` of the server and cli generated by `goa example`.

**Tests**
I'm not sure how to write a test for this that will identify either problem while testing on non-Windows. I was able to add a test for the backslashes (04e0c4b), but it feels tightly coupled to my solution, and only failed on Windows. I'm less clear how to write a test for the imports generated by `goa example`. In getting there, I also tweaked the golden file comparisons (66e406d) as they didn't handle the CRLF newlines introduced by checking out on Windows with whatever gitconfig applied in my case.

### DIFF
--- a/codegen/import.go
+++ b/codegen/import.go
@@ -1,6 +1,9 @@
 package codegen
 
-import "fmt"
+import (
+	"fmt"
+	"path/filepath"
+)
 
 type (
 	// ImportSpec defines a generated import statement.
@@ -25,7 +28,7 @@ func SimpleImport(path string) *ImportSpec {
 // Code returns the Go import statement for the ImportSpec.
 func (s *ImportSpec) Code() string {
 	if len(s.Name) > 0 {
-		return fmt.Sprintf(`%s "%s"`, s.Name, s.Path)
+		return fmt.Sprintf(`%s "%s"`, s.Name, filepath.ToSlash(s.Path))
 	}
-	return fmt.Sprintf(`"%s"`, s.Path)
+	return fmt.Sprintf(`"%s"`, filepath.ToSlash(s.Path))
 }
--- a/codegen/import_test.go
+++ b/codegen/import_test.go
@@ -1,0 +1,31 @@
+package codegen
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestImportPaths(t *testing.T) {
+	var i0 = SimpleImport("simple")
+	var i1 = NewImport("renamed", "direct")
+	var i2 = NewImport("sub", filepath.Join("base", "sub"))
+	var i3 = NewImport("remote", "example.com/something/remote")
+
+	tests := []struct {
+		i        *ImportSpec
+		expected string
+	}{
+		{i0, `"simple"`},
+		{i1, `renamed "direct"`},
+		{i2, `sub "base/sub"`},
+		{i3, `remote "example.com/something/remote"`},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.i.Name, func(t *testing.T) {
+			if tc.i.Code() != tc.expected {
+				t.Errorf("unexpected import code: %q != %q", tc.i.Code(), tc.expected)
+			}
+		})
+	}
+}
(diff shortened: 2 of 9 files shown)


MichaelUrman:
Confirmed that adding another directory layer provokes the behavior I originally saw. So move `...\go\src\calcsvc` to `...\go\src\svcs\calcsvc`; delete gen\\* and cmd\\* and calc.go if present. Upon running `goa gen svcs\calcsvc\design`, the output includes:
```
...\go\src\svcs\calcsvc> goa gen svcs\calcsvc\design
...\go\src\svcs\calcsvc\goa097345391\main.go:20:10: unknown escape sequence
...\go\src\svcs\calcsvc\goa097345391\main.go:20:18: unknown escape sequence

========
Content:
// Code generated by goa v2.0.0-wip, DO NOT EDIT.
//
// Code Generator
//
// Command:
// $ goa

package main

import (
        "flag"
        "fmt"
        "os"
        "path/filepath"
        "sort"
        "strings"
        "goa.design/goa/codegen/generator"
        "goa.design/goa/eval"
        "goa.design/goa/pkg"
        _ "svcs\calcsvc\design"
)
```
With 5ac020c6:
```
...\go\src\svcs\calcsvc> goa gen svcs\calcsvc\design
gen\calc\client.go
gen\calc\endpoints.go
gen\calc\service.go
gen\http\calc\client\cli.go
gen\http\calc\client\client.go
gen\http\calc\client\encode_decode.go
gen\http\calc\client\paths.go
gen\http\calc\client\types.go
gen\http\calc\server\encode_decode.go
gen\http\calc\server\paths.go
gen\http\calc\server\server.go
gen\http\calc\server\types.go
gen\http\cli\calc\cli.go
gen\http\openapi.json
gen\http\openapi.yaml

go\src\svcs\calcsvc> goa example svcs\calcsvc\design
calc.go
cmd\calc-cli\http.go
cmd\calc-cli\main.go
cmd\calc\http.go
cmd\calc\main.go
```
But cmd\calc\main.go includes an import of `calc "."`, resulting in a build error `local import "." in non-local package`. With 72940a0, deleting cmd\\*, and calling example again, cmd\calc\main.go includes an import of `calc "svcs/calcsvc"` instead and builds fine.

MichaelUrman:
Done; I created #1981 with more concise reproduction details, and less meandering text about how I tried to fix it.

PR #1987, opened 2019-02-03
Incorrect file to put import for implementing swagger

mnoster:
Modification to the instructions for implementing swagger.

The file to add the swagger import is in the `cmd/calcsvc/calcsvc.go` not `cmd/calcsvc/main.go`.


raphael:
Good catch, thank you!

mnoster:
yay!

PR #1990, opened 2019-02-04
Fix import generation in Windows

### LINKED ISSUE #1981: Import paths handled as file paths
Today goa (v2) processes import path strings with filepath, resulting in invalid code generation and/or processing on Windows. See PR #1976 for additional details, including code that fixes the symptoms, but not address the underlying problem. @raphael requested I open this as an issue instead, presumably to identify a better fix. Cheers!

## Expected Behavior
Running through the calcsvc tutorial on Windows should work and generate correct go code, both with calcsvc directly in `%GOPATH%\src\calcsvc` and in the nested `%GOPATH%\src\svcs\calcsvc`. Specifically, the commands `goa gen` and `goa example` (possibly others?) would handle, generate and parse import paths as import paths instead of file paths. Namely, in the nested case it should generate paths of the form `import _ "svcs/calcsvc/design"` and `import "svcs/calcsvc"` in calc.go and cmd\calc\main.go respectively.

## Actual Behavior
In both direct and nested paths the back slash is used instead of forward slash: `goa example` generates a calc.go with an invalid import path of the form `import _ "svcs\calcsvc\design"` and a cmd\calc\main.go with an invalid import of the form `import calcsvc "."`. In the direct case, `goa gen` runs to completion, and `goa example` generates files that cannot be built. In the nested case, `goa gen` fails, and addressing its proximate cause reveals the further problems in the code generated by `goa example`.

## Steps to Reproduce the Problem
Follow the [v2 branch readme](https://github.com/goadesign/goa/tree/233b20f76ec977d99910dd7a55cf45eacb2f979c) either as is, or varied to nest calcsvc under a path such as svcs\calcsvc, through the calls to `goa gen` and `goa example`, and when possible the call to `go build` in `cmd\calcsvc`. 

## Specifications

  - goa Version: v2.0.0-wip (tested at 233b20f7)
  - Platform: Windows

nitinmohan87:
Fixes #1981


raphael:
Looks great, thank you!

PR #2000, opened 2019-02-11
IND-1925 X-Ray tracing record in-progress requests

jim-slattery-rs:
Modernize the the goa X-Ray segments to support writing "In Progress" segments to the xray-daemon.

This way, in the X-Ray UI we can see a live view of what is happening in the system, with "Pending" segments and their hierarchy visible.

The current behavior is that X-Ray is only notified about a segment when it is fully completed, which could take several minutes. Meanwhile, the trace in the UI is confusing, missing the key top-level hierarchy while subsystems have already started reporting their completed segments.

This PR makes the following changes:
- Add the `SubmitInProgress` method to `xray.Segment`. Whoever creates the segment can call this method shortly afterwards to ensure the X-Ray UI will immediately show the segment, as "Pending".
- `xray` middleware, `xray.WrapDoer`, and `xray.WrapTransport` will automatically submit their newly created in-progress segments, with useful information populated regarding the request.
- Each (sub)segment now writes just itself to the xray-daemon, rather than having the parent submit a large document (potentially exceeding the 64KB limit) containing all of its subsegment descendants.
See: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-subsegments
- Save segment space by omitting empty fields: `in_progress`, `error`, `fault`, `throttle`.
This matches the AWS SDK: https://github.com/aws/aws-xray-sdk-go/blob/852dc097d2b83915ece2ab4953e4de711d953626/xray/segment_model.go#L40-L45

### DIFF
--- a/middleware/xray/middleware.go
+++ b/middleware/xray/middleware.go
@@ -46,6 +46,15 @@ const (
 //     }
 //     return
 //
+// An X-Ray trace is limited to 500 KB of segment data (JSON) being submitted
+// for it. See: https://aws.amazon.com/xray/pricing/
+//
+// Traces running for multiple minutes may encounter additional dynamic limits,
+// resulting in the trace being limited to less than 500 KB. The workaround is
+// to send less data -- fewer segments, subsegments, annotations, or metadata.
+// And perhaps split up a single large trace into several different traces.
+//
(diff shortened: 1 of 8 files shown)


raphael:
Great PR, thank you!

Issue #2002, opened 2019-02-13
V2: example CLIs produce no ouput in verbose/debug mode.


tim775:
## Expected Behavior
Running the example CLIs with the verbose flag should produce debug output.  E.g., the calc-cli should print something like:
```
$  go run examples/basic/cmd/calc-cli/* -v calc add -a 3 -b 213515
> GET http://localhost:8000/add/3/213515
< 200 OK
< Content-Length: 7
< Content-Type: application/json
< Date: Wed, 13 Feb 2019 18:37:50 GMT
213518

213518
$
```

## Actual Behavior
Running without the -v or --verbose flag produces the expected response, but with those flags nothing is output:
```
$  go run examples/basic/cmd/calc-cli/* --verbose calc add -a 3 -b 213515
$  go run examples/basic/cmd/calc-cli/* calc add -a 3 -b 213515
213518
$
```
## Steps to Reproduce the Problem (including complete and simple design if relevant)
Run the calc server: 
```go run examples/basic/cmd/calc/*```
Access the server with the calc-cli using the -verbose flag:
```go run examples/basic/cmd/calc-cli/* --verbose calc add -a 3 -b 2```

## Specifications

  - goa Version: 2
  - Platform (Linux/Windows etc.): All

**Comment by nitinmohan87** [comment]
This is because we don't call `Fprint` implemented by HTTP `debugDoer` after the endpoint function is invoked. Hence no actual HTTP request is made and debug doer does not print anything.

I propose a top-level `Debugger` interface
```
type Debugger interface {
  Fprint(io.Writer)
}
```
We can then remove the `DebugDoer` interface in HTTP and have `debugDoer` implement the `Debugger` interface. Similarly in gRPC client, we can have a `debugInvoker` which implements the `Invoker` and the `Debugger` interface. Now, in the example CLI code, we can return this `Debugger` interface from `doHTTP` and `doGRPC` only when `debug = true`. Then in CLI `main.go` we can do
```
var (
    endpoint goa.Endpoint
    payload  interface{}
    dbg        Debugger     // It is nil if debug == false
    err          error
  )
  {
    switch scheme {
    case "http", "https":
      endpoint, payload, dbg, err = doHTTP(scheme, host, timeout, debug)
    case "grpc", "grpcs":
      endpoint, payload, dbg, err = doGRPC(scheme, host, timeout, debug)
    default:
[...]

PR #2003, opened 2019-02-13
Fix the verbose option of the cli to output details and the response

tim775:


tim775:
This partially fixes #2002.  If we agree this is the right approach I'll update this PR to fix the other examples.

raphael:
LGTM

tim775:
@nitinmohan87 I've updated the rest of the examples and the example generator, please give it the once over then I'll merge it.

nitinmohan87:
Looks good

PR #2005, opened 2019-02-15
[BREAKING] Expose stream context to application code

### LINKED ISSUE #1972: v2: Websocket identifying stale clients
In v2, in server streaming requests using http/websocket transport, if the client closes the connection, the server currently can't identify this and the request handler only completes when it needs to send an event to the client. This may cause high number of gouroutines to be active if the streams are idle for longer periods of time. Goa provides a function to configure the connection.
```
ConnConfigureFunc func(*websocket.Conn) *websocket.Conn
```
I have custom pingers defined in this function which looks for stale client connections and closes them properly but the server request handler doesn't get notified of the connection closing. It would be nice if this function signature is changed to
```
ConnConfigureFunc func(context.Context, *websocket.Conn) (context.Context, *websocket.Conn)
```
which will let me cancel the context upon the client connection close and my request hander can look for context cancellation and complete the request properly.

nitinmohan87:
Fixes #1972 

## Breaking Changes

* Signature of goahttp.ConnConfigureFunc is changed to include context (#1972)
```
type ConnConfigureFunc func(context.Context, *websocket.Conn) (context.Context, *websocket.Conn)
```
* For streaming endpoints, the Service interface functions no longer takes `ctx` as an argument.
```
// The chatter service implements a simple client and server chat.
type Service interface {
  // Creates a valid JWT token for auth to chat.
  Login(context.Context, *LoginPayload) (res string, err error)
  // Echoes the message sent by the client.
  Echoer(*EchoerPayload, EchoerServerStream) (err error)
  // Listens to the messages sent by the client.
  Listener(*ListenerPayload, ListenerServerStream) (err error)
  // Summarizes the chat messages sent by the client.
  Summary(*SummaryPayload, SummaryServerStream) (err error)
  // Returns the chat messages sent to the server.
  // The "view" return value must have one of the following views
  //  - "tiny"
  //  - "default"
  History(*HistoryPayload, HistoryServerStream) (err error)
}
```
Context for streaming endpoint can be obtained from `Context()` method. For example,
```
func (s *chatterSvc) Echoer(p *chattersvc.EchoerPayload, stream chattersvc.EchoerServerStream) (err error) {
  stream.Context()  // gets the context for the streaming endpoint
```
NOTE: This above change is due to the new signature for the `ConnConfigureFunc` type to include context. In HTTP streaming, the configure function is executed only when `Send`/`Recv`/`Close` is called for the first time (to upgrade connection) in the application code which may change the context. Hence, the application code needs to call `stream.Context()` after calling `Send`/`Recv`/`Close` for the first time. Subsequent calls to `Send`/`Recv`/`Close` in the application code will not change the context.
* Dialer interface now has `DialContext` instead of `Dial`
* HTTP server constructor now takes a `Streamer` interface.
```
// New instantiates HTTP handlers for all the chatter service endpoints.
func New(
  e *chattersvc.Endpoints,
  mux goahttp.Muxer,
  dec func(*http.Request) goahttp.Decoder,
  enc func(context.Context, http.ResponseWriter) goahttp.Encoder,
  eh func(context.Context, http.ResponseWriter, error),
  up goahttp.Upgrader,
  connConfigFn goahttp.ConnConfigureFunc,
  stream goahttp.Streamer,  <---- New argument
) *Server {
```
This makes it possible for users to provide custom Streamer implementations. goa v2 provides a basic implementation which the example code generation initializes.

## Actions to resolve breaking changes

* If your app uses `ConnConfigureFunc` type update it to the new signature.
* If your app uses a custom `Dialer` interface, make sure it implements the `DialContext` method.
* Re-run `goa example` to generate the correct HTTP server initialization code (OR)
Update the HTTP server initialization code in `cmd` as follows
```
// Server-side
var (
    chatterServer *chattersvcsvr.Server
  )
  {
    eh := errorHandler(logger)
    upgrader := &websocket.Upgrader{}
    chatterServer = chattersvcsvr.New(
        chatterEndpoints, mux, dec, enc, eh, upgrader, nil,
       goahttp.NewServerStream(ctx),  <--- uses goa's HTTP streamer implementation
    )
  }

// Client-side
return cli.ParseEndpoint(
    scheme,
    host,
    doer,
    goahttp.RequestEncoder,
    goahttp.ResponseDecoder,
    debug,
    dialer,
    connConfigFn,
    goahttp.NewClientStream(ctx),  <--- uses goa's HTTP streamer implementation
  )
```


nitinmohan87:
Closing this in favor of #2011

PR #2006, opened 2019-02-15
Html encoding

tim775:


examples/basic/calc.go
```diff
@@ -2,6 +2,7 @@ package calc
```
raphael: Would be good to keep the calc example as is. The idea of this basic example is to have a minimal functional "first time" design.

examples/basic/cmd/calc-cli/http.go
```diff
@@ -10,12 +10,16 @@ import (
 	goahttp "goa.design/goa/http"
 )
 
-func doHTTP(scheme, host string, timeout int, debug bool) (goa.Endpoint, interface{}, error) {
+func doHTTP(scheme, host string, timeout int, acceptType string, debug bool) (goa.Endpoint, interface{}, error) {
```
raphael: Not sure I understand why this needs to be specified as a parameter, this should all be driven by the design?

expr/http_response.go
```diff
@@ -139,6 +139,17 @@ func (r *HTTPResponseExpr) Validate(e *HTTPEndpointExpr) *eval.ValidationErrors
 	if isrt {
 		inview = " all views in"
 	}
+
+	// text/html can only encode strings so make sure that
```
raphael: Please finish comment :) We should also allow for `Bytes` because that's what the Go templates produce.

http/encoding.go
```diff
@@ -236,3 +245,45 @@ func SetContentType(w http.ResponseWriter, ct string) {
 	}
 	w.Header().Set("Content-Type", h+suffix)
 }
+
+func newTextHtmlEncoder(w io.Writer) Encoder {
```
raphael: Please use `HTML` instead of `Html` in Go identifiers.

http/encoding.go
```diff
@@ -236,3 +245,45 @@ func SetContentType(w http.ResponseWriter, ct string) {
...
+}
+
+type textHtmlEncoder struct {
+	w io.Writer
+}
+
+func (e *textHtmlEncoder) Encode(v interface{}) error {
+	var b []byte
+
+	strptr, isstrptr := v.(*string)
```
raphael: We should protect against panics here and return an error if v is not a string. Do we really need to support pointers? Also we should support slices of bytes.

http/encoding.go
```diff
@@ -236,3 +245,45 @@ func SetContentType(w http.ResponseWriter, ct string) {
...
+func newTextHtmlDecoder(r io.Reader) Decoder {
+	return &textHtmlDecoder{r}
+}
+
+type textHtmlDecoder struct {
+	r io.Reader
+}
+
+func (e *textHtmlDecoder) Decode(v interface{}) error {
+	strptr := v.(*string)
```
raphael: We should also support slices of bytes. The code shouldn't panic if `v` is not a string nor a slice of bytes.

http/encoding.go
arangamani: If you have a design that produces a result like
```go
Result(func() {
  Attribute("content", String, "The html content", func() {
    Example("<html>...</html>")
  })
  // .
  // .
})
```
where the "content" field is optional and you specify this field for the response body like
```go
HTTP(func() {
  GET("")
  Response(StatusOK, func() {
    Body("content")
    ContentType("text/html")
  })
})
```
I've seen the encoder called with a string pointer.

raphael: OK makes sense, thanks.

examples/basic/calc.go
tim775: I reverted the changes to calc and added a new "encodings" example.  I'm not sure if these changes warrant a new example so I will get rid of it entirely if you think we don't need it.

tim775:
@raphael I updated this to address your comments.

examples/encodings/README.md
```diff
@@ -0,0 +1,23 @@
+# goa v2 Encodings Example
+
+This simple example demonstrates HTTP responses with the 'text/html' content type encoding.  It consists of a simple
+endpoint concatenates strings.
```
raphael: "that"

raphael:
LGTM

PR #2010, opened 2019-02-21
Fix header attribute validated in the body

### LINKED ISSUE #2009: Payload combining required header and body field validates incorrectly
Using the following design: "$GOPATH/src/foo/design/api.go"
```
package design

import (
	. "goa.design/goa/dsl"
)

var _ = Service("Something", func() {
	Error("bad_request")

	HTTP(func() {
		Path("/foo")
	})

	Method("create", func() {
		Payload(CreatePayload)
		Result(Foo)

		HTTP(func() {
			POST("/")
			Header("api_version:Api-Version")
			Response(StatusOK)
		})
	})

	Method("hide", func() {
		Payload(func() {
			Extend(CommonAttributes)
			Attribute("name", String)
			Attribute("user_id", Int)
			Required("api_version", "name", "user_id")
		})
		Result(Empty)
		Error("not_found")

		HTTP(func() {
			POST("/{name}/hide")
			Header("api_version:Api-Version")
			Response(StatusOK)
			Response("not_found", StatusNotFound)
		})
	})
})

	Attribute("api_version", String, func() {
		Enum("1.0")
	})
})

var CreatePayload = Type("CreatePayload", func() {
	Extend(CommonAttributes)
	Attribute("name", String)
	Required("api_version", "name")
})

var Foo = ResultType("application/vnd.something.foo", func() {
	TypeName("Foo")

	Attributes(func() {
		Attribute("name", String)
		Required("name")
	})

	View("default", func() {
		Attribute("name")
	})
})
```
The validation for the "hide" action code-generates such that it always fails validation:
```
// ValidateHideRequestBody runs the validations defined on HideRequestBody
func ValidateHideRequestBody(body *HideRequestBody) (err error) {
	if body.UserID == nil {
		err = goa.MergeErrors(err, goa.MissingFieldError("user_id", "body"))
	}
	if body.APIVersion != nil {
		if !(*body.APIVersion == "1.0") {
			err = goa.MergeErrors(err, goa.InvalidEnumValueError("body.api_version", *body.APIVersion, []interface{}{"1.0"}))
		}
	}
	return
}
```
This is because the "API-Version" header is supposed to be checked after body validation, which it is in the generated DecodeHideRequest() method. It is incorrect to check it in body validation since it is a header. Note that ValidateCreateRequestBody() is correct, apparently because the CreatePayload is formally declared. If I change to use a formally declared HidePayload then it also code generates correctly:
```
		Payload(HidePayload)
...
var HidePayload = Type("HidePayload", func() {
	Extend(CommonAttributes)
	Attribute("name", String)
	Attribute("user_id", Int)
	Required("api_version", "name", "user_id")
})
...

// ValidateHideRequestBody runs the validations defined on HideRequestBody
func ValidateHideRequestBody(body *HideRequestBody) (err error) {
	if body.UserID == nil {
		err = goa.MergeErrors(err, goa.MissingFieldError("user_id", "body"))
	}
	return
}
```

nitinmohan87:
Fixes #2009

### DIFF
--- a/expr/http_endpoint.go
+++ b/expr/http_endpoint.go
@@ -525,9 +525,18 @@ func (e *HTTPEndpointExpr) Finalize() {
 	init(e.Params)
 	init(e.Headers)
 
-	e.Body = httpRequestBody(e)
 	if e.Body != nil {
+		// rename type to add RequestBody suffix so that we don't end with
+		// duplicate type definitions - https://github.com/goadesign/goa/issues/1969
+		e.Body = httpRequestBody(e)
 		e.Body.Finalize()
+	} else {
+		// Compute body from the payload expression
+		e.Body = httpRequestBody(e)
+		// Don't call e.Body.Finalize() after computing the body because the
+		// payload expression might define bases and references which will be
+		// added to the body even when design explicitly maps them to headers or
+		// params.
 	}
 
 	e.StreamingBody = httpStreamingBody(e)
--- a/http/codegen/client_body_types_test.go
+++ b/http/codegen/client_body_types_test.go
@@ -69,6 +69,7 @@ func TestClientTypes(t *testing.T) {
 		Code string
 	}{
 		{"multiple-methods", testdata.MultipleMethodsDSL, MultipleMethodsClientTypesFile},
+		{"payload-extend-validate", testdata.PayloadExtendedValidateDSL, PayloadExtendedValidateClientTypesFile},
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


expr/http_endpoint.go
```diff
@@ -525,9 +525,18 @@ func (e *HTTPEndpointExpr) Finalize() {
 	init(e.Params)
 	init(e.Headers)
 
-	e.Body = httpRequestBody(e)
```
raphael: Couldn't the below just be:
```go
e.Body = httpRequestBody(e)
if e.Body != nil {
   e.Body.Finalize()
}
```
(+comments)

PR #2011, opened 2019-02-22
[BREAKING] Pass context cancellation to ConnConfigureFunc

### LINKED ISSUE #1972: v2: Websocket identifying stale clients

nitinmohan87:
Fixes #1972 

## Breaking Changes

* Signature of goahttp.ConnConfigureFunc is changed to include context cancel function
```
type ConnConfigureFunc func(*websocket.Conn, context.CancelFunc) *websocket.Conn
```
* HTTP server and client constructor now takes a connection configure struct.
* Dialer interface now has `DialContext` instead of `Dial`

## Actions to resolve breaking changes

* If your app uses `ConnConfigureFunc` type, update it to the new signature.
* If your app uses a custom `Dialer` interface, make sure it implements the `DialContext` method.
* Pass a reference to `ConnConfigurer` struct when initializing [HTTP server](https://github.com/goadesign/goa/blob/eed4a1149392746cd82aac210c9667aa2f79eccc/examples/streaming/cmd/chatter/http.go#L58-L60) and HTTP client CLI.

See [streaming example](https://github.com/goadesign/goa/tree/30a29fce665742c1b0761c4ad9f54a092201dbdc/examples/streaming) for reference.


examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
 ]
 ```
 
+### `subscribe` Endpoint
+
+This endpoint illustrates **streaming result**. The client sends a
```
raphael: illustrates a ...

examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
...
+payload defined in `Payload` DSL containing the JWT token. Once auth succeeds,
```
raphael: defined in the `Payload` DSL and that contains a JWT token.

examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
...
+the server establishes a stream through which it sends events to the client
```
raphael: a stream. The server sends messages to the client using the stream.

examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
...
+{
+    "Message": "hello",
+    "Action": "added",
+    "AddedAt": "2018-08-14T12:32:30-07:00"
+}
+```
+
 ### `history` Endpoint
 
-This endpoint illustrates **streaming result**. The client sends a payload
```
raphael: illustrates a

examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
...
+This endpoint illustrates **streaming result** with views. The client sends a
 payload defined in `Payload` DSL containing the JWT token and an optional
```
raphael: defined in the `Payload` DS and that contains a JWT token and

examples/streaming/README.md
```diff
@@ -65,9 +65,28 @@ check check check
...
 "view" parameter. Once auth succeeds, the server streams all the
 messages sent by the client rendered using the optional "view" parameter.
```
raphael: Is it sent by the client or to the client? Should that be:
The server sends results to the client using the stream. Each result struct is first projected to the view matching the the "view" parameter.

examples/streaming/README.md
```diff
@@ -134,28 +153,19 @@ var (
     ReadBufferSize: 512,
     WriteBufferSize: 512,
   }
-  myConnConfigurer := func(conn *websocket.Conn) *websocket.Conn {
-    conn.SetReadDeadline(time.Now()+time.Minute*2)
-    return conn
-  }
-  chatterServer = chattersvcsvr.New(chatterEndpoints, mux, dec, enc, eh, myUpgrader, myConnConfigurer)
+  chatterServer = chattersvcsvr.New(chatterEndpoints, mux, dec, enc, eh, myUpgrader, nil, nil, nil, nil)
```
raphael: I think we should group all the configurers into one struct like we do with endpoints.

examples/streaming/README.md
```diff
@@ -166,10 +176,18 @@ endpoint, payload, err := cli.ParseEndpoint(
   goahttp.ResponseDecoder,
   debug,
   myDialer,
-  myConnConfigurer,
+  nil,
```
raphael: We should group all these into one struct containing one field per configurer similar to what we do with endpoints. The generated code should also product a `NewConfigurers` function similar to the `NewEndpoints` function that accepts a single configurer and produces the struct with each field initialzed with that configurer. This way the user can simply create the server struct with something like `New(chatterEndpoints, mux, dec, enc, eh, myUpgrader, NewConfigurers(myconfig))`. The user can also create the configurers struct and override each field prior to creating the server if need be.

examples/streaming/README.md
```diff
@@ -166,10 +176,18 @@ endpoint, payload, err := cli.ParseEndpoint(
...
+  nil,
+  nil,
+  nil,
 )
 ```
 
+Check out the `pingPonger` defined in `cmd/chatter/http.go` on how to use
```
raphael: for how to use

examples/streaming/cmd/chatter-cli/http.go
```diff
@@ -38,9 +38,14 @@ func doHTTP(scheme, host string, timeout int, debug bool) (goa.Endpoint, interfa
 		goahttp.ResponseDecoder,
 		debug,
 		dialer,
-		connConfigFn,
+		nil,
```
raphael: Change this to accept the configurers struct instead.

examples/streaming/cmd/chatter/http.go
```diff
@@ -55,7 +55,7 @@ func handleHTTPServer(ctx context.Context, u *url.URL, chatterEndpoints *chatter
 	{
 		eh := errorHandler(logger)
 		upgrader := &websocket.Upgrader{}
-		chatterServer = chattersvcsvr.New(chatterEndpoints, mux, dec, enc, eh, upgrader, nil)
+		chatterServer = chattersvcsvr.New(chatterEndpoints, mux, dec, enc, eh, upgrader, nil, nil, nil, pingPonger(logger), pingPonger(logger))
```
raphael: Change this to use the configurers struct.

http/codegen/client.go
```diff
@@ -256,7 +256,11 @@ type {{ .ClientStruct }} struct {
 	decoder    func(*http.Response) goahttp.Decoder
 	{{- if streamingEndpointExists . }}
 	dialer goahttp.Dialer
-	connConfigFn goahttp.ConnConfigureFunc
+		{{- range .Endpoints }}
```
raphael: Replace that with a generated struct (see above).

http/codegen/client.go
```diff
@@ -272,7 +276,11 @@ func New{{ .ClientStruct }}(
 	restoreBody bool,
 	{{- if streamingEndpointExists . }}
 	dialer goahttp.Dialer,
-	connConfigFn goahttp.ConnConfigureFunc,
+		{{- range .Endpoints }}
```
raphael: same as above

http/codegen/client.go
```diff
@@ -286,7 +294,11 @@ func New{{ .ClientStruct }}(
 		encoder:           enc,
 		{{- if streamingEndpointExists . }}
 		dialer: dialer,
-		connConfigFn: connConfigFn,
+			{{- range .Endpoints }}
```
raphael: ditto

http/codegen/client_cli.go
```diff
@@ -678,7 +681,13 @@ func ParseEndpoint(
 	restore bool,
 	{{- if streamingCmdExists . }}
 	dialer goahttp.Dialer,
-	connConfigFn goahttp.ConnConfigureFunc,
+	{{- range . }}
```
raphael: replace with struct

http/codegen/client_cli.go
```diff
@@ -770,9 +779,9 @@ func ParseEndpoint(
 	)
 	{
 		switch svcn {
-	{{- range . }}
+			{{- range . }}
 		case "{{ .Name }}":
-			c := {{ .PkgName }}.NewClient(scheme, host, doer, enc, dec, restore{{ if .NeedStream }}, dialer, connConfigFn{{- end }})
+			c := {{ .PkgName }}.NewClient(scheme, host, doer, enc, dec, restore{{ if .NeedStream }}, dialer{{ range .Subcommands }}{{ if .NeedStream }}, {{ .FullName }}ConfigFn{{ end }}{{ end }}{{- end }})
```
raphael: replace with struct

http/codegen/example_cli.go
```diff
@@ -137,7 +136,13 @@ const (
 		debug,
 		{{- if needStreaming .Services }}
 		dialer,
-		connConfigFn,
+			{{- range .Services }}
```
raphael: Just one `nil`

http/codegen/example_server.go
```diff
@@ -238,7 +238,7 @@ func handleHTTPServer(ctx context.Context, u *url.URL{{ range $.Services }}{{ if
 	{{- end }}
 	{{- range .Services }}
 		{{-  if .Endpoints }}
-		{{ .Service.VarName }}Server = {{ .Service.PkgName }}svr.New({{ .Service.VarName }}Endpoints, mux, dec, enc, eh{{ if needStream $.Services }}, upgrader, nil{{ end }}{{ range .Endpoints }}{{ if .MultipartRequestDecoder }}, {{ $.APIPkg }}.{{ .MultipartRequestDecoder.FuncName }}{{ end }}{{ end }})
+		{{ .Service.VarName }}Server = {{ .Service.PkgName }}svr.New({{ .Service.VarName }}Endpoints, mux, dec, enc, eh{{ if needStream $.Services }}, upgrader{{ range .Endpoints }}{{ if .ServerStream }}, nil{{ end }}{{ end }}{{ end }}{{ range .Endpoints }}{{ if .MultipartRequestDecoder }}, {{ $.APIPkg }}.{{ .MultipartRequestDecoder.FuncName }}{{ end }}{{ end }})
```
raphael: replace with struct

http/codegen/server.go
```diff
@@ -300,7 +306,11 @@ func {{ .ServerInit }}(
 	eh func(context.Context, http.ResponseWriter, error),
 	{{- if streamingEndpointExists . }}
 	up goahttp.Upgrader,
-	connConfigFn goahttp.ConnConfigureFunc,
+		{{- range .Endpoints }}
```
raphael: use struct

Issue #1705, opened 2018-05-03
Add tests for expr.DocsExpr.EvalName()


tchssk:
Reference implementation: #1702

Issue #1704, opened 2018-05-03
Add tests for expr.ContactExpr.EvalName()


tchssk:
Reference implementation: #1702

PR #2012, opened 2019-02-23
Add tests for expr.DocsExpr.EvalName() and expr.ContactExpr.EvalName()

laches1sm:
This solves issues #1705 and #1704 by adding tests for these two functions.

If there is any issues, feel free to give me a shout :)

### DIFF
--- a/expr/api_test.go
+++ b/expr/api_test.go
@@ -130,6 +130,34 @@ func TestApiExpr_Hash(t *testing.T) {
 	}
 }
 
+func TestDocsExprEvalName(t *testing.T) {
+	cases := map[string]struct {
+		url      string
+		expected string
+	}{
+		"parrot": {url: "http://parrot.com", expected: "Documentation http://parrot.com"},
+	}
+	for key, testcase := range cases {
+		docExpr := DocExpr{URL: testcase.url}
+		if actual := docExpr.EvalName(); actual != tc.expected {
+			t.Errorf("%s: got %#v, expected %#v", k, actual, tc.expected)
+		}
+	}
+}
+func TestContactExprEvalName(t *testing.T) {
+	cases := map[string]struct {
+		url      string
+		expected string
+	}{
+		"parrot": {name: "parrot@email.com", expected: "Contact parrot@email.com"},
+	}
+	for key, testcase := range cases {
+		contactExpr := ContactExpr{Name: testcase.name}
+		if actual := contactExpr.EvalName(); actual != tc.expected {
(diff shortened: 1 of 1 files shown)


raphael:
Looks great and I would merge as is but just to provide constructive feedback:
* The idiomatic way of naming local variables in Go favors short names, so for example `tc` instead of `testcase` etc. (see http://wordaligned.org/articles/go-for-short-variable-names)
* There could be additional tests that initialize other fields of the expression structs to test that these do NOT affect the value of `EvalName()`.

Again these are nitpicks!

laches1sm:
Awesome,thank you @raphael ! :D I'll add those in ASAP.

expr/api_test.go
```diff
@@ -132,29 +132,32 @@ func TestApiExpr_Hash(t *testing.T) {
...
 func TestDocsExprEvalName(t *testing.T) {
 	cases := map[string]struct {
-		url      string
-		expected string
+		url         string
+		expected    string
+		description string
 	}{
-		"test": {url: "http://parrot.com", expected: "Documentation http://parrot.com"},
+		"test": {url: "http://parrot.com", description: "A website for a parrot API", expected: "Documentation http://parrot.com"},
```
raphael: Instead of replacing the existing test you could leave it and add new test cases:
```go
"uses name": {url: "http://parrot.com", expected: "Documentation http://parrot.com"},
"does not use description": {url: "http://parrot.com", description: "A website for a parrot API", expected: "Documentation http://parrot.com"},
"does not use documentation" ....
```

expr/api_test.go
```diff
@@ -132,29 +132,32 @@ func TestApiExpr_Hash(t *testing.T) {
...
 		expected string
 	}{
-		"test": {name: "parrot@email.com", expected: "Contact parrot@email.com"},
+		"test": {name: "parrot", email: "parrot@gopher.com", url: "https://parrot.com", expected: "Contact parrot"},
 	}
-	for key, testcase := range cases {
-		contactExpr := ContactExpr{Name: testcase.name}
-		if actual := contactExpr.EvalName(); actual != testcase.expected {
-			t.Errorf("%s: got %#v, expected %#v", key, actual, testcase.expected)
+	for k, tc := range cases {
```
raphael: Same as above

expr/api_test.go
laches1sm: Ah, I'm silly, sorry about that! That's been addressed in the latest commit. Thank you so much! :D

raphael:
Perfect, thank you!

Issue #1996, opened 2019-02-08
v2: Graceful shutdown of streaming requests in gRPC


arangamani:
The `GracefulStop` function in the gRPC server works great but it waits for all requests to finish including streaming requests. If the requests are server side streaming requests that keep the connection open and send updated events to clients, they need to get notified that the server is stopping. The way to achieve this seems to be using a custom logic that cancels the context upon server shutdown. I have written the following middleware and tested it in my application. 

It would be nice if it can be a built-in gRPC interceptor and the examples include this middleware and show how to stop a gRPC server gracefully.

Here is what I have started with:
```go
// StreamCanceler provides a middleware that can be used to gracefully stop streaming requests.
// To stop streaming requests, simply pass in a context with cancellation and cancel the context.
// When the context given to the StreamCanceler is canceled, it does the following:
//   1. Stops accepting further streaming requests and returns the code Unavailable with message
//      "server is stopping".
//   2. Cancels the context of all streaming requests. Your request handler should obey to the
//      cancelation of request context.
// Example:
//   var (
//     ctxCancel  context.Context
//     cancelFunc context.CancelFunc
//   )
//   ctxCancel, cancelFunc = context.WithCancel(parentCtx)
//   streamInterceptor := StreamCanceler(ctxCancel)
//   // Use the interceptor in your server and when you need to shutdown
//   // your server, simply cancel the context given to the StreamCanceler interceptor.
//   cancelFunc()
//
//   // In your application code, look for context cancellation and respond with proper code.
//     for {
//       select {
//       case <-ctx.Done():
//         return status.Error(codes.Canceled, "canceled")
//       ...
//
func StreamCanceler(ctx context.Context) grpc.StreamServerInterceptor {
	var (
		cancels   []context.CancelFunc
		cancelMu  = new(sync.Mutex)
		canceling uint32
	)
	go func() {
		<-ctx.Done()
		atomic.StoreUint32(&canceling, 1)
		cancelMu.Lock()
		defer cancelMu.Unlock()
		for _, cancel := range cancels {
			cancel()
		}
	}()
	return grpc.StreamServerInterceptor(func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
		if atomic.LoadUint32(&canceling) == 1 {
			return status.Error(codes.Unavailable, "server is stopping")
		}
		var (
			cctx   = ss.Context()
			cancel context.CancelFunc
		)
[...]

PR #2014, opened 2019-02-25
Stream canceler middleware

tim775:


tim775:
Fixes #1996

grpc/middleware/canceler.go
```diff
@@ -0,0 +1,88 @@
...
+		)
+		cctx, cancel = context.WithCancel(cctx)
+
+		// add the cancel function
+		cancelMu.Lock()
+		cancels[&cancel] = struct{}{}
+		cancelMu.Unlock()
+
+		// invoke rpc
+		err := handler(srv, &ctxServerStream{cctx, ss})
```
nitinmohan87: We already have a [WrappedServerStream](https://github.com/goadesign/goa/blob/v2/grpc/middleware/helpers.go#L14) which we can use here instead of defining a new type.

grpc/middleware/canceler.go
```diff
@@ -0,0 +1,88 @@
...
+
+		// remove the cancel function
+		cancelMu.Lock()
+		delete(cancels, &cancel)
```
arangamani: So the `cancel` function never gets called if the request completes?

tim775: Yes that was my intention.  Maybe I'm not understanding, but why would you want to cancel a completed request?

PR #2015, opened 2019-02-26
Port in progress xray segment

### LINKED ISSUE #1998: v2: Tracing streaming requests
Tracing (with X-Ray) for streaming requests today doesn't seem to work very well. I'm proposing some changes here that may work for both gRPC streaming as well as websocket requests in the HTTP transport.

## Stream Server Interceptor
The stream server interceptor doesn't finish until the streaming request finishes. Segments are not sent to X-Ray daemon until the request finishes (including any subsegments created by the service). Also error information is not recorded for streaming requests. Streaming requests can run for an indefinite amount time. There may be other downstream services reporting segments and without the segment from the originating service we may not have the proper context when looking at the traces.

### Proposal
The proposal here is to send an in-progress segment to X-Ray daemon as we receive a streaming request. We do the same for any stream subsegments that get created. When the streaming request finished, we can complete the in-progress segments. We can include a boolean in the Segment struct that helps us identify whether a (sub)segment corresponds to a streaming request. The X-Ray daemon handles batching segments sent to X-Ray so it should be okay to send in-progress segments as soon as we identify them.

## Stream Client Interceptor
The stream client interceptor creates a client stream and finishes immediately so currently the subsegments for streaming request is very short and only corresponds to initiating the streaming request. 

### Proposal
The example here wraps the client stream and intercepts the messages sent by the server stream and finishes the client span if any error is returned: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/client_interceptors.go It seems very simple to do and should work for us.

I think the proposal here can apply to websocket requests as well. If it makes things simpler, we can actually send in-progress segments for regular requests as well. It might just be a lot of messages sent to the X-Ray daemon if the request finishes quickly.

nitinmohan87:
Fixes #1998


grpc/middleware/xray/middleware_test.go
```diff
@@ -52,3 +98,472 @@ func TestNewStreamServer(t *testing.T) {
...
+			Segment:  Seg{status.Error(codes.Unknown, "error"), true},
+		},
+		"fault": {
+			Trace:    Tra{traceID, spanID, ""},
+			Request:  Req{remoteAddr, clientIP, agent},
+			Response: Res{codes.InvalidArgument},
+			Segment:  Seg{status.Error(codes.InvalidArgument, "error"), true},
+		},
+	}
+	for k, c := range cases {
```
jim-slattery-rs: for new tests like this, I'd recommend using `t.Run` within the for loop:
```go
t.Run(k, func(t *testing.T) {
```
That way you don't need to worry about `k` in every error message.

nitinmohan87: Agree! I also follow the above pattern usually. But got a bit lazy when copy-pasting stuff over.

nitinmohan87: Changed the tests. Thanks!

grpc/middleware/xray/middleware.go
```diff
@@ -143,18 +173,51 @@ func StreamClient(host string) grpc.StreamClientInterceptor {
...
 		} else {
 			sub.RecordResponse(nil)
 		}
-		return cs, err
+		return &xrayStreamClientWrapper{cs, sub}, err
 	})
 }
+
+func (c *xrayStreamClientWrapper) SendMsg(m interface{}) error {
+	err := c.ClientStream.SendMsg(m)
```
raphael: You could write all these as:
```go
if err := c.ClientStream.SendMsg(m); err != nil {
...
   return err
}
return nil
```

PR #2017, opened 2019-02-28
Improve package documentation for goa and codegen packages

nitinmohan87:
Improves godoc documentation for the top-level `goa`, `eval`, and `codegen` packages.


codegen/funcs.go
```diff
@@ -80,7 +63,7 @@ func Indent(s, prefix string) string {
 }
 
 // Add adds two integers and returns the sum of the two.
-func Add(a, b int) int { return a + b }
+//func Add(a, b int) int { return a + b }
```
raphael: Should that be removed entirely?

codegen/server/docs.go
```diff
@@ -0,0 +1,3 @@
+// Package server contains code generation algorithms to produce an example
+// server and client implementation for the transports defined in the design.
```
raphael: does package server really produce an example client implementation?

nitinmohan87: I named the package after the `Server` expression. It produces both the example server and the client. Do you want to change the package name to something else?

raphael-rs: We probably should yes. If the package generates the example code we could call it `example`?

PR #2018, opened 2019-03-01
Add tests for expr.Object.Rename()

tchssk:
#1423

### DIFF
--- a/expr/types_test.go
+++ b/expr/types_test.go
@@ -706,17 +706,17 @@ func TestArrayIsCompatible(t *testing.T) {
 
 func TestObjectRename(t *testing.T) {
 	cases := map[string]struct {
-		n, m     string
+		old, new string
 		expected []string
 	}{
 		"renamed": {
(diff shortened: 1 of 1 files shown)


expr/types_test.go
```diff
@@ -704,6 +704,47 @@ func TestArrayIsCompatible(t *testing.T) {
 	}
 }
 
+func TestObjectRename(t *testing.T) {
+	cases := map[string]struct {
+		n, m     string
```
nitinmohan87: It is a little difficult to understand what `n` and `m` variables are. Adding some comments on what `n`, `m`, and `expected` or renaming the variables would be helpful.

nitinmohan87:
Thanks for the PR! I made a minor comment to make it easier to understand the tests.

expr/types_test.go
tchssk: OK. 88edc4b renamed the variables.

PR #2019, opened 2019-03-01
Fix a bug where array of bytes is generated as pointer

nitinmohan87:


http/codegen/typedef.go
```diff
@@ -59,7 +59,7 @@ func goTypeDef(scope *codegen.NameScope, att *expr.AttributeExpr, ptr, useDefaul
 				fn = codegen.GoifyAtt(at, name, true)
 				tdef = goTypeDef(scope, at, ptr, useDefault)
 				if expr.IsPrimitive(at.Type) {
-					if ptr || mat.IsPrimitivePointer(name, useDefault) {
+					if (ptr || mat.IsPrimitivePointer(name, useDefault)) && at.Type != expr.Bytes && at.Type != expr.Any {
```
raphael: I'm wondering whether `IsPrimitivePointer` shouldn't include these checks? maybe it can't because we need to differentiate in other places.

nitinmohan87: The problem here is we force the generation of pointers using `ptr` param. Even if `IsPrimitivePointer` is `false` we generate the type as pointer if `ptr` is true. We need to create exceptions for the `Byte` and `Any` type like above.

raphael: Makes sense, thanks for the explanation.

PR #2020, opened 2019-03-01
Fix conflicting API package name when API not defined in design

nitinmohan87:


codegen/funcs.go
```diff
@@ -237,6 +238,22 @@ func WrapText(text string, maxChars int) string {
...
+// APIPkg returns the name of the API package for the given root.
+func APIPkg(root *expr.RootExpr) string {
+	pkg := strings.ToLower(Goify(root.API.Name, false))
+	for _, s := range root.Services {
+		// When API DSL is not specified in the design, goa generates a default API
+		// expression with same name as the first service in the design. We need to
+		// make sure we the API package name does not conflict with the service
+		// package name in this case.
+		if strings.ToLower(Goify(s.Name, false)) == pkg {
+			pkg += "api"
```
raphael: Not very likely but there could be a service called `svc` and one called `svcapi` and if this iterates over `svcapi` first then we could end up with a conflict in the end. I am wondering if we couldn't add a method to the `NameScope` struct to make it possible to add names to it without needing to use the `HashedUnique` function. Then we could use `Unique` to produce the unique name for the package.

nitinmohan87: I thought about the same about having a service name with `api` prefix but felt it is very unlikely. Regarding `NameScope` there is already a `Unique` function that returns a unique name without adding it to the scope. We could use that.

raphael: Yes but to use it we need to be able to add names to the scope without calling `HashedUnique` hence my suggestion to add a method that allows doing that. Agree it's a small chance but I believe this pattern could be reused in other places (I know we have other code that renames by "taking chances").

nitinmohan87: The problem with using scope here is we [initialize the scope with the API name](https://github.com/goadesign/goa/blob/v2/codegen/scope.go#L38-L40). We could change the service package name to be suffixed with `svc` instead of coming up with a unique name for the API package name. How does that sound?

raphael-rs: Wouldn't that have the same issue? If the API name finishes with `svc` then it could clash. I was suggesting to create a new scope just for this - not reusing the existing scope. The idea is to reuse the existing algorithm implemented in the scope method to create a unique name. We'd create a new scope struct, initialize it with the existing service names and then ask it to create a new unique name based on a given prefix/suffix.

raphael:
LGTM, thank you!

PR #2024, opened 2019-03-05
Add tests for expr.ServiceExpr.Error()

tchssk:
#1423

### DIFF
--- a/expr/service_test.go
+++ b/expr/service_test.go
@@ -79,9 +79,11 @@ func TestServiceExprError(t *testing.T) {
 		},
 	}
 	for k, tc := range cases {
-		if actual := s.Error(tc.name); actual != tc.expected {
-			t.Errorf("%s: got %#v, expected %#v", k, actual, tc.expected)
-		}
+		t.Run(k, func(t *testing.T) {
+			if actual := s.Error(tc.name); actual != tc.expected {
+				t.Errorf("got %#v, expected %#v", actual, tc.expected)
+			}
+		})
 	}
 }
 
--- a/expr/service_test.go
+++ b/expr/service_test.go
@@ -43,6 +43,42 @@ func TestServiceExprMethod(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


expr/service_test.go
```diff
@@ -43,6 +43,42 @@ func TestServiceExprMethod(t *testing.T) {
...
+		expected *ErrorExpr
+	}{
+		"exist": {
+			name:     "foo",
+			expected: errorFoo,
+		},
+		"not exist": {
+			name:     "baz",
+			expected: nil,
+		},
```
nitinmohan87: Would it be possible to add a test case where the `Root` expression has an error with the given `name` and `service.Error()` returns the correct error expression? Right now, your test cases does not cover this condition - https://github.com/goadesign/goa/blob/v2/expr/service.go#L70

expr/service_test.go
```diff
@@ -43,6 +43,42 @@ func TestServiceExprMethod(t *testing.T) {
...
+	}
+
+	for k, tc := range cases {
```
nitinmohan87: Let's start using `t.Run` for new tests. This we don't need to print `k` in `t.Errorf`.
```go
for k, tc := range cases {
  t.Run(k, func(t *testing.T) {
    ...
    if actual := s.Error(tc.name); actual != tc.expected {
      t.Errorf("invalid error expression: got %#v, expected %#v", actual, tc.expected)
    }
  })
}
```

expr/service_test.go
tchssk: Woops, https://github.com/goadesign/goa/pull/2024/commits/7d3a3363dda4cd999c1c795bc9e2d556e4d7368c adds the test.

expr/service_test.go
tchssk: https://github.com/goadesign/goa/pull/2024/commits/e21a2308251d4044c2c6bd6106ed7a61992958ae

nitinmohan87:
Looks great.. Thank you so much!

PR #2032, opened 2019-03-11
Messages called from rpc match message names when reserved name

allofthesepeople:
`protoBufify` is applied inconsistently when the method is a reserved name. Resulting in different signatures to the protobuf message types and method signature.

## Example design

```go
package design

import . "goa.design/goa/dsl"

var _ = Service("My Great Cat Service", func() {
	Method("rpc", func() {
		Result(Empty)
		GRPC(func() {
			Response(CodeOK)
		})
	})
})
```

```shell
➜ goa gen github.com/allofthesepeople/goatest/design
exit status 1
failed to run protoc: exit status 1: my_great_cat_service.proto:17:18: "Rpc_Request" is not defined.
my_great_cat_service.proto:17:40: "Rpc_Response" is not defined.
```

resulting in a protofile where the service definition signatures do not match the Message types:

```proto
…
service MyGreatCatService {
	// RPC implements rpc.
	rpc RPC (Rpc_Request) returns (Rpc_Response);
}

message RPCRequest {
}

message RPCResponse {
}
```

Change applies `protoBufify` to the whole message name rather than just the method part.

### DIFF
--- a/grpc/codegen/proto_test.go
+++ b/grpc/codegen/proto_test.go
@@ -3,9 +3,9 @@ package codegen
 import (
 	"testing"
 
+	"github.com/allofthesepeople/goa/grpc/codegen/testdata"
 	"goa.design/goa/codegen"
 	"goa.design/goa/expr"
-	"goa.design/goa/grpc/codegen/testdata"
 )
 
 func TestProtoFiles(t *testing.T) {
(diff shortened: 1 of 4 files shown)


raphael:
LGTM, @nitinmohan87 would you care to give it a review as well? Are there other places that might be affected? Thanks!

PR #2068, opened 2019-04-02
Update Upgrading.md

jim-slattery-rs:
Tiny typos.

### DIFF
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -71,9 +71,8 @@ var Person = MediaType("application/vnd.goa.person", func() {
 		Required("name")
 	})
 	View("default", func() {  // View defines a rendering of the media type.
-		Attribute("id")   // Media types may have multiple views and must
-		Attribute("href") // have a "default" view.
-		Attribute("name")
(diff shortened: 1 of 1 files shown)


docs/Upgrading.md
```diff
@@ -70,7 +70,7 @@ var Person = MediaType("application/vnd.goa.person", func() {
 		Attribute("age", Integer, "Age of person")
 		Required("name")
 	})
-	View("default", func() { // View defines a rendering of the media type.
+	View("default", func() {  // View defines a rendering of the media type.
 		Attribute("id")   // Media types may have multiple views and must
 		Attribute("href") // have a "default" view.
```
jim-slattery-rs: should these actually just be `name` and `age`? Not sure where `id` and `href` come from.

nitinmohan87: It must be `name` and `age`. Good catch!

jim-slattery-rs: Fixed 👍

raphael:
Whoops, thank you!

Issue #2056, opened 2019-03-25
Support for custom errors in gRPC


nitinmohan87:
We must support encoding/decoding custom errors that are user types in gRPC. We could encode the custom error type in the status details like we do with standard errors - https://github.com/goadesign/goa/blob/v2/grpc/error.go#L62-L65
We could make the custom error type implement the protocol buffer message interface which is all is needed to encode as a status detail.

PR #2071, opened 2019-04-04
Support custom errors gRPC

nitinmohan87:
#2056

### DIFF
--- a/expr/service.go
+++ b/expr/service.go
@@ -119,6 +119,9 @@ func (e *ErrorExpr) Validate() error {
 					continue
 				}
 				errField = n.Name
+				if !rt.IsRequired(n.Name) || rt.HasDefaultValue(n.Name) {
+					verr.Add(e, "attribute %q with 'struct:error:name' in the meta must be required or have a default value", n.Name)
+				}
 			}
 		}
 		if errField == "" {
--- a/expr/service.go
+++ b/expr/service.go
@@ -128,27 +128,16 @@ func (e *ErrorExpr) Finalize() {
 		if dt != ErrorResult {
 			// If this type contains an attribute with "struct:error:name" meta
 			// then no need to do anything.
-			var found bool
 			for _, nat := range *AsObject(dt) {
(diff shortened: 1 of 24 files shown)


expr/service.go
```diff
@@ -133,7 +123,34 @@ func (e *ErrorExpr) Validate() error {
...
-	if _, ok := att.Type.(UserType); !ok {
+	switch dt := att.Type.(type) {
+	case UserType:
+		if dt != ErrorResult {
+			// If this type contains an attribute with "struct:error:name" meta
+			// then no need to do anything.
+			var found bool
+			for _, nat := range *AsObject(dt) {
+				if _, ok := nat.Attribute.Meta["struct:error:name"]; ok {
+					found = true
```
raphael: Why not just `return` here? no need for `found `at all?

expr/service.go
```diff
@@ -133,7 +123,34 @@ func (e *ErrorExpr) Validate() error {
...
+			}
+			if found {
+				return
+			}
+
+			// This type does not have an attribute with "struct:error:name" meta.
+			// It means the type is used by at most one error (otherwise validations
+			// would have failed). Dup type and add "struct:error:name" meta to
+			// indicate that the type is an error type.
+			dt = Dup(att.Type).(UserType)
```
raphael: Not sure we need to dup the type? duping is expensive and as you know can have interesting side effects (e.g. generated types etc.)

grpc/error.go
```diff
@@ -30,7 +30,11 @@ type (
 	}
 )
 
-// NewErrorResponse create a gRPC error response from the given error.
+// NewErrorResponse creates a new ErrorResponse protocol buffer message from
+// the given error. If the given error is a goa ServiceError, the ErrorResponse
+// message will be set with the corresponding Timeout, Temporary, and Fault
+// characteristics. If the error is not a goa ServiceError, it creates an
+// ErrorResponse message with the Fault characteristic set.
```
raphael: with the Fault field set to true.

PR #2073, opened 2019-04-10
Add some predeclared types and constants to Go reserved words

tmaflexera:


raphael:
This is great, thank you!

PR #2074, opened 2019-04-11
Add WWW-Authenticate header to 401 response if using Basic auth

tmaflexera:
If Basic authentication is in effect, and a StatusUnauthorized response is about to be returned, add
`WWW-Authenticate: Basic realm="$Method.Name"`
header to cause the browser to prompt for username and password.
This applies whenever an error that is mapped to StatusUnauthorized will be returned, due to
1. the authorization logic returning an error that is mapped to StatusUnauthorized, or
2. the username or password field of the payload is marked required and was not supplied, and "missing_field" is mapped to StatusUnauthorized in the design.

If you think this is a reasonable change, let me know and I can add a test to http/codegen/testdata

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -1106,11 +1106,15 @@ func {{ .ErrorEncoder }}(encoder func(context.Context, http.ResponseWriter) goah
 			return encodeError(ctx, w, v)
 		}
 		switch en.ErrorName() {
+	{{- $basicScheme := .BasicScheme }}{{- $realm := .Method.Name }}
 	{{- range $gerr := .Errors }}
 	{{- range $err := .Errors }}
 		case {{ printf "%q" .Name }}:
 			res := v.({{ $err.Ref }})
 			{{- with .Response}}
+				{{- if and $basicScheme (eq .StatusCode "http.StatusUnauthorized") }}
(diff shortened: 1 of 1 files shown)


tmaflexera:
It looks like I'll need to update the example output too in a couple of places

nitinmohan87:
@tmaflexera Thanks for the PR! However I don't think this should be done in goa. It is the responsibility of the application to add any special headers using custom encoders.

raphael:
Thank you Tim! I would tend to agree with Nitin, this seems more of a front service use case (i.e. public API hit by browsers) and not necessarily something that would always apply (e.g probably not for requests sent in between 2 back services). I would probably have that done in the specific authorization middleware instead.

tmaflexera:
Thanks Nitin, your suggestion worked without needing a custom encoder. Closing this PR

```go
var Unauthorized = Type("unauthorized", func() {
	Attribute("wwwauth_header", String)
	Attribute("message", String)
	Description("Request lacks valid authentication credentials for resource.")
	Required("wwwauth_header", "message")
})

var _ = Service("abc", func() {
	Description("The abc service serves data...")
	Error("unauthorized", Unauthorized, "Credentials are invalid")

	HTTP(func() {
		Path("/api")
		Response("unauthorized", StatusUnauthorized, func() {
			Header("wwwauth_header:WWW-Authenticate")
		})
	})
```

PR #2077, opened 2019-04-19
Add support for Params and Headers DSL

nitinmohan87:
* Fixes bug where adding `Params` or `Headers` DSL inside HTTP endpoint had no effect
* All headers are encoded/decoded in the [canonical header key format](https://godoc.org/net/http#CanonicalHeaderKey).
* Fixed DSL documentation for `Headers` and `Params` to take only function type as argument. It doesn't make sense to use a UserType here to define params and headers since we map the params and headers to the service payload and result types.
* Path parameters are now always required flags in the HTTP CLI

### DIFF
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -1107,30 +1107,40 @@ func buildPayloadData(e *expr.HTTPEndpointExpr, sd *ServiceData) *PayloadData {
 			for _, sc := range r.Schemes {
 				if sc.Type == "Basic" {
 					uatt := e.MethodExpr.Payload.Find(sc.UsernameAttr)
+					uref := svc.Scope.GoTypeRef(uatt)
+					if sc.UsernamePointer {
+						uref = "*" + uref
+					}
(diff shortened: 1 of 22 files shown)


http/codegen/service_data.go
```diff
@@ -2143,13 +2168,18 @@ func extractPathParams(a *expr.MappedAttributeExpr, service *expr.AttributeExpr,
 		if !expr.IsObject(service.Type) {
 			fieldName = ""
 		}
+		fieldPtr := false
```
raphael: Couldn't this just be:
```go
fieldPtr := expr.IsObject(service.Type) && service.IsPrimitivePointer(name, true)
```

http/codegen/service_data.go
```diff
@@ -2187,11 +2217,16 @@ func extractQueryParams(a *expr.MappedAttributeExpr, service *expr.AttributeExpr
...
```
raphael: same as a bove

Issue #1114, opened 2017-03-13
Error xml: unsupported type: map[string]interface {}


EvgenyGavrilov:
Hello!
I get error "xml: unsupported type: map[string]interface {}", when I make a GET request with a non-valid parameter.
```go
Param("p", Integer, func() {
	Description("Номер страницы")
	Minimum(1)
	Default(1)
})
```
GET Request URL:http://localhost:3000/api/attribute?p=!
Accept:application/xml
Accept-Encoding:gzip, deflate, sdch, br
Accept-Language:ru,en-US;q=0.8,en;q=0.6
Cache-Control:no-cache
Connection:keep-alive
Host:localhost:3000
Row with error https://github.com/goadesign/goa/blob/master/encoding.go#L215

raphael:
ugh yeah that makes sense. The workaround would be to override the package variable `goa.ErrInvalidRequest` so that it returns an object with no `metadata` field which is the one that's a `map[string]interface{}`. Maybe the `Metadata` field should have a `xml:"-"` tag on it...

EvgenyGavrilov:
If we do like this
```go
package main

import (
	"fmt"
	"encoding/xml"
	"os"
)

type (
	stringMap map[string]interface{}
	
	ErrorResponse struct {
		ID string `xml:"id"`
		Meta stringMap `xml:"meta,omitempty"`
	}
)

func (s stringMap) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
    tokens := []xml.Token{start}
    for k, v := range s {
	v := fmt.Sprintf("%v", v)
        t := xml.StartElement{Name: xml.Name{"", k}}
        tokens = append(tokens, t, xml.CharData(v), xml.EndElement{t.Name})
    }
    tokens = append(tokens, xml.EndElement{start.Name})
    for _, t := range tokens {
        err := e.EncodeToken(t)
        if err != nil {
            return err
        }
    }
    err := e.Flush()
    if err != nil {
        return err
    }

    return nil
}

func main() {
	m := make(stringMap, 2)
	m["key1"] = "value"
	m["key2"] = 1024
	e := xml.NewEncoder(os.Stdout)
	er := ErrorResponse{
		ID:  "id request",
		Meta: m,
	}
	if err := e.Encode(er); err != nil {
		fmt.Printf("%v\n", err)
	}
	
}
```

https://play.golang.org/p/deF7XB7zR2

raphael:
[...]

PR #2078, opened 2019-04-22
Fix error `xml: unsupported type: map[string]interface {}`

aita:
I fixed an error in XML marshalling of ErrorResponse which is reported with #1114.


raphael:
Thanks!

PR #2082, opened 2019-04-23
Add BasicAuth security in OpenAPI specs

nitinmohan87:
Goa implicitly sets Authorization header when an endpoint is secured with BasicAuth. But OpenAPI specs don't show the endpoint as secured at all. Hence we need to explicitly add the Authorization header in the OpenAPI spec when using BasicAuth.


expr/http_endpoint.go
```diff
@@ -660,44 +660,67 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
-		if IsObject(nat.Attribute.Type) {
-			verr.Add(e, "header %s cannot be an object, header type must be primitive or array", nat.Name)
-		} else if arr := AsArray(nat.Attribute.Type); arr != nil {
-			if !IsPrimitive(arr.ElemType.Type) {
-				verr.Add(e, "elements of array header %s must be primitive", nat.Name)
+
+	// We have to figure out the actual type for the params because the actual
+	// type is initialized only during the finalize phase. In the validation
+	// phase, all param types are string type by default unless specified
+	// expliclty.
```
raphael: typo

expr/http_endpoint.go
```diff
@@ -660,44 +660,67 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
+	init := func(ma *MappedAttributeExpr) *MappedAttributeExpr {
+		ma = DupMappedAtt(ma)
```
raphael: Can this be encapsulated in a private function that is used by both the validation and finalization steps?

expr/http_endpoint.go
```diff
@@ -660,44 +660,67 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
+
+	headers := init(e.Headers)
+	WalkMappedAttr(headers, func(name, _ string, a *AttributeExpr) error {
+		switch {
+		case IsObject(a.Type):
+			verr.Add(e, "header %s cannot be an object, header type must be primitive or array", name)
+		case IsArray(a.Type):
+			arr := AsArray(a.Type)
+			if !IsPrimitive(arr.ElemType.Type) {
+				verr.Add(e, "elements of array header %s must be primitive", name)
```
raphael: `%q` might be better than `%s`?

expr/http_endpoint.go
```diff
@@ -660,44 +660,67 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
-			}
-		case *Map:
-			if len(*headers) > 0 {
-				verr.Add(e, "Payload type is map but HTTP endpoint defines headers. Map payloads can only be decoded from HTTP request bodies or query strings.")
+
+	headers := init(e.Headers)
+	WalkMappedAttr(headers, func(name, _ string, a *AttributeExpr) error {
+		switch {
+		case IsObject(a.Type):
+			verr.Add(e, "header %s cannot be an object, header type must be primitive or array", name)
```
raphael: `%q` might be better than `%s`?

expr/http_endpoint.go
```diff
@@ -660,44 +660,67 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
+		case IsArray(a.Type):
+			arr := AsArray(a.Type)
+			if !IsPrimitive(arr.ElemType.Type) {
+				verr.Add(e, "elements of array header %s must be primitive", name)
 			}
+		default:
+			ctx := fmt.Sprintf("header %s", name)
```
raphael: `%q` might be better than `%s`?

http/codegen/openapi/openapi_v2_builder.go
```diff
@@ -158,8 +158,8 @@ func securitySpecFromExpr(root *expr.RootExpr) map[string]*SecurityDefinition {
 			sd.Type = "basic"
 		case expr.APIKeyKind:
 			sd.Type = "apiKey"
-			sd.In = s.In
-			sd.Name = s.Name
+			sd.In = "header"
```
raphael: Why are we hardcoding here? API keys can be set in parameters as well for example (maybe we need a test for that).

nitinmohan87: This defines the top-level `securityDefinitions` object. It is possible that the there are two endpoints secured by API key where one passes the key in the header and the other passes in the param, correct?

expr/http_endpoint.go
```diff
@@ -638,44 +628,51 @@ func (e *HTTPEndpointExpr) validateParams() *eval.ValidationErrors {
...
 	verr := new(eval.ValidationErrors)
-	for _, nat := range *headers {
-		if IsObject(nat.Attribute.Type) {
-			verr.Add(e, "header %s cannot be an object, header type must be primitive or array", nat.Name)
-		} else if arr := AsArray(nat.Attribute.Type); arr != nil {
+
+	// We have to figure out the actual type for the headers because the actual
+	// type is initialized only during the finalize phase. In the validation
+	// phase, all param types are string type by default unless specified
+	// explicity.
```
raphael: Typo

PR #2084, opened 2019-04-26
Add tests for expr.UserTypeExpr.Name()

tchssk:
#1423

### DIFF
--- a/expr/user_type_test.go
+++ b/expr/user_type_test.go
@@ -2,6 +2,53 @@ package expr
 
 import "testing"
 
+func TestUserTypeExprName(t *testing.T) {
+	var (
+		userTypeExprWithoutAttribute = UserTypeExpr{
+			TypeName: "foo",
+		}
+		userTypeExprHasMeta = UserTypeExpr{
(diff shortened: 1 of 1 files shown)


expr/user_type_test.go
```diff
@@ -2,6 +2,53 @@ package expr
...
+		},
+		"user type has another meta": {
+			userType: userTypeExprHasAnotherMeta,
+			expected: "foo",
+		},
+	}
+
+	for k, tc := range cases {
+		if actual := tc.userType.Name(); actual != tc.expected {
+			t.Errorf("%s: got %#v, expected %#v", k, actual, tc.expected)
```
nitinmohan87: Let's use `t.Run` pattern for new tests
```
for k, tc := range cases {
  t.Run(k, func(t *testing.T) {
    ...
    t.Errorf("got %#v, expected %#v", actual, tc.expected)
  })
}
```

tchssk: Woops, fixed it.

raphael:
Great, thank you!

PR #2120, opened 2019-05-24
Allow security scopes on BasicAuth and APIKey authentication

### LINKED ISSUE #2105: v3: Using BasicAuth with custom scopes
As mentioned in #1959, we kinda 'misuse' BasicAuth in v1 by supplying scopes to it. This allows us to extract the scopes via goa.ContextRequiredScopes(ctx) in the middleware, authenticate the user against LDAP and match the scopes against LDAP groups. It even generates correct swagger, which contains the required scope, e.g.: BasicAuth (users:list). As far as I can see, v3 does not allow scopes to be defined on Security(BasicAuth) and also requires to have username and password payload defined on every method, instead of handling everything in middleware only.

Is there any approach to move this to v2/v3 or any idea how we could solve this? :smiley:

cubic3d:
This PR resolves #2105 and allows to define `Scope` on `BasicAuth` and `APIKeyAuth` definitions.

Description on endpoint now contains the `Security` name to distinguish between the methods. Additionally consistent formatting is applied to descriptions containing security scopes.


dsl/security.go
```diff
@@ -519,6 +519,8 @@ func TokenField(tag interface{}, name string, args ...interface{}) {
 // Scope has two uses: in JWTSecurity or OAuth2Security it defines a scope
 // supported by the scheme. In Security it lists required scopes.
 //
+// Scope can be defined in BasicAuth and APIKeyAuth for custom usage.
```
raphael: Please just add these two expressions to the list below (`Scope must appear in …`).

http/codegen/openapi/openapi_v2_builder.go
```diff
@@ -159,10 +159,34 @@ func securitySpecFromExpr(root *expr.RootExpr) map[string]*SecurityDefinition {
 					switch s.Kind {
 					case expr.BasicAuthKind:
 						sd.Type = "basic"
+						// Generate scopes to add to description
```
raphael: Would it make sense to refactor these snippets into a helper function something like:
```go
func addScopeDescription(scopes ..., sd *ServiceData) {
    lines := []string{}
    …
        sd.Description += fmt.Sprintf…
    }
}
```
so that it could be called in all these places?

raphael (commented):
This looks great, thank you! I left a couple of suggestions.

dsl/security.go
cubic3d: Changed ;)

http/codegen/openapi/openapi_v2_builder.go
cubic3d: Refactored in a separate function

cubic3d:
Thank you! I made the changes.

raphael:
Looks great! thank you. Do you think you could backport these changes to v2 as well? Simply cherry-pick the merge commit into a branch created off of v2. Thank you!

cubic3d:
@raphael Sorry for the late reaction, had some time off ;) back port in #2162

PR #2122, opened 2019-05-24
generate code that can support a TLS enabled server

kmeehl:
Passes the TLS certificate and key paths through from command line


kmeehl:
Note: I assume there is a way to regenerate updated test data, but I do not yet know how. So this PR does not include updated test data.

codegen/example/example_server.go
```diff
@@ -157,10 +157,22 @@ func main() {
 	{{- range .Server.Variables }}
 	{{ .VarName }}F = flag.String({{ printf "%q" .Name }}, {{ printf "%q" .DefaultValue }}, "{{ .Description }}{{ if .Values }} (valid values: {{ join .Values ", " }}){{ end }}")
 	{{- end }}
+		tlsCertPath   = flag.String("tls-cert", "", "TLS certificate file path (required in secure mode)")
```
raphael, line 160: These should only be added if needed, i.e. if HTTPS or TLS with gRPC ("grpcs" scheme) is being used.

codegen/example/example_server.go
```diff
@@ -278,7 +290,7 @@ func main() {
 			} else if u.Port() == "" {
 				u.Host += ":{{ $u.Port }}"
 			}
-			handle{{ toUpper $u.Transport.Name }}Server(ctx, u, {{ range $.Services }}{{ if .Methods }}{{ .VarName }}Endpoints, {{ end }}{{ end }}&wg, errc, logger, *dbgF)
+			handle{{ toUpper $u.Transport.Name }}Server(ctx, u, tlsCertPath, tlsKeyPath, {{ range $.Services }}{{ if .Methods }}{{ .VarName }}Endpoints, {{ end }}{{ end }}&wg, errc, logger, *dbgF)
```
raphael, line 293: These should only be arguments if needed.

http/codegen/example_server.go
```diff
@@ -189,7 +189,7 @@ func {{ .FuncName }}(mw *multipart.Writer, p {{ .Payload.Ref }}) error {
 
 	// input: map[string]interface{}{"Services":[]*ServiceData}
 	httpSvrStartT = `{{ comment "handleHTTPServer starts configures and starts a HTTP server on the given URL. It shuts down the server if any error is received in the error channel." }}
-func handleHTTPServer(ctx context.Context, u *url.URL{{ range $.Services }}{{ if .Service.Methods }}, {{ .Service.VarName }}Endpoints *{{ .Service.PkgName }}.Endpoints{{ end }}{{ end }}, wg *sync.WaitGroup, errc chan error, logger *log.Logger, debug bool) {
+func handleHTTPServer(ctx context.Context, u *url.URL, tlsCertPath *string, tlsKeyPath *string {{ range $.Services }}{{ if .Service.Methods }}, {{ .Service.VarName }}Endpoints *{{ .Service.PkgName }}.Endpoints{{ end }}{{ end }}, wg *sync.WaitGroup, errc chan error, logger *log.Logger, debug bool) {
```
raphael, line 192: These arguments should only be defined if needed. Another approach could be to define a `handleHTTPSServer` with this signature and invoke that instead in case the `https` scheme is being used.

raphael (changes requested):
Thank you for the PR! I left a few comments that basically boil down to: the TLS specific arguments should only be defined and required if there is a HTTPS server.

codegen/example/example_server.go
kmeehl: I can only call `flag.Parse` once. Otherwise I run into the following scenario: 
- starting the server in non-tls mode works without error
- starting the server by only passing `-secure` will yield the error `missing required argument argument -tls-cert` (as expected)
- starting the server by passing `-secure` and the tls paths will yield the error `flag provided but not defined: -tls-cert`

In other words, I must define this flag up front, or the flag package will throw an error.

codegen/example/example_server.go
kmeehl: Continuing from my response above, I'm forced to define the TLS paths up front for flag parsing to work correctly. In the case that TLS is not being used, these arguments pass through harmlessly as empty strings, and are further ignored inside `handleHTTPServer`.

If you feel strongly about it, I can define a `handleHTTPServer` with a footprint omitting the arguments, but the implementation of that function would most logically be to simply call `handleHTTPServer` passing empty strings for the TLS paths. I personally did not see much value in that, but I'll make the change if you'd prefer it.

codegen/example/example_server.go
raphael: Sorry I should have been more clear: I meant that the code that define these flags should only be generated in the case where TLS is required (so there is a "https" scheme in one of the server URIs). The reason for this is that many setup have TLS terminated by a load balancer and thus it's actually pretty common to have server URIs that do not use https. In this common case we don't want to define command line flags that are not used. So there should be something like:
```go
{{ if hasTLS . }} // or something else...
tlsCertPath   = flag.String("tls-cert", "", "TLS certificate file path (required in secure mode)")
// ...
```
This means that code elsewhere that uses these flags also need to be protected with the same guard.

codegen/example/example_server.go
raphael: As per the above what I'm hoping to avoid is having the code define these flags when no URI uses TLS. Agree that it's really more of a nicety but it doesn't seem too complicated to add and will probably match expectations better.

raphael:
@kmeehl I'm wondering whether to close this PR? Let me know if you are still interested in proceeding with this, thanks!

raphael:
Closing for now.

PR #2129, opened 2019-05-29
Fixed ArrayOf(UInt32) conversion to use the right cast

cquinn:
The generated DecodeXyz function for uint32 arrays was using an int32() cast instead of uint32.


raphael:
Whoops, good catch - thank you! Looks like there's another test that needs to be updated: https://travis-ci.org/goadesign/goa/builds/539002047#L778. Would you mind updating that as well?

cquinn:
I was unable to run the build + tests locally, otherwise, I wouldn't be using Travis as my only checker :)

cquinn:
Oh, it looks like the latest test failure is in checking the cellar example, and maybe that test is out of date with master?

raphael:
Yes that was the case, updated the examples and reran the tests - all green now! thank you.

raphael:
Sorry, one last thing: do you think you could backport this fix to v2? Simply cherry-pick the merge commit in a branch created off of v2. Thank you!

Issue #2127, opened 2019-05-29
v2: crash when two fields use the same type


tmcladam:
Use this design to reproduce:

[repro2.go.txt](https://github.com/goadesign/goa/files/3232685/repro2.go.txt)

Error message / stack trace:

[crash.log.txt](https://github.com/goadesign/goa/files/3232715/crash.log.txt)

Suggested patch:

[crash.diff.txt](https://github.com/goadesign/goa/files/3232722/crash.diff.txt)

A nil TransformFunctionData is returned when the function has already been seen. This should either not be appended to the array (as in the patch above), or nil elements should be ignored whenever iterating over the array.

raphael:
Good catch, makes sense! Would you mind putting a couple of PRs together (one for `v2` and one for `v3`)? Thank you!

tmcladam:
Done, cheers

PR #2133, opened 2019-05-30
Fix crash when a user type is used in multiple fields (v3)

tmcladam:
Fix for https://github.com/goadesign/goa/issues/2127 in v3


raphael:
Thank you!

PR #2144, opened 2019-06-06
Reserve 'c' for the client's receiver name

### LINKED ISSUE #2138: V3: Conflict between variable name and receiver name
It causes conflict between variable name and receiver name if path includes wildcard named ‘c’.

# Example design
```go
    package design
    
    import (
       . "goa.design/goa/v3/dsl"
    )
    
    var _ = API("calc", func() {
       Title("Calculator Service")
       Description("Service for adding numbers, a Goa teaser")
       Server("calc", func() {
          Host("localhost", func() {
             URI("http://localhost:8000")
          })
       })
    })
    
       Method("add", func() {
          Payload(func() {
             Field(2, "c", Int, "Right operand")
          })
    
          Result(Int)
    
          HTTP(func() {
             GET("/add/{a}/{c}")  // ← ★
             //GET("/add/{a}")  // ok
          })
       })
    })
```

# Build error

```go
$ go build ./cmd/calc
# calc/gen/http/calc/client
gen/http/calc/client/encode_decode.go:26:3: c redeclared in this block
	previous declaration at gen/http/calc/client/encode_decode.go:23:86
gen/http/calc/client/encode_decode.go:40:25: c.scheme undefined (type int has no field or method scheme)
gen/http/calc/client/encode_decode.go:40:41: c.host undefined (type int has no field or method host)
```

# Generated code

gen/http/calc/client/encode_decode.go
```go
    // BuildAddRequest instantiates a HTTP request object with method and path set
    // to call the "calc" service "add" endpoint
    func (c *Client) BuildAddRequest(ctx context.Context, v interface{}) (*http.Request, error) {
       var (
          a int
          c int // ← ★
       )
       {
          p, ok := v.(*calc.AddPayload)
          if !ok {
             return nil, goahttp.ErrInvalidType("calc", "add", "*calc.AddPayload", v)
          }
          if p.A != nil {
             a = *p.A
          }
          if p.C != nil {
             c = *p.C
          }
       }
       u := &url.URL{Scheme: c.scheme, Host: c.host, Path: AddCalcPath(a, c)}
       req, err := http.NewRequest("GET", u.String(), nil)
       if err != nil {
          return nil, goahttp.ErrInvalidURL("calc", "add", u.String(), err)
       }
       if ctx != nil {
          req = req.WithContext(ctx)
       }
    
       return req, nil
    }
```

ikawaha:
Fix #2138.

# Design
```

	Method("add", func() {
		Payload(func() {
			Field(2, "c", Int, "Right operand")
		})

		Result(Int)

		HTTP(func() {
			GET("/add/{a}/{c}")  // ← ★
		})
	})
})

```
# Generated code
```go
// BuildAddRequest instantiates a HTTP request object with method and path set
// to call the "calc" service "add" endpoint
func (c *Client) BuildAddRequest(ctx context.Context, v interface{}) (*http.Request, error) {
	var (
		a  int
		c2 int // ← ★
	)
	{
		p, ok := v.(*calc.AddPayload)
		if !ok {
			return nil, goahttp.ErrInvalidType("calc", "add", "*calc.AddPayload", v)
		}
		if p.A != nil {
			a = *p.A
		}
		if p.C != nil {
			c2 = *p.C
		}
	}
	u := &url.URL{Scheme: c.scheme, Host: c.host, Path: AddCalcPath(a, c2)}
	req, err := http.NewRequest("GET", u.String(), nil)
	if err != nil {
		return nil, goahttp.ErrInvalidURL("calc", "add", u.String(), err)
	}
	if ctx != nil {
		req = req.WithContext(ctx)
	}

	return req, nil
}

```

### DIFF
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -769,8 +769,12 @@ func (d ServicesData) analyze(hs *expr.HTTPServiceExpr) *ServiceData {
 			)
 			{
 				name = fmt.Sprintf("Build%sRequest", ep.VarName)
+				s := codegen.NewNameScope()
+				s.Unique("c") // 'c' is reserved as the client's receiver name.
 				for _, ca := range routes[0].PathInit.ClientArgs {
 					if ca.FieldName != "" {
+						ca.Name = s.Unique(ca.Name, "_")
+						ca.Ref = ca.Name
 						args = append(args, ca)
 					}
 				}
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -2191,7 +2191,7 @@ func extractPathParams(a *expr.MappedAttributeExpr, service *expr.AttributeExpr,
 			StringSlice:    arr != nil && arr.ElemType.Type.Kind() == expr.StringKind,
 			Map:            false,
 			MapStringSlice: false,
-			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn),
+			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn, name),
 			DefaultValue:   c.DefaultValue,
 			Example:        c.Example(expr.Root.API.Random()),
 		})
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -610,12 +610,12 @@ func (svc *ServiceData) Endpoint(name string) *EndpointData {
 	return nil
 }
 
-func newHTTPServiceData(svc *service.Data) *ServiceData {
-	// 'c' is reserved as the client's receiver name.
-	scope := codegen.NewNameScope()
-	scope.Unique("c")
+// analyze creates the data necessary to render the code of the given service.
+// It records the user types needed by the service definition in userTypes.
+func (d ServicesData) analyze(hs *expr.HTTPServiceExpr) *ServiceData {
+	svc := service.Services.Get(hs.ServiceExpr.Name)
 
-	return &ServiceData{
+	rd := &ServiceData{
 		Service:          svc,
 		ServerStruct:     "Server",
 		MountPointStruct: "MountPoint",
@@ -625,15 +625,8 @@ func newHTTPServiceData(svc *service.Data) *ServiceData {
 		ClientStruct:     "Client",
 		ServerTypeNames:  make(map[string]bool),
 		ClientTypeNames:  make(map[string]bool),
-		Scope:            scope,
+		Scope:            codegen.NewNameScope(),
 	}
-}
-
-// analyze creates the data necessary to render the code of the given service.
-// It records the user types needed by the service definition in userTypes.
-func (d ServicesData) analyze(hs *expr.HTTPServiceExpr) *ServiceData {
-	svc := service.Services.Get(hs.ServiceExpr.Name)
-	rd := newHTTPServiceData(svc)
 
 	for _, s := range hs.FileServers {
 		paths := make([]string, len(s.RequestPaths))
@@ -2191,7 +2184,7 @@ func extractPathParams(a *expr.MappedAttributeExpr, service *expr.AttributeExpr,
 			StringSlice:    arr != nil && arr.ElemType.Type.Kind() == expr.StringKind,
 			Map:            false,
 			MapStringSlice: false,
-			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn, name),
+			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn),
 			DefaultValue:   c.DefaultValue,
 			Example:        c.Example(expr.Root.API.Random()),
 		})
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -610,7 +610,7 @@ func (svc *ServiceData) Endpoint(name string) *EndpointData {
 	return nil
 }
 
-func newHTTPServiceData(svc *service.Data) *ServiceData{
+func newHTTPServiceData(svc *service.Data) *ServiceData {
 	// 'c' is reserved as the client's receiver name.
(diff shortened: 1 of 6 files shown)


http/codegen/service_data.go
```diff
@@ -2184,7 +2191,7 @@ func extractPathParams(a *expr.MappedAttributeExpr, service *expr.AttributeExpr,
 			StringSlice:    arr != nil && arr.ElemType.Type.Kind() == expr.StringKind,
 			Map:            false,
 			MapStringSlice: false,
-			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn),
+			Validate:       codegen.RecursiveValidationCode(c, ctx, true, varn, name),
```
ikawaha: Setting original path param name, to avoid setting escaped name to the validation target.

before:
```
err = goa.MergeErrors(err, goa.ValidatePattern("c2", c2, "patternc")) 
```
after:
```
err = goa.MergeErrors(err, goa.ValidatePattern("c", c2, "patternc"))
```

---

There are several places where it is better to use the original name.
If this PR is adopted, correct them as well.
```
$ git grep codegen.RecursiveValidationCode|grep varn
grpc/codegen/service_data.go:			Validate:     codegen.RecursiveValidationCode(c, ctx, required, varn),
http/codegen/service_data.go:					Validate:       codegen.RecursiveValidationCode(pAtt, httpsvrctx, required, varn),
http/codegen/service_data.go:			Validate:     codegen.RecursiveValidationCode(c, ctx, required, varn),
http/codegen/service_data.go:			Validate:      codegen.RecursiveValidationCode(hattr, svcCtx, required, varn),
```

nitinmohan87 (changes requested):
Thanks for the PR!

Since you introduced a new parameter to `RecursiveValidationCode` this becomes an interface-breaking change. I am thinking if we can try to avoid making interface-breaking change for this particular bug.

Only the `BuildXXXRequest` gets this problem because of the client handle variable `c`. So we can simply modify [this particular section](https://github.com/goadesign/goa/blob/v3/http/codegen/service_data.go#L763-L805) to not conflict with the client handle variable, instead of doing it all over the generated code. I think it is just about [checking if any of the `routes[0].PathInit.ClientArgs` has a name `c`](https://github.com/goadesign/goa/blob/v3/http/codegen/service_data.go#L772-L776) and producing a new name for it. Also, we may need to modify [this template](https://github.com/goadesign/goa/blob/v3/http/codegen/service_data.go#L2562-L2605) to use `.Args` instead of `.PathInit.ClientArgs` where necessary. I think this will eliminate any interface-breaking changes. Please let me know if that makes sense!

ikawaha:
Thank you for your review!
I tried to fix it in the way you suggested. I hope it makes sense.
There may be a better way. If so please close this pr.
Thank you.

---

Example desgin
```go

package design

import (
   . "goa.design/goa/v3/dsl"
)

var _ = API("calc", func() {
   Title("Calculator Service")
   Description("Service for adding numbers, a Goa teaser")
   Server("calc", func() {
      Host("localhost", func() {
         URI("http://localhost:8000")
      })
   })
})

   Method("add", func() {
      Payload(func() {
         Field(1, "a", Int)
         Field(5, "b", Int)
         Field(2, "c", Int)
         Field(3, "c1", Int)
         Field(4, "c_", Int)
      })
      Result(Int)
      HTTP(func() {
         GET("/add/{a}/{b}/{c}/{c1}/{c_}")  // ← ★
      })
   })
})
```

Generated code
```go

// BuildAddRequest instantiates a HTTP request object with method and path set
// to call the "calc" service "add" endpoint
func (c *Client) BuildAddRequest(ctx context.Context, v interface{}) (*http.Request, error) {
   var (
      a   int
      b   int
      c_  int
      c1  int
      c_2 int
   )
   {
      p, ok := v.(*calc.AddPayload)
      if !ok {
         return nil, goahttp.ErrInvalidType("calc", "add", "*calc.AddPayload", v)
      }
      if p.A != nil {
         a = *p.A
      }
      if p.B != nil {
         b = *p.B
      }
      if p.C != nil {
         c_ = *p.C
      }
      if p.C1 != nil {
         c1 = *p.C1
      }
      if p.C != nil {
         c_2 = *p.C
      }
   }
   u := &url.URL{Scheme: c.scheme, Host: c.host, Path: AddCalcPath(a, b, c_, c1, c_2)}
   req, err := http.NewRequest("GET", u.String(), nil)
   if err != nil {
      return nil, goahttp.ErrInvalidURL("calc", "add", u.String(), err)
   }
   if ctx != nil {
      req = req.WithContext(ctx)
   }

   return req, nil
}
```

http/codegen/service_data.go
```diff
@@ -769,8 +769,12 @@ func (d ServicesData) analyze(hs *expr.HTTPServiceExpr) *ServiceData {
 			)
 			{
 				name = fmt.Sprintf("Build%sRequest", ep.VarName)
+				s := codegen.NewNameScope()
+				s.Unique("c") // 'c' is reserved as the client's receiver name.
 				for _, ca := range routes[0].PathInit.ClientArgs {
 					if ca.FieldName != "" {
+						ca.Name = s.Unique(ca.Name, "_")
```
nitinmohan87: Instead of using `Unique` you can use `Name` which will automatically suffix a counter variable to the name. You don't have to suffix with `_`
```
ca.Name = s.Name(ca.Name)
```

nitinmohan87:
Looks great! One minor comment and then this is good to go.

nitinmohan87 (approved):

raphael:
Thank you!

raphael:
@ikawaha do you think you could backport this fix to `v2`? (simply cherry-pick the merge commit into a branch created off of `v2`). Thank you!

Issue #2151, opened 2019-06-14
v3: NameScope Unique() returns not unique values


ikawaha:
It seems that NameScope Unique() sometimes returns not unique values. 
Is this a bug?

https://play.golang.org/p/ANFB4Xgx3Ta

raphael:
Thank you for the fix!

PR #2152, opened 2019-06-15
Fix NameScope Unique method

ikawaha:
see #2151


nitinmohan87 (approved):
Looks good to me. Will let @raphael make the final say.

raphael:
Looks great, thank you for the fix!

raphael:
@ikawaha do you think you could backport this fix to `v2`? (simply cherry-pick the merge commit into a branch created off of `v2`). Thank you!

PR #2158, opened 2019-06-19
Fix validation for attribute with recursive extends

### LINKED ISSUE #2146: Extending recursively does not define JWT token
Extending a type that defines token works but extending a type that extends a type that defines token does not work. Example:
`$GOPATH/src/foo/design/api.go`
```
package design

import (
	. "goa.design/goa/v3/dsl"
)

var JWTAuth = JWTSecurity("jwt", func() {
	Scope("foo:write")
})

	Token("token", String)
})

var Intermediary = Type("Intermediary", func() {
	Extend(CommonAttributes)
})

var _ = Service("Foo", func() {
	HTTP(func() {
		Path("/foo")
	})

	Method("create", func() {
		Security(JWTAuth, func() {Scope("foo:write")})
		Payload(func() {
			Extend(Intermediary)
			Attribute("bar", String)
			Required("bar")
		})
		Result(Empty)

		HTTP(func() {
			POST("/")
			Response(StatusCreated)
		})
	})
})
```
Code generation fails:
```
$ goa gen foo/design
exit status 1
service "Foo" HTTP endpoint "create": Payload must define JWT token attribute (use Token) required by security scheme "jwt".
```
If you say `Extend(CommonAttributes)` directly then it succeeds.

nitinmohan87:
Fixes #2146

### DIFF
--- a/expr/attribute.go
+++ b/expr/attribute.go
@@ -159,7 +159,11 @@ func TaggedAttribute(a *AttributeExpr, tag string) string {
 		}
 	}
 	for _, b := range a.Bases {
-		if n := TaggedAttribute(&AttributeExpr{Type: b}, tag); n != "" {
+		at := &AttributeExpr{Type: b}
+		if ut, ok := b.(UserType); ok {
+			at = ut.Attribute()
+		}
+		if n := TaggedAttribute(at, tag); n != "" {
 			return n
 		}
 	}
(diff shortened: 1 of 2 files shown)


raphael:
Thank you!

raphael:
@nitinmohan87 do you think you could backport this fix to `v2`? (simply cherry-pick the merge commit into a branch created off of `v2`). Thank you!

PR #2159, opened 2019-06-20
Goify response definitions in openapi

### LINKED ISSUE #2118: v3: $ref generated in OpenAPI are not RFC3986 compliant
For a method named `Resolve IP` the generated references look like:
```
$ref: '#/definitions/DNSResolve IPNot FoundResponseBody'
```
Those must be percent-encoded as [URIs](https://www.ietf.org/rfc/rfc3986.txt) (the white-spaces in that case) or changed to underscores to avoid.

nitinmohan87:
Fixes #2118


raphael:
This is great, thank you! Do you think you could backport this fix to `v2`? (simply cherry-pick the merge commit into a branch created off of `v2`). Thank you!

PR #2163, opened 2019-06-27
Fix sample code in README

johnmanjiro13:
Because goa just generates AddPayload struct in calc package, calcsvc doesn't have it.
I fixed sample code in README.


raphael:
Whoops, thank you!

PR #2165, opened 2019-06-28
Issue 2157 - Add support for text plain content type

ilmaruk:
Since the implementations of `Encode` and `Decode` for `http.textHtml*` and `http.textPlain*` would have been the very same, I opted for introducing a single `http.textEncoder` and `http.textDecoder`.

It should be easy enough to split them in the future, if needed.

I also took the chance of adding some tests around the code I modified.


raphael:
Thank you for the great PR, love the additional tests! Do you think you could backport this to `v2` as well? Simply cherry-pick the merge commit into a branch created off of `v2`, thank you!.

ilmaruk:
Sure, I'll do it ASAP.

PR #2166, opened 2019-07-01
V1: Porting #858 from master to v1

ikawaha:
The difference #858 is lost in v1.


raphael:
Good catch! thank you.

PR #2173, opened 2019-07-02
Issue 2157 - Add support for text plain content type (#2165)

ilmaruk:
WIP: porting changes from v3 to allow usage of `plain/text` content-type.


raphael:
Great, thank you!

PR #2172, opened 2019-07-02
V3: Fix the case where non-unified package names are generated

ikawaha:
Fix #2153 

### Example design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = API("calc", func() {
	Title("Calculator Service")
	Description("Service for adding numbers, a Goa teaser")
	Server("calc", func() {
		Host("localhost", func() {
			URI("http://localhost:8000")
			URI("grpc://localhost:8080")
		})
	})
})

var _ = Service("calc", func() {
	Description("The calc service performs operations on numbers.")

	Method("add", func() {
		Payload(func() {
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})

		Result(Int)

		HTTP(func() {
			GET("/add/{a}/{b}")
		})

		GRPC(func() {
		})
	})

	Files("/openapi.json", "./gen/http/openapi.json")
})

var _ = Service("calcapi", func() {
	Description("The calc service performs operations on numbers.")

	Method("add", func() {
		Payload(func() {
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})

		Result(Int)

		HTTP(func() {
			GET("/add/{a}/{b}")
		})

		GRPC(func() {
		})
	})

	Files("/openapi.json", "./gen/http/openapi.json")
})

var _ = Service("calcapi1", func() {
	Description("The calc service performs operations on numbers.")

	Method("add", func() {
		Payload(func() {
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})

		Result(Int)

		HTTP(func() {
			GET("/add/{a}/{b}")
		})

		GRPC(func() {
		})
	})

	Files("/openapi.json", "./gen/http/openapi.json")
})

```
### goa gen & goa example

```shellsession
$ goa gen calc/design
gen/calc/client.go
gen/calc/endpoints.go
gen/calc/service.go
gen/calcapi/client.go
gen/calcapi/endpoints.go
gen/calcapi/service.go
gen/calcapi1/client.go
gen/calcapi1/endpoints.go
gen/calcapi1/service.go
gen/grpc/calc/client/cli.go
gen/grpc/calc/client/client.go
gen/grpc/calc/client/encode_decode.go
gen/grpc/calc/client/types.go
gen/grpc/calc/pb/calc.proto
[...]

PR #2178, opened 2019-07-04
V2 back port: Fix the case where non-unified package names are generated (#2172)

ikawaha:
* Fix the case where non-unified package names are generated

* Add tests


raphael:
This is the one where we should remove `v3` from the import package paths.

ikawaha:
I was confused. Thank you for your comment!

raphael:
Thank you!

PR #2179, opened 2019-07-04
V3: Remove v3 dependencies from imports

ikawaha:
I made a mistake to describe the v3 dependency in #2172 .

### DIFF
--- a/codegen/service/example_svc_test.go
+++ b/codegen/service/example_svc_test.go
@@ -5,9 +5,9 @@ import (
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


raphael:
I think this was correct no? this is merging into the `v3` branch which is using Go modules. We need to remove these in the `v2` branch. Hmm which means we can't just cherry-pick into `v2` but I still believe we need to use `v3` for tests in the `v3` branch.

ikawaha:
It was misunderstood to be managed by go mod. You are right. Thank you!

Issue #2181, opened 2019-07-06
V3: The version of goa specified in go.mod is ignored and rewritten


ikawaha:
According to [The go blog: Using Go modules](https://blog.golang.org/using-go-modules), `go get` changes the required version of a dependency (or adds a new dependency).

Since `goa gen` (or `goa example`) executes `go get` at runtime, 
even if the goa version is specified in the go.mod file, it is ignored and rewritten.

Isn't this behavior causing confusion? There is also a problem that the latest 
goa specified by the latest commit number can not be used.

### Example:

**before**
go.mod
```
module github.com/ikawaha/calc

go 1.12

require (
...snip...
        goa.design/goa/v3 v3.0.3-0.20190704022140-85024ebc66dc  // latest version
)
```

**after goa gen/example**

```shellsession
$ goa gen calc/design
```

```
module github.com/ikawaha/calc

go 1.12

require (
...snip...
        goa.design/goa/v3 v3.0.2
)
```

raphael:
Thank you for the fix!

PR #2182, opened 2019-07-06
V3: Fix to use the version of goa specified in go.mod file, when 'goa gen'/'goa example'

ikawaha:
Fix issue #2181

### DIFF
--- a/cmd/goa/gen.go
+++ b/cmd/goa/gen.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
 	"go/parser"
 	"go/token"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -13,7 +15,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/sirkon/goproxy/gomod"
 	"goa.design/goa/v3/codegen"
 	"golang.org/x/tools/go/packages"
 )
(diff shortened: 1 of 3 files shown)


raphael:
Thank you for the great issue report and the PR! A couple of questions:

1. Would it make sense to search for "go.mod" recursively going up the directories to support the case where a single Go module hosts multiple Goa projects, e.g.:
```
root
├── go.mod
├── project1
│   └── design
└── project2
    └── design
```
Here the `goa` command would be run within `project1` and `project2` and it would be nice if it could pickup the `go.mod` file under `root` like the `go` command does.

2. Instead of adding a dependency on a third party package ("github.com/sirkon/goproxy/gomod") would it be better to just copy the `modfile` package from the Go source code: https://github.com/golang/go/tree/master/src/cmd/go/internal/modfile. It's unfortunate that it is an internal package but it seems better to copy and give proper attribution than to add a dependency on another package.

Thank you again, this is a great contribution!

ikawaha:
Thank you for your comment !

1. The location of go.mod to be used can be found by `go env`.
(cf. https://dave.cheney.net/2013/06/14/you-dont-need-to-set-goroot-really)

```
$ echo $GOMOD

$ go env GOMOD
/Users/ikawaha/go/src/github.com/ikawaha/calc/go.mod
```
So, I fixed to use `go env MOD` instead of "go.mod".

2. That's right.  
However, go/internal/modfile (It can not be used alone. I think there needs modfetch, modload etc) seems a bit complicated. I think that it is best to be porting go/internal, but how about using this package until then?

cmd/goa/gen.go
```diff
@@ -142,16 +142,34 @@ func (g *Generator) Write(debug bool) error {
 	return err
 }
 
+const GOMODENVKEY = "GOMOD"
+
+func findGoMod() (string, error) {
```
raphael: Looks like this function can never return an error which is good so the error return value can be removed.

raphael (commented):
Looks good, thanks

raphael:
Thinking a bit more about parsing `go.mod`: we really don't need a full parser. All we need is to find the line that starts with `goa.design/goa/v3` and capture what's after. Seems a little heuristic here might be better than bringing in all these dependencies. Thank you!

ikawaha:
I replaced the third party package with a heuristic parser.
Does it make sense?

ikawaha:
It may be better to output the version of goa when go get. 🤔

raphael:
The parser looks great! Thank you for making the change. I'm happy to merge the PR as is.

ikawaha:
That's all.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/manveru/gobdd v0.0.0-20131210092515-f1a17fdd710b // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/sergi/go-diff v1.0.0
+	github.com/sirkon/goproxy v1.4.0
```
ikawaha: Oops

raphael:
I wonder if we need to print the version on the console? this may break scripts that wrap goa and capture its output.

raphael:
Thank you!

ikawaha:
What about displaying the version of goa modules by 'go get', when 'goa version' command run?  If it looks good, I create another pull request.

raphael:
Not sure about `goa version` but it would probably be OK to show it when running with the `--debug` flag.

Issue #2148, opened 2019-06-10
V3: How to handle unmarshalling errors?


ilmaruk:
I have a design that looks roughly like this:
```package design

import (
    . "goa.design/goa/v3/dsl"
)

var _ = API("test-api", func() {
    Server("test-server", func() {
        Services("test-service")
    })
})

var _ = Service("test-service", func() {
    Method("put-exams", func() {
        Payload(func() {
            Attribute("studentId", Int, "The student id (numeric).")
            Attribute("exams", ArrayOf(examType), "The list of exams for a specific student.")
            Required("studentId", "exams")
        })

        Error("not-found")

        HTTP(func() {
            PUT("/students/{studentId}/exams")

            Params(func() {
                Param("studentId")
            })

            Response(StatusNoContent)
            Response("not-found", StatusNotFound)
        })
    })
})

var examType = ResultType("exam", func() {
    Attribute("name", String)
    Attribute("mark", Int)
    Attribute("notes", String)
    Required("name", "mark")
})
```
and I'm getting the following error, if the request has a body that doesn't match the expected structure:
```
2019/06/10 14:47:34 http: panic serving 127.0.0.1:50452: runtime error: invalid memory address or nil pointer dereference
goroutine 40 [running]:
net/http.(*conn).serve.func1(0xc0001d41e0)
	/usr/local/go/src/net/http/server.go:1769 +0x139
panic(0x7fe980, 0xbc3210)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
goa/gen/http/test_service/server.unmarshalExamRequestBodyToTestserviceExam(...)
	/tmp/goa/gen/http/test_service/server/encode_decode.go:99
[..]
```
E.g.:
```
$ echo '{"exams":[{}]}' | http PUT localhost:9090/users/1973/exams
```
which happens before the handling function is invoked.
Is there a way of handling such incorrect payloads nicely (e.g.: returning a 400)?

nitinmohan87:
Definitely a bug! The generated validation code for the request body only checks if the `exams` is present in the request body. It must have also validated the contents of each `exam` item.

ilmaruk:
I believe the bug lies in the assumption that unmarshalling would return an error if the provided JSON was not "compatible" with the receiving structure. Unmarshalling instead returns an error only if the JSON blob is malformed or if an object is defined where a list is expected, etc.
[...]

PR #2191, opened 2019-07-10
Add failing test case for validation of array of UserTypes

jamesrom:
An existing issue (#2148) shows unmarshalling errors due to validation not occurring for `ArrayOf(UserType)`.

Looking at the code, it looks like this _should_ be handled in [validation.go@226](https://github.com/goadesign/goa/blob/599df23a2ff78d4353b91a814790f9ab229d43bc/codegen/validation.go#L226) but for a reason unknown to me, `val` is always empty.

This failing test case highlights the issue.

### DIFF
--- a/codegen/validation.go
+++ b/codegen/validation.go
@@ -222,12 +222,12 @@ func recurseValidationCode(att *expr.AttributeExpr, attCtx *AttributeContext, re
 		ctx.Pointer = false
 		val := recurseValidationCode(a.ElemType, ctx, true, "e", context+"[*]", seen).String()
 
-		switch a.ElemType.Type.(type) {
-		case expr.UserType:
-			// For user and result types, call the Validate method
-			val = runUserValT(attCtx.Scope.Name(a.ElemType, ctx.Pkg), "e")
-		}
 		if val != "" {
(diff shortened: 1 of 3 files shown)


codegen/validation.go
```diff
@@ -221,12 +221,13 @@ func recurseValidationCode(att *expr.AttributeExpr, attCtx *AttributeContext, re
 		ctx := attCtx.Dup()
 		ctx.Pointer = false
 		val := recurseValidationCode(a.ElemType, ctx, true, "e", context+"[*]", seen).String()
+
```
raphael, line 224: Hello James, thank you for the PR! this check was supposed to ensure that Goa would only generate a validation method when necessary. Seems that we would need to understand why `val` is the empty string in this case when it shouldn't be rather than removing the test?

jamesrom: Thanks for the quick response Raphael. I do agree with you that we need to understand why `val` is empty.

jamesrom:
I don't have time to work on this right now. For anyone who finds this issue in the future, the failing test case can be found [here](https://github.com/jamesrom/goa/tree/validate-recursive-array-types).

PR #2190, opened 2019-07-09
Move xray test helpers into separate package

tmcladam:
Fixes #2160 ("testing" imported by production code)

PR #2192, opened 2019-07-10
(v2) Move xray test helpers into separate package (#2190)

tmcladam:
* Rename testing.go to testing_test.go so that
"testing" is not imported in release code

* Move testing.go into its own xraytest pkg

* Change tests to refer to separate xraytest pkg

* Added package comment for staticcheck

(cherry picked from commit eeb68ca3a4fa4d7d821f932fe209ce764545bcbe)


raphael:
Thank you!

PR #2194, opened 2019-07-11
Fixed #2180 in v3 by only deleting subdirs within the target gen dir

### LINKED ISSUE #2180: v3: Would be nice if gen didn't delete non .go files
I like to put a README in the gen dir for project documentation and to keep the gen dir around in git. But currently `goa gen` deletes all files in gen, including my README. Could it be changed to only delete *.go files, and least at the gen/ top-level?

cquinn:
cleanupDirs() now returns the subdirs in output/gen whenever it can. This leaves all top-level files as-is in the gen dir to allow for documentation or scripts to live here in source control. Go source is never generated into this top-level dir, so no worries there.

https://github.com/goadesign/goa/issues/2180

### DIFF
--- a/cmd/goa/gen.go
+++ b/cmd/goa/gen.go
@@ -272,8 +272,8 @@ func (g *Generator) runGoCmd(args ...string) error {
 	return nil
 }
 
-// cleanupDirs returns the names of the directories to delete before generating
-// code.
+// cleanupDirs returns the paths of the subdirectories under gendir to delete
+// before generating code.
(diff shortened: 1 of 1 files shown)


raphael (commented):
Same comment as the v2 PR:
Thank you, looks great! Just one tweak that might be good: could we update the header comment of `cleanupDirs` to indicate the new behavior?

raphael:
Thank you!

PR #2196, opened 2019-07-11
Fixed #2193 in v3 for incorrect generated cast for Bytes payloads for client CLI

### LINKED ISSUE #2193: v3: Generated incorrect cast in Bytes Decoder for client
Having a design with a Method Payload containing
  `Attribute("body", Bytes, "New Item Body")`
and HTTP containing
  `Body("body")`
Results in the http <service> client cli code containing an incorrect `string()` cast that should be `[]byte()`

cquinn:
Updated the cast for CLI args for []byte payload from string() to []byte()

https://github.com/goadesign/goa/issues/2193


raphael:
Thank you! Would you mind making a PR for v2 as well?

PR #2209, opened 2019-07-18
Remove security validations in HTTPEndpointExpr since MethodExpr already validates them

nitinmohan87:
* No need to validate security requirements in HTTPEndpointExpr as [MethodExpr already does that](https://github.com/goadesign/goa/blob/v3/expr/method.go#L104-L135)
* Fixes a bug where validation fails if security attribute is extended inside a UserType ([this test case](https://github.com/goadesign/goa/blob/v3/expr/http_endpoint_test.go#L61) reproduces the bug but was returning false positive)


raphael:
Thank you!

PR #2210, opened 2019-07-18
Port to v2: Remove security validations in HTTPEndpointExpr since MethodExpr already validates them

nitinmohan87:
#2209


raphael:
Thank you!

PR #2211, opened 2019-07-18
Use err when decoding error response type

nitinmohan87:
Fixes a bug where `err` variable is not used after initialization when decoding error response types.

### DIFF
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -1664,13 +1664,23 @@ func buildErrorsData(e *expr.HTTPEndpointExpr, sd *ServiceData) []*ErrorGroupDat
 			}
 
 			headers := extractHeaders(v.Response.Headers, v.ErrorExpr.AttributeExpr, svcctx, sd.Scope)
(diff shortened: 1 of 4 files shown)


http/codegen/service_data.go
```diff
@@ -1664,13 +1664,23 @@ func buildErrorsData(e *expr.HTTPEndpointExpr, sd *ServiceData) []*ErrorGroupDat
...
-				ServerBody:  serverBodyData,
-				ClientBody:  clientBodyData,
-				ResultInit:  init,
+				StatusCode:   statusCodeToHTTPConst(v.Response.StatusCode),
+				Headers:      headers,
+				ErrorHeader:  v.Name,
+				ServerBody:   serverBodyData,
+				ClientBody:   clientBodyData,
+				ResultInit:   init,
+				MustValidate: mustValidate,
```
arangamani: Seems like this field is newly introduced in this struct? Where is this being used?

raphael:
Looks great, thank you!

PR #2213, opened 2019-07-19
Escape URLs in JSON Encoder

83naff:
There is an issue with the JSON encoder in which URLs in fields are not escaped. That is, a URL currently will be represented as:

**https://www.foo.com?var1=foo\u0026uvar2=foo\u0026uvar3=baz**

The fix I am proposing will represent URLs in JSON as:

https://www.foo.com?var1=foo&var2=bar&var3=baz

### DIFF
--- a/http/encoding.go
+++ b/http/encoding.go
@@ -71,7 +71,7 @@ func RequestDecoder(r *http.Request) Decoder {
 	}
 	switch contentType {
 	case "application/json":
-		return json.NewDecoder(r.Body)
+		json.NewDecoder(r.Body)
 	case "application/gob":
 		return gob.NewDecoder(r.Body)
 	case "application/xml":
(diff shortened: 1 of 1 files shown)


raphael:
Thank you for the PR! I don't think that whether or not HTML should be escaped in the JSON encoding is something that Goa can decide for the users. Goa opted to follow the same default as the Go standard library JSON encoding. If that default does not work in your case then provide your own encoder/decoder that is configured correctly for your needs.

83naff:
@raphael Hi Raphael, please point me in the direction as to how one provides their own custom encoders/decoders as the guide [here](https://goa.design/implement/encoding/) is out of date and the documentation [here](https://godoc.org/goa.design/goa/dsl#Produces) does not match the guide given. :-)

raphael:
@damien-lloyd here is where the default example initialize the encoder/decoder:
https://github.com/goadesign/examples/blob/master/basic/cmd/calc/http.go#L35-L38
This is using the Goa default `RequestDecoder` and `ResponseEncoder`. These can be replaced with your own and are used by the functions that instantiate the HTTP server:
https://github.com/goadesign/examples/blob/master/basic/gen/http/calc/server/server.go#L47-L48
The function signatures are defined in the Goa `http` package:
https://github.com/goadesign/goa/blob/v3/http/encoding.go
Hope this helps! And yes it would be good to have more docs around that, could make a great first PR! ;)

PR #2214, opened 2019-07-19
Return default goa error honoring Accept request header for non-existent endpoints

nitinmohan87:
Fixes #2174

PR #2215, opened 2019-07-19
Port to v2: Return default goa error honoring Accept request header for non-existent endpoints

nitinmohan87:
#2214


raphael:
Thank you!

PR #2217, opened 2019-07-22
v3: Fix grpc bidirectional streaming client close

### LINKED ISSUE #2216: v3: grpc bidirectional streaming client Close() doesn't cause EOF on server
GRPC bidirectional streaming client.Close() should call CloseSend() on the internal client stream in order that the server side RecvMsg() will get io.EOF.

PR incoming

tmcladam:
- Make client.Close() call CloseSend() on the internal client stream
- Don't implement Close() in the server streaming client because Close() isn't in the corresponding interface

Fixes #2216

### DIFF
--- a/grpc/codegen/client.go
+++ b/grpc/codegen/client.go
@@ -90,7 +90,7 @@ func client(genpkg string, svc *expr.GRPCServiceExpr) *codegen.File {
 						Data:   e.ClientStream,
 					})
 				}
-				if e.ServerStream.MustClose {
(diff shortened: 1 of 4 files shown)


raphael:
Thank you Tim! @nitinmohan87 can you give Tim's PR a once over? thank you!

nitinmohan87 (approved):
Looks great! Could you also please port this to v2?

tmcladam:
See #2220 for v2 port

PR #2218, opened 2019-07-22
Codegen fixes v3

nitinmohan87:
Fix various bugs in codegen


raphael:
Thank you! Can you please backport these to v2 as well? thanks!

PR #2220, opened 2019-07-22
v2: Fix grpc bidirectional streaming client close

tmcladam:
Backport of #2217 to v2

- Make client.Close() call CloseSend() on the internal client stream
- Don't implement Close() in the server streaming client because Close() isn't in the corresponding interface


raphael:
Thank you!

PR #2223, opened 2019-07-24
v3: Call validate function in grpc streaming RecvMsg

tmcladam:
Fixes #2222

PR #2224, opened 2019-07-25
v2: Call validate function in grpc streaming RecvMsg

tmcladam:
* Call validation code on received stream messages

Port of #2223 to v2


raphael:
Thank you!

PR #2246, opened 2019-08-16
Fixes streaming endpoints with no payload

### LINKED ISSUE #2242: v3: Streaming method without payload does generate incorrect server code
Simple design:
```
package design

var _ = API("api", func() {
	Server("apisvr", func() {
		Host("default", func() {
			URI("http://0.0.0.0:8000")
			URI("grpc://0.0.0.0:8080")
		})
	})
})

var _ = Service("ws", func() {
	HTTP(func() {
		Path("/")
	})

	Method("Retrieve", func() {
		StreamingResult(String)

		HTTP(func() {
			GET("/")
			Response(StatusOK)
		})
	})
})
```
The method `Retrieve` has no `Payload` defined which leads to the following error on running:
```
gen/http/ws/server/server.go:155:6: undefined: err
gen/http/ws/server/server.go:157:6: undefined: err
gen/http/ws/server/server.go:158:16: undefined: err
gen/http/ws/server/server.go:161:34: undefined: err
```
The line defining the `err` variable comes from `payload, err := decodeRequest(r)` which is missing without the payload and `err` is never declared.

cubic3d:
Fixes #2242. This declares the error variable outside of the payload generating section to avoid missing declaration on streaming endpoints without payload.


raphael:
Thank you for opening the issue and for the PR! Here is an [example](https://github.com/goadesign/goa/commit/918b85c89cee6ed405f68eb71df145773aa7e579) on how to fix the build (see how the import statements should be formatted). This must be a recent change with `golint`, updating your copy should help reproduce the issue.

http/codegen/server.go
```diff
@@ -454,8 +454,10 @@ func {{ .HandlerInit }}(
 		ctx = context.WithValue(ctx, goa.MethodKey, {{ printf "%q" .Method.Name }})
 		ctx = context.WithValue(ctx, goa.ServiceKey, {{ printf "%q" .ServiceName }})
 
+		var err error
```
raphael: Nitpick but I'm wondering if changing the code so that the `err` variable only gets defined that way in the case that there is no payload wouldn't be better so the change would be:
```go
{{- if .Payload.Ref }}
		payload, err := decodeRequest(r)
		...
- {{- end }}
+ {{- else }}
+    var err error
+ {{- end }}
```

cubic3d: Thank you for the feedback! In my opinion, trading a the simple declaration `var err error` for a if statement makes it less readable. But I'm fine with both of course, tell me which one you'd like to see in the PR ;)

raphael: You're right that simply declaring the error variable makes the Goa code easier to read but it also ends up generating Go code that isn't idiomatic:
```go
var err error
payload, err = decodeRequest(r)
```
While certainly not perfect one of the design goals of the Goa code generators has been to generate idiomatic code so I would tend to favor adding the `if`. Thank you!

cubic3d: I'll do ;)

cubic3d:
Should be all fine now

raphael:
Great! thank you, could you backport this to v2 as well please?

cubic3d:
@raphael Sure, but looks like v2 does not pass builds at the moment

PR #2247, opened 2019-08-16
Add support for Int, UInt, UInt32, UInt64 as map key types in OpenAPI

nitinmohan87:
Fixes bug in OpenAPI generation for design below
```
var _ = Service("foo", func() {
  Method("bar", func() {
    Payload(func() {
      Attribute("error", MapOf(Int, String))
    })
    HTTP(func() {
      GET("/")
    })
  })
})
```
Error:
```
template: openapi:1:3: executing "openapi" at <toJSON .>: error calling toJSON: openapi: json: unsupported type: map[interface {}]string
```


raphael:
@nitinmohan87 I fixed an issue with some import statement not being formatted properly for the Go 1.12 linter. There seems to be some genuine test failures now.

raphael:
Thank you!

PR #2248, opened 2019-08-16
Port to v2: Add support for Int, UInt, UInt32, UInt64 as map key types in OpenAPI

nitinmohan87:
https://github.com/goadesign/goa/pull/2247


raphael:
Thank you!

PR #2272, opened 2019-09-04
Primitive user types

nitinmohan87:
This PR adds the ability to define user types that alias primitive types. This makes it possible to define a description, validations and examples and have these apply to all attributes that use the type. For example:

```go
var IDType = Type("ID", String, func() {
    Description("ID describes an identifier")
    Example("8edee126-a15c-427d-b1c6-9d58143a112c")
    Format(FormatUUID)
})

var SomeType = Type("SomeType", func() {
    Attribute("id", IDType)
    //....
})
```


MichaelUrman:
I want this for using defined types with gRPC, but it doesn't do the trick as of 8fdffd8. (The .proto is invalid once you have a field with type like `Type("MyInt", Int)`.)

I've extended the tests and hacked together a [functional implementation](https://github.com/goadesign/goa/compare/primitive_user_types...MichaelUrman:primitive_user_types); apologies if the names are using incorrect nomenclature. I'm looking either for feedback on the "right" way to do the ugly bits, or to be a drive-by code donation and let @nitinmohan87 consume and finish this.

raphael:
Thank you for the fix @MichaelUrman! Looks good to me. @nitinmohan87 could you also give it a look? thanks!

nitinmohan87:
@MichaelUrman Your implementation looks good to me. I think we can get that merged in.

codegen/funcs.go
```diff
@@ -2,9 +2,33 @@ package codegen
...
+	"fmt"
 	"os"
 	"strings"
 	"unicode"
+
+	"goa.design/goa/v3/expr"
+)
+
+type (
+	// InitArgData contains the data needed to render code to initlize struct
```
raphael: typo 'initialize'

codegen/validation.go
```diff
@@ -325,8 +349,8 @@ func recurseAttribute(att *expr.AttributeExpr, attCtx *AttributeContext, nat *ex
 		if hasValidations {
 			var buf bytes.Buffer
 			tgt := fmt.Sprintf("%s.%s", target, attCtx.Scope.Field(nat.Attribute, nat.Name, true))
-			if expr.IsArray(nat.Attribute.Type) {
```
raphael, line 328: Do we  need to keep the `expr.IsArray` test?

codegen/funcs.go
raphael: typo: initialize

raphael (commented):
This is really great!! I spotted a typo and left a question on one change I don't understand, Will merge once these two items are resolved.

codegen/validation.go
nitinmohan87: No.. Before, we generated the validation code for a result type collection inline instead of creating a new `ValidateXXX` function with the validation code. By this change, we will be generating validation code for aliased primitive types inline and create a new `ValidateXXX` function for result type collection and other user types (aliased or unaliased). I felt it made managing this code a little easier.

nitinmohan87: Here is the effect of this change - https://github.com/goadesign/goa/blob/4f9dd318c61be04cd8dfb3fbbdd4e46c1b34e78a/codegen/testdata/validation_code.go#L383-L387
Notice how it generates this function `ValidateResultCollection` for validating result type collection
```go
        if err2 := ValidateResultCollection(target.Collection); err2 != nil {
		err = goa.MergeErrors(err, err2)
	}
```
Without this change, it would have generated a `for` loop and ran the validation for each item in the collection. Now, the iteration is done inside the generated function. I feel this is more consistent because now we know a new `ValidateXXX` function will be generated for all user types except for aliased primitive types.

raphael: OK makes sense, the initial intent was to avoid generating extra functions but you make a good point that it is more consistent this way.

Issue #2271, opened 2019-09-03
swagger:extension is ignored in API level


townewgokgok:
I specified a swagger extension according to [ReDoc vendor extensions](https://github.com/Redocly/redoc/blob/0eb1e66a10ad3add8db147e2bd6ce6ee333edd68/docs/redoc-vendor-extensions.md) like

```
var _ = API("...", func () {
    // ...
    Meta("swagger:extension:x-tagGroups", `[{"name":...}]`)
})
```
but it does not appear in `openapi.yaml`.

Shouldn't this `root.Meta` to be `root.API.Meta` ? 🤔 

https://github.com/goadesign/goa/blob/0e5c54ed0fd48ebc00de62a09090b542ee6ea27a/http/codegen/openapi/openapi_v2_builder.go#L51

townewgokgok:
Oops, it seems unlikely that I can do what I want to do because the metadata output destination is in `info`, not the root.
~The problem is that the ReDoc vendor extension does not meet the OpenAPI specification (however, this is often the case and I am often in trouble).~
~I will try to deal with workarounds.~

It seems that vendor extensions are also allowed in the root.
https://github.com/OAI/OpenAPI-Specification/blob/597a638b50fe9848fff449fcad67290a839229be/schemas/v2.0/schema.json#L12-L16
Are you planning to add `Extensions` field to the type ` V2` ?

townewgokgok:
It seems that @nitinmohan87 has changed this time.
https://github.com/goadesign/goa/commit/6d9200ca48bab5261f3929cef8f114cf7d55a7ec#diff-6ecbb03fe70b40b52aa6080b1153f04aL53

townewgokgok:
Same here
https://github.com/goadesign/goa/commit/6d9200ca48bab5261f3929cef8f114cf7d55a7ec#diff-6ecbb03fe70b40b52aa6080b1153f04aL23

townewgokgok:
Aside from the root level vendor extension, please check pull request #2273 to fix the original problem of this issue.
I'll create another issue about root level vendor extension later.

townewgokgok:
This issue has been resolved by #2298.

PR #2273, opened 2019-09-05
Fix collecting OpenAPI vendor extensions and tags in root level (#2271)

townewgokgok:
The API level vendor extension was ignored, so when I blame the code, it seemed that this change was the cause.

https://github.com/goadesign/goa/commit/6d9200ca48bab5261f3929cef8f114cf7d55a7ec#diff-6ecbb03fe70b40b52aa6080b1153f04aL53

Similarly for `swagger:tags`

https://github.com/goadesign/goa/commit/6d9200ca48bab5261f3929cef8f114cf7d55a7ec#diff-6ecbb03fe70b40b52aa6080b1153f04aL23


raphael:
Thank you this is great! do you think you could add a test that would fail without this change?

townewgokgok:
OK, I will try it.

townewgokgok:
Please check the results of #2277 and close it, then review this PR again.

townewgokgok:
Please check the results of #2277 and close it, then review this PR again.

After fixing `swagger:tag:xxx` problem, another problem was found.
Details:
https://github.com/goadesign/goa/pull/2273/files#diff-6ecbb03fe70b40b52aa6080b1153f04aR328-R331
https://github.com/goadesign/goa/pull/2273/files#diff-a9ff15d94de98ce0d197b7bd57e370a8R34

Is the behavior that API level extensions are merged into all tags expected?

raphael:
Thank you for all the work @townewgokgok, this is great. I don't think merging the API level extensions is the right thing to do. Also I believe we should remove the field `Meta` from the `RootExpr` struct (https://github.com/goadesign/goa/blob/40417a02bce8e16ea2c489e5010fa49e131c4e1c/expr/root.go#L35-L37) and remove the line that sets it in the `Meta` function in the `dsl` package (https://github.com/goadesign/goa/blob/40417a02bce8e16ea2c489e5010fa49e131c4e1c/dsl/meta.go#L153-L154) since that's not being used anywhere. Thank you again!

townewgokgok:
Thank you for reviewing.
Do you think it's better to delete these lines together? (The line 331 already exists before sending this PR, and what I did is only adding comments.)
This PR (resolving the problem about API level extensions) will make this behavior reproduced.
https://github.com/goadesign/goa/blob/40417a02bce8e16ea2c489e5010fa49e131c4e1c/http/codegen/openapi/openapi_v2_builder.go#L328-L331

raphael:
@tchssk do you remember why we merge the API level extensions to all other tags?
https://github.com/goadesign/goa/pull/2273#issuecomment-529297937

tchssk:
Sorry, I don't know about it. I was involved swagger extensions in v1 and not v2.

raphael:
@townewgokgok it seems to me that we should be OK removing the merge, thank you!

townewgokgok:
Please review again and merge if there is no problem.

raphael:
Looks great, thank you! Do you think you could back-port these changes to v2? (simply cherry-pick the merge commit into a branch created off of v2 and submit a PR) Thanks!

PR #2279, opened 2019-09-10
Fix go1 13 compatibility v2

### LINKED ISSUE #2274: `goa gen` and `goa example` uses relative paths instead of packages with Go 1.13
When running `goa gen` or `goa example` on a project while using Go 1.13, the generated files reference servers/clients by relative path, instead of package.

Using the example from the getting started guide, the import of the `calc` service in:
`gen/http/calc/server/server.go` is as follows:

`calc "../../../../path/to/project/gen/calc"`

With no other changes, switching back to Go 1.12 fixes the issue.

raphael:
Make Goa compatible with Go 1.13

Main issue was that the behavior of 'packages.Load' when loading a non-
existent package changed: the field `PkgPath` of the return package
object is not set correctly anymore.

Also updated the validation tests to handle changes in how the 'mail'
package formats error messages.

Fix #2274

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 dist: xenial
 go:
-- 1.12.x
+- 1.13.x
 cache:
   directories:
     - $HOME/.cache/go-build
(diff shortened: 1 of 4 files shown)


raphael:
Circle CI image still uses Go 1.12 hence the broken build there.

PR #2280, opened 2019-09-10
Re-enable building examples during CI

raphael:
Now that the 'goa' tool on the v3 branch supports Go 1.13


raphael:
I'm closing this in favor of a release make target that will take care of updating both the examples and plugins when releasing instead of doing it for each single commit which renders the build very slow.

PR #2283, opened 2019-09-17
Properly inherit params and payload attributes to child services

tchssk:
When `Parent` DSL is used, headers and params are inherited to the child service but the payload definition is not. This patch fixes it.


raphael:
Thank you for the PR! I am not sure that we always want to merge the parent and child method payloads. The parent payload could have many other attributes that the child method does not use. I think the idea here is to redefine the `parent_id` attribute in the child payload explicitly instead of Goa doing the merge automatically. For example the design type for the child method can use `Extend` if indeed merging is what the user wants to do. It's probably better to let the user define the behavior in the design instead of forcing the merge. Does that make sense to you?

tchssk:
It makes sense. Doing the merge all attributes automatically was too much. However, it seems better to merge only the path parameters. They are REQUIRED in the child service methods.

tchssk:
Hmm
In the current implementation, both path and query parameters are inherited from the parent service so users must define payload attributes for all parameters including query into child services.

tchssk:
* 146e44b failed because query parameters are inherited but the payload attributes aren't. Query parameters don't have to be inherited because not every service need same params as parent.
* f974ae9 fixed to inherit only path params to child services.
* d00be5a changed to inherit attributes for path params to child services.

tchssk:
I finished to fix the PR. Please take a look.

raphael:
Thank you the PR looks great!. It might be good to check whether the child payload already includes an attribute with the name of the query param attribute and not override if that's the case (there could be different validations for example). Also it might be good to explain that behavior in the comment of the `Parent` DSL function. Something like:
```go
// Attributes of the parent method payload that map to parent path parameters
// are automatically merged into the child method payload type if not already
// defined.
```
Thank you!

tchssk:
Thank you for your review. I fixed it.

raphael:
Great! Can you please port to v3 as well? Thank you!

PR #2292, opened 2019-09-25
Remove empty validate methods in gRPC

nitinmohan87:
`field` is internal to goa and should never be validated.


raphael:
Great, thank you! Is there no code being generated that assume that field is not nil?

nitinmohan87:
@raphael Do you mean in the examples?

raphael:
I guess wherever we generate code that accesses the field. Before the validation would have returned a response right away but now it won't. So for example the code that transforms the transport struct back into a service level struct.

PR #2293, opened 2019-09-26
trace_discards_v3

smessier:
added DiscardFromTrace() option for tracing in HTTP and gRPC.
useful to omit API calls such as health-checks, which happen frequently but are not interesting to trace.

### DIFF
--- a/grpc/middleware/trace.go
+++ b/grpc/middleware/trace.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"context"
+	"regexp"
 
 	"goa.design/goa/v3/middleware"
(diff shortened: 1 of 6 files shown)


raphael:
Looks great, thank you. Just a tiny change that would be good: the change is in the generic tracing middleware - not specific to X-Ray so it would be better if the comments didn't mention X-Ray:
```go
// DiscardFromTrace adds a regular expression for matching a request path to be discarded from tracing
```
instead of
```go
// DiscardFromTrace adds a regular expression for matching a request path to be discarded from X-Ray tracing
```
Thank you!

raphael:
Thanks!

PR #2295, opened 2019-09-26
trace_discards_v2

smessier:
added DiscardFromTrace() option for tracing in HTTP and gRPC.
useful to omit API calls such as health-checks, which happen frequently but are not interesting to trace.


raphael:
Thanks!

PR #2299, opened 2019-10-01
OpenAPI operation level extensions

townewgokgok:
This PR makes extensions in `Method` level (instead of `Route` level) merged to `openapi.Operation` .

The reason for this is:

- `Method` level extensions were not merged to anywhere before this change.
  This can be confirmed by passing the 139eda9e33571c8019b6acf0c5f7f21641dcf1ce test.
- `Meta` DSL cannot be used in `GET`, `POST`, etc. so extensions cannot be added into `expr.Route`.
- OpenAPI `Operation` level extensions are [valid](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-objects-4).

Example:

- [Design](https://github.com/goadesign/goa/blob/5fb43a72fa8309ec9b50bcf1a97846b89d21fda5/http/codegen/testdata/openapi_v2_dsls.go#L283-L291)
  
  ```go
  Method("testEndpoint", func() {
      Payload(Empty)
      Result(Empty)
      HTTP(func() {
          POST("/")
          Meta("swagger:extension:x-test-foo", "bar")   // not this (Path Item Object level extension)
      })
      Meta("swagger:extension:x-test-operation", "Operation")   // <- THIS
  })
  ```
- [Generated openapi.yaml](https://github.com/goadesign/goa/blob/5fb43a72fa8309ec9b50bcf1a97846b89d21fda5/http/codegen/testdata/openapi_v2/TestExtensions/endpoint_file1.golden#L15-L28)
  
  ```yaml
  paths:
    /:
      post:
        operationId: testService#testEndpoint
        responses:
          "204":
            description: No Content response.
        schemes:
        - https
        summary: testEndpoint testService
        tags:
        - testService
        x-test-operation: Operation   # <- THIS
      x-test-foo: bar   # not this (Path Item Object level extension)
  ```


townewgokgok:
Related PRs: #2299(this) #2300 #2301 
These conflict, so if you merge them one by one, I will rebase each time.

raphael:
This is great, thank you for the explanation and for the tests. I'll merge these PRs one by one starting with this one. Once all three are merged can you also back-port this to v2? (a single PR combining these 3 is fine for v2). Thank you!

townewgokgok:
Thank you for reviewing.
#2300 and #2301 were rebased onto `v3` merged this PR.

PR #2302, opened 2019-10-02
v1: Use http.NewRequestWithContext in goagen

furusax0621:
Some methods generated by `goagen` do not use `ctx` defined as arguments, *http.Request always uses context.Background(). 

To include `ctx` in *http.Request, use http.NewRequestWithContext() added in Go 1.13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.11.x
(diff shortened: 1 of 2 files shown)


raphael:
Great, thank you!

shogo82148:
👍 
Could you release new version, please?

PR #2304, opened 2019-10-02
V3 Refactoring: Remove go-get executed before go-build in goa gen command

ikawaha:
Previously I fixed issue #2181 with #2182.
However, this fix was redundant, and it was not necessary to go get the goa package.
This is because `go mod init` has been performed in the project folder and `go get` goa.

Even if there is no goa package when `goa gen/example`, `go build` should automatically download goa because it is described in go.mod (although there is no such case).

Is it make sense?


raphael:
Yes that makes sense! big simplification and it's also going to make the tool faster. Thank you!

PR #2306, opened 2019-10-02
Fix invalid paths in generated OpenAPI definitions

tchssk:
`goa gen` generates invalid paths in OpenAPI definitions. 

## Before

```yaml
paths:
  /assets/{*filepath}}:
           ^         ^
```

## After

```yaml
paths:
  /assets/{filepath}:
```


tchssk:
86298c1 should fail.

raphael:
This is great, thank you! Can you please port to v3 as well? thanks!

PR #2310, opened 2019-10-07
Support text/html and text/plain by RequestDecoder of http

tchssk:


raphael:
Great, thank you! Would you mind porting this to v3 as well? thanks!

PR #2313, opened 2019-10-09
Fix incorrect fields of operations in generated OpenAPI definitions

tchssk:
For a multipart operation, `goa gen` should generate `multipart/form-data` as `consumes` and `in: formData` to the operation parameters.

## Before

```yaml
paths:
  /foo:
    parameters:
    - in: body
```

## After

```yaml
paths:
  /foo:
    consumes:
    - multipart/form-data
    parameters:
    - in: formData
```

### DIFF
--- a/http/codegen/openapi/openapi_v2_builder.go
+++ b/http/codegen/openapi/openapi_v2_builder.go
@@ -625,7 +625,7 @@ func buildPathFromExpr(s *V2, root *expr.RootExpr, h *expr.HostExpr, route *expr
 			responses[strconv.Itoa(er.Response.StatusCode)] = resp
 		}
 
-		consumes := []string{}
+		var consumes []string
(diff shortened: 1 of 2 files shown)


tchssk:
ca109d7  will fail.

http/codegen/openapi/openapi_v2_builder.go
```diff
@@ -625,10 +625,19 @@ func buildPathFromExpr(s *V2, root *expr.RootExpr, h *expr.HostExpr, route *expr
 			responses[strconv.Itoa(er.Response.StatusCode)] = resp
 		}
 
+		consumes := []string{}
```
raphael: Could that be `var consumes []string` instead?

raphael:
Looks great, just left one question - thank you!

raphael:
Thank you! can you please port to v3?

PR #2316, opened 2019-10-11
Add example generation for JSON format

tchssk:


raphael:
Thank you! I wonder if the code couldn't just always use the hard-coded string instead of trying to use the JSON marshaler first?

raphael:
Thank you! Can you port that to v3 as well please? thanks!

PR #2317, opened 2019-10-12
fix Decode{{Name}}Request method

dvictor:
For client streaming with payload, Goa uses metadata to also send a unary payload.
This change fixes code generation for non-string payload fields.


raphael:
Thank you! Could you please back-port this to v2 as well? (just cherry-pick the merge commit into a branch created off-of v2)

PR #2318, opened 2019-10-14
Fix dsl.Description doc comment

sevein:
[Description](https://godoc.org/goa.design/goa/dsl#Description) refers to `FileServer` as one of the places where it can be used.
It should read `Files` instead.


raphael:
Thank you!

PR #2323, opened 2019-10-17
Fix incorrect fields of operations in generated OpenAPI definitions

tchssk:
This is backport of #2313 to v3.


raphael:
Thank you!

PR #2327, opened 2019-10-18
Fix package name of generated dummy multipart file

tchssk:
Generated dummy multipart file should have same package name as example service files.


raphael:
This is great thank you! can you please port to `v3`? thanks!

PR #2328, opened 2019-10-19
Fix package name of generated dummy multipart file

tchssk:
#2327 was based onto `fix_files_dsl_doc_v2` branch by mistake so I deleted the branch.


raphael:
OK thank you.

Issue #2303, opened 2019-10-02
Field uses the grpc tag of the Reference field instead of its own


tmcladam:
When a Field is a Reference to a Field in another type, the attribute ends up with two "rpc:tag" values, the first belonging to the reference field, and the second belonging to itself. The first Meta("rpc:tag") value seen is the one used for code generation (and validation).

Obvious approaches are:

1. Have Field set its tag before the reference type's
2. Use the last "rpc:tag" value instead of the first
3. Put a special case in Meta to force a single value only for "rpc:tag", with overwrite rather than append behavior.

Thoughts? Are there any other Meta tags for which it doesn't make sense to have multiple values, and that maybe we should apply 3. to, to avoid similar issues with references?

Or, is the intention that field references don't change the tag?

raphael:
Thank you for raising this issue. It seems that references should also bring in metadata as that is consistent with the intent. In all the options listed option #2 might make the most sense: the metadata associated with the name "rpc:tag" is "special" and its handling requires understanding the semantic (i.e. picking the last value). Happy to hear other point of views though.

tmcladam:
Fixed using approach 2 above.

PR #2330, opened 2019-10-21
Fix grpc field tag inheritance and validation

tmcladam:
- Correctly merge grpc endpoint and response validation errors (bug found in testing)
- Allow rpc:tag from reference field to be overridden (#2303)

### DIFF
--- a/codegen/service/testdata/service_dsls.go
+++ b/codegen/service/testdata/service_dsls.go
@@ -529,12 +529,12 @@ var NamesWithSpacesDSL = func() {
 		})
 	})
 	var APayload = Type("Payload With Space", func() {
-		Attribute("String", String)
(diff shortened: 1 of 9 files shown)


raphael (commented):
Looks great, thank you! Just one nitpick: I'm not a big fan of the Csharpy `Get` prefix for `GetFieldTag()` and `GetLast()`, could we name these `FieldTag()` and `Last()` for example instead?

raphael:
It so happens that https://github.com/goadesign/goa/pull/2336 already introduced `Last` so just rename `GetFieldTag` to `FieldTag` and make it use `Last` instead of `GetLast`, thank you!

tmcladam:
Done - that was a nice coincidence!

PR #2331, opened 2019-10-21
Make it possible to create custom encoder/decoder of html

tchssk:
This is a feature proposal. I want to hear your opinion.

`http.RequestDecoder` and `http.ResponseEncoder` can be used as before.

```go
var (
	dec = goahttp.RequestDecoder
	enc = goahttp.ResponseEncoder
)
```

A custom decoder can be created by `NewRequestDecoder()`.

```go
var (
	dec = NewRequestDecoder([]DecoderConstructorSet{
		{
			ContentType: "application/x-msgpack",
			Suffix:      "+msgpack",
			DecoderConstructor: func(r io.Reader) Decoder {
				return msgpack.NewDecoder(r)
			},
		},
	})
)
```

Append decoders can be used with `DefaultDecoderConstructorSets`.

```go
var (
	dec = NewRequestDecoder(append(DefaultDecoderConstructorSets, {
		ContentType: "application/x-msgpack",
		Suffix:      "+msgpack",
		DecoderConstructor: func(r io.Reader) Decoder {
			return msgpack.NewDecoder(r)
		},
	}))
)
```

### DIFF
--- a/http/encoding.go
+++ b/http/encoding.go
@@ -54,6 +54,7 @@ type (
 //     * application/json using package encoding/json
 //     * application/xml using package encoding/xml
 //     * application/gob using package encoding/gob
+//     * text/html and text/plain for strings
 //
 // RequestDecoder defaults to the JSON decoder if the request "Content-Type"
 // header does not match any of the supported mime type or is missing
--- a/http/encoding.go
+++ b/http/encoding.go
@@ -46,7 +46,127 @@ type (
 
 	// private type used to define context keys.
 	contextKey int
+
+	// DecoderConstructor returns a new decoder.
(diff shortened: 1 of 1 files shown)


raphael:
Thank you for the proposal! I understand the intent here but the idea was that the built-in decoders are just a simple default that can easily be replaced by applications as decoders are injected in the generated code. So the idea was to avoid building a complex system that would allow any encoder/decoder to be used but instead allow users to inject their own implementations. Maybe this is better as part of the [encoding](https://github.com/goadesign/examples/tree/master/encodings) example so that people that do want to use a generic solution can see how to do it?

raphael:
Closing this PR, feel free to create a PR in the examples repo.

PR #2332, opened 2019-10-22
Validate a method when a security but no payload

tchssk:


tchssk:
df4217ddd149eb7bbe346aceafeed7210be87c2b will fail.

raphael:
This is great, thank you! Can you please port to v3 as well? thanks!

PR #2333, opened 2019-10-22
Handle Benchmark testing 

dtmkeng:
Handle Benchmark testing


raphael:
Hello, thank you for the work! What is the intent of this PR? Why should this be part of the Goa repo? Seems like something whose intent is to compare performance across multiple framework so probably something that should be kept separately?

dtmkeng:
I make senior project about Explore  Framework with Benchmark  testing  I choose Goa in 1 of 10 Framework ,but went I use Goa with benchmark testing in package testing. In testing Model I want function is return `http.Handler`  
``` go
func loadEcho(routes []route) http.Handler {


	e := echo.New()
       ...
	return e
}
```
but I don't know to call in `Goa package`  then I will  write testing in main repo. it will work. then I PR for review step this. I don't know PR it benefit for this repo.  So if you  kept separately or reject. 
Depending on you.
thank you

raphael:
I'm going to close this PR, feel free to create a new one in the https://github/goadesign/examples repo that adds the test benchmarks as an example.

PR #2336, opened 2019-10-24
Issue 2206 honour swagger example false at attribute level

ilmaruk:
Here I am proposing a fix to the fact that `swagger:example = false` is not honoured at attribute level, as reported at issue 2206.

Tests are passing alright, but I didn't manage to run the new code against any design, so I cannot tell whether I'm been missing anything, or worse.

If alright and approved, I'm more than happy to port the same changes to v2.

### DIFF
--- a/expr/root.go
+++ b/expr/root.go
@@ -278,7 +278,7 @@ func (m MetaExpr) Merge(src MetaExpr) {
 	}
 }
 
-// Return the last value for a specific key, if the key exists and
+// Last returns the last value for a specific key, if the key exists and
 // has values; otherwise return nil, with the "ok" flag set to false.
 func (m MetaExpr) Last(key string) (string, bool) {
 	v, ok := m[key]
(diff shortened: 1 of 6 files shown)


expr/root.go
```diff
@@ -277,3 +277,19 @@ func (m MetaExpr) Merge(src MetaExpr) {
 		}
 	}
 }
+
+// Last returns the last value for a specific key, if the key exists and
+// has values; otherwise return nil, with the "ok" flag set to false.
```
raphael: nitpick: "otherwise returns the empty string, ..."

raphael (commented):
This is a great PR, thank you for it! I left a little nitpick - will merge once addressed. And yes please a backport to `v2` would be great.

expr/root.go
ilmaruk: Well spotted; changed.

raphael:
Perfect, thank you!

PR #2337, opened 2019-10-24
Issue 2206 honour swagger example false at attribute level - V2

ilmaruk:
Same changes as per https://github.com/goadesign/goa/pull/2336, applied to v2.


raphael:
Thank you!

PR #2341, opened 2019-10-28
close dummy gen file and handle remove error

### LINKED ISSUE #2340: gen/temp.*.go not cleaned up sometimes on Windows
I recently updated to v3.0.6 on some projects and started seeing this file under gen.

It was introduced with https://github.com/goadesign/goa/pull/2281

It gets left behind when I run both a gen and example in a single command like:

`goa example goa.design/examples/basic/design && goa gen goa.design/examples/basic/design`

If I run the two command independently, it doesn't happen. I can sometimes get it if I call these two from make targets without the && where the commands end up executing back to back.

It happens on a windows development box. Haven't seen the behavior on any of my linux build nodes. I suspect some kind of os level file locking issue and some kind of race condition between the two invocations using the same filename. Seems far fetched but I can't think of anything else.

I see similar with the goa* (goa34252433) directory in the pwd sometimes on windows but not sure if it is related and those do not break my builds.

There is no feedback if the `defer os.Remove(dummy.Name())` fails. Perhaps changing that to something like:

```
defer func() { if err := os.Remove(dummy.Name()); if err != nil {
// panic ? report the error on os.Stderr?
}
```

would at least give me some feedback that it happened and why.

I've been able to ignore the ./goa* issue because it doesn't affect my build, but the gen/temp.*.go files cause my builds to fail when I lint with goimports

It isn't an option for me to ditch my windows development environment ...insert sad face emoji... so another solution for others stuck in my predicament might be to keep the gen file around maybe named consistently something like doc.go with a package level comment and formatted so it passes goimports linting checks.

Right now, I'm working around this in my makefile with:

`if stat -t  gen/temp.*.go >/dev/null 2>&1; then rm gen/temp.*.go; fi`

after I run gen and example

eric-isakson:
fixes #2340

### DIFF
--- a/codegen/generator/generate.go
+++ b/codegen/generator/generate.go
@@ -12,7 +12,7 @@ import (
 )
 
 // Generate runs the code generation algorithms.
-func Generate(dir, cmd string) ([]string, error) {
+func Generate(dir, cmd string) (outputs []string, err1 error) {
 	// 1. Compute design roots.
 	var roots []eval.Root
 	{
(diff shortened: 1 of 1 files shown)


raphael:
Thank you!

Issue #2342, opened 2019-10-29
V3: Problems occur when using the Parent DSL


ikawaha:
# goa version 

v3 latest commit: cc3f94b310ab29015c3dc214fb8f1d8fee9731f0

# Following design without problems with Parent DSL.

```go
package design

import (
   . "goa.design/goa/v3/dsl"
)

var _ = Service("ParentService", func() {
   Method("show", func() {
      Payload(func(){
         Attribute("parentParam",String)
      })
      Result(Int)
      HTTP(func() {
         GET("/hoge/{parentParam}")
         Response(StatusOK)
      })
   })
})
    
var _ = Service("ChildService", func() {
   HTTP(func() {
      Parent("ParentService")
      Path("baa")
   })
   Method("child-method", func() {
      Result(Int)
      HTTP(func() {
         GET("/fuga")
         Response(StatusOK)
      })
   })
})
```

# Designs that cause problems

## Case 1

When service that does not refer to Parent is added, a problem occurs in the generated http client code.

```go

package design

import (
   . "goa.design/goa/v3/dsl"
)

var _ = Service("ParentService", func() {
   Method("show", func() {
      Payload(func(){
         Attribute("parentParam",String)
      })
      Result(Int)
      HTTP(func() {
         GET("/hoge/{parentParam}")
         Response(StatusOK)
      })
   })
})

var _ = Service("ChildService", func() {

   HTTP(func() {
      Parent("ParentService")
   })

   Method("child-method", func() {
      Result(Int)
      HTTP(func() {
         GET("/fuga")
         Response(StatusOK)
      })
   })
})

// ★ This service added.
var _ = Service("NotRelativeService", func() {
   Method("some-method", func() {
      Result(Int)
      HTTP(func() {
         GET("/zzz")
         Response(StatusOK)
      })
   })
})
```

### Error

```shellsession
exit status 1
/Users/ikawaha/go/src/github.com/ikawaha/calc/gen/http/non_relative_service/client/encode_decode.go:46:15: expected type, found ')'
...snip...
// EncodeSomeMethodRequest returns an encoder for requests sent to the
// NonRelativeService some-method server.
func EncodeSomeMethodRequest(encoder func(*http.Request) goahttp.Encoder) func(*http.Request, interface{}) error {
        return func(req *http.Request, v interface{}) error {
                p, ok := v.() // ←★The type required for casting is not specified.
                if !ok {
[...]

PR #2343, opened 2019-10-30
Reset payload type before inherit attributes for path params from parent service

tchssk:
#2342 Case 1


tchssk:
0952b1e98a9891ead6c828bc7891860a45a1752d will panic.

raphael:
Great fix, thank you!

PR #2345, opened 2019-10-31
Reset payload type before inherit attributes for path params from parent service

tchssk:
This is backport of #2343 to v3.


raphael:
Perfect, thank you!

PR #2348, opened 2019-10-31
Fix viewd result type uses unkeyed fields.

ikawaha:
# Example design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var MyResult = ResultType("application/add-result+json", func() {
	Attributes(func() {
		Attribute("result", Int)
	})
})

	Method("add", func() {
		Payload(func() {
			Required("a", "b")
		})
		Result(MyResult)
		HTTP(func() {
			GET("/add/{a}/{b}")
			Response(StatusOK)
		})
	})
})
```

# goa gen

gen/calc/service.go

**before**
```go
// NewViewedAddResult initializes viewed result type AddResult from result type
// AddResult using the given view.
func NewViewedAddResult(res *AddResult, view string) *calcviews.AddResult {
	var vres *calcviews.AddResult
	switch view {
	case "default", "":
		p := newAddResultView(res)
		vres = &calcviews.AddResult{p, "default"}  // ★★★
	}
	return vres
}
```

**after**
```go
// NewViewedAddResult initializes viewed result type AddResult from result type
// AddResult using the given view.
func NewViewedAddResult(res *AddResult, view string) *calcviews.AddResult {
	var vres *calcviews.AddResult
	switch view {
	case "default", "":
		p := newAddResultView(res)
		vres = &calcviews.AddResult{Projected: p, View: "default"}  //  ★★★
	}
	return vres
}
```

### DIFF
--- a/codegen/service/service_data.go
+++ b/codegen/service/service_data.go
@@ -1568,7 +1568,7 @@ const (
 		case {{ printf "%q" .Name }}{{ if eq .Name "default" }}, ""{{ end }}:
 			{{- if $.ToViewed }}
 				p := {{ $.InitName }}{{ if ne .Name "default" }}{{ goify .Name true }}{{ end }}({{ $.ArgVar }})
-				{{ $.ReturnVar }} = {{ if not $.IsCollection }}&{{ end }}{{ $.TargetType }}{ p,  {{ printf "%q" .Name }} }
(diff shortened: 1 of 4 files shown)


raphael:
Thank you! Can you please back port to v2 as well? thanks!

PR #2351, opened 2019-11-01
Validate a method when no security scheme but a security attribute

tchssk:


raphael:
This is great, thank you!

PR #2353, opened 2019-11-02
Validate a method when no security scheme but a security attribute

tchssk:
This is backport of #2351 to v3.


raphael:
Thank you!

PR #2354, opened 2019-11-03
v3: Fix a bug that caused a build error when a variable of a payload is not referenced

ikawaha:
In gRPC, a payload variable is always prepared 
even though the payload reference might be unnecessary.

**An example of design**

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

	Method("add", func() {
		StreamingPayload(func() {
			Required("a", "b")
		})
		Result(Int)
		GRPC(func() {
		})
	})
```

**Error**

```
$ go run ./cmd/calc
# calc/gen/grpc/calc/server
gen/grpc/calc/server/server.go:56:2: p declared and not used
```

**Generated code**

```go
// Add implements the "Add" method in calcpb.CalcServer interface.
func (s *Server) Add(stream calcpb.Calc_AddServer) error {
	ctx := stream.Context()
	ctx = context.WithValue(ctx, goa.MethodKey, "add")
	ctx = context.WithValue(ctx, goa.ServiceKey, "calc")
	p, err := s.AddH.Decode(ctx, nil)  // ← var p is not referenced anywhere
	if err != nil {
		return goagrpc.EncodeError(err)
	}
	ep := &calc.AddEndpointInput{
		Stream: &AddServerStream{stream: stream},
	}
	err = s.AddH.Handle(ctx, ep)
	if err != nil {
		return goagrpc.EncodeError(err)
	}
	return nil
}
```

### DIFF
--- a/grpc/codegen/server.go
+++ b/grpc/codegen/server.go
@@ -242,7 +242,7 @@ func (s *{{ .ServerStruct }}) {{ .Method.VarName }}(
 	ctx = context.WithValue(ctx, goa.ServiceKey, {{ printf "%q" .ServiceName }})
 
 {{- if .ServerStream }}
-	p, err := s.{{ .Method.VarName }}H.Decode(ctx, {{ if .Method.StreamingPayload }}nil{{ else }}message{{ end }})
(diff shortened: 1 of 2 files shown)


raphael:
This is great! thank you. Can you please backport to v2? thanks!

PR #2357, opened 2019-11-04
v2: Fix a bug that caused a build error when a variable of a payload is not referenced

ikawaha:
This is backporting of #2354


raphael:
Thank you!

PR #2364, opened 2019-11-05
Fix missing generated http client types

tchssk:
Duplicate http client types should  be omitted in each service but it is working for around the services for now.


tchssk:
d4e00ab8fdad94a3f1deb70cd97a13b7adbe2637 will fail.

raphael:
Thank you for the fix, makes sense.

raphael:
I ported the fix to `v3` in #2370

PR #2366, opened 2019-11-05
Fix short-circuiting in hasTag and iterate through all bases

nitinmohan87:


raphael:
Thank you! Can you please back-port to v2 as well?

PR #2362, opened 2019-11-05
v3: Create the template function for authentication in the file for each service

ikawaha:
see #2361 

## Example

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var BasicAuth = BasicAuthSecurity("basicauth")

var _ = Service("calc1", func() {
	Security(BasicAuth)
	Method("add", func() {
		Payload(func() {
			Username("username")
			Password("password")
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/add/{a}/{b}")
		})
	})
})

var _ = Service("calc2", func() {
	Security(BasicAuth)
	Method("mul", func() {
		Payload(func() {
			Username("username")
			Password("password")
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/mul/{a}/{b}")
		})
	})
})

```

## goa gen

The authentication method is in each service file.

calc1.go
```go
package calc1api

import (
	calc1 "calc/gen/calc1"
	"context"
	"fmt"
	"log"

	"goa.design/goa/v3/security"
)

// calc1 service example implementation.
// The example methods log the requests and return zero values.
type calc1srvc struct {
	logger *log.Logger
}

// NewCalc1 returns the calc1 service implementation.
func NewCalc1(logger *log.Logger) calc1.Service {
	return &calc1srvc{logger}
}

// BasicAuth implements the authorization logic for service "calc1" for the
// "basicauth" security scheme.
func (s *calc1srvc) BasicAuth(ctx context.Context, user, pass string, scheme *security.BasicScheme) (context.Context, error) {
	//
	// TBD: add authorization logic.
	//
	// In case of authorization failure this function should return
	// one of the generated error structs, e.g.:
	//
	//    return ctx, myservice.MakeUnauthorizedError("invalid token")
	//
	// Alternatively this function may return an instance of
	// goa.ServiceError with a Name field value that matches one of
	// the design error names, e.g:
	//
	//    return ctx, goa.PermanentError("unauthorized", "invalid token")
	//
	return ctx, fmt.Errorf("not implemented")
}

// Add implements add.
func (s *calc1srvc) Add(ctx context.Context, p *calc1.AddPayload) (res int, err error) {
	s.logger.Print("calc1.add")
	return
}
```
calc2.go
```go
[...]

Issue #2361, opened 2019-11-05
Proposal: Create the template function for authentication in the file for each service


ikawaha:
Security DSL defines authentication requirements to access a service or a service method.  An authentication template function generated by `goa example` is a method that uses each service as a receiver, but these methods for all services are collected in a single file called `auth.go` .

Collecting methods that have different receivers into one file could lead to poor readability.

I think that the method for authentication should be written to the file that describes the service of its receiver.

Does it make sense?

## Example

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var BasicAuth = BasicAuthSecurity("basicauth")

var _ = Service("calc1", func() {
	Security(BasicAuth)
	Method("add", func() {
		Payload(func() {
			Username("username")
			Password("password")
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/add/{a}/{b}")
		})
	})
})

var _ = Service("calc2", func() {
	Security(BasicAuth)
	Method("mul", func() {
		Payload(func() {
			Username("username")
			Password("password")
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/mul/{a}/{b}")
		})
	})
})

```

## goa gen (goa example)

**Current**

`calcsrvc1.BasicAuth()` and `calcsrvc2.BasicAuth()` are in same file `auth.go`.

auth.go
```go
package calc1api

import (
	"context"
	"fmt"

	"goa.design/goa/v3/security"
)

// BasicAuth implements the authorization logic for service "calc1" for the
// "basicauth" security scheme.
func (s *calc1srvc) BasicAuth(ctx context.Context, user, pass string, scheme *security.BasicScheme) (context.Context, error) {
	//
	// TBD: add authorization logic.
	//
	// In case of authorization failure this function should return
	// one of the generated error structs, e.g.:
	//
	//    return ctx, myservice.MakeUnauthorizedError("invalid token")
	//
	// Alternatively this function may return an instance of
	// goa.ServiceError with a Name field value that matches one of
	// the design error names, e.g:
	//
	//    return ctx, goa.PermanentError("unauthorized", "invalid token")
	//
	return ctx, fmt.Errorf("not implemented")
}

// BasicAuth implements the authorization logic for service "calc2" for the
[...]

PR #2368, opened 2019-11-06
v2: Create the template function for authentication in the file for each service

ikawaha:
issue #2361
This is backporting of #2362.


raphael:
Perfect, thank you!

PR #2369, opened 2019-11-06
Generated client returns nil for streaming payload with no result

tmcladam:
When there is a streaming payload with no result, the generated transport-agnostic service client was returning nil instead of the client stream.

New test case "streaming-payload-no-result" would fail without this change.


raphael:
Thank you! Can you please back port to v2 as well? thanks!

PR #2376, opened 2019-11-11
v2: Generated client returns nil for streaming payload with no result

tmcladam:
Generated client should return client stream instead of nil when there is a streaming payload with no result (v2 port of #2369)


raphael:
Thank you!

PR #2377, opened 2019-11-13
Use scope to generate http mount functions

tchssk:
`Files` should be able to serve a file as multiple paths.


tchssk:
f20a526a0abc9ab743e7603ff794ee52eb5b4856 will fail.

raphael:
This is great, thank you! Can you please port to `v3` as well? thanks!

PR #2382, opened 2019-11-19
v3: Files must appear in Service

ikawaha:
According to the specification, `Files` DSL must appear in `Service`.
But it is available in the `API`, etc...


raphael:
Whoops, good catch, thank you!

raphael:
Can you please back port to v2 as well? thank you.

PR #2383, opened 2019-11-19
v3: Add t.Helper() to test helper functions

ikawaha:
How about adding t.Helper function to test helper functions?


raphael:
This is great, thank you!

PR #2386, opened 2019-11-20
v3: Add metadata to Files DSL

ikawaha:
Since meta tags cannot be set for endpoints generated by Files, it was always `default` in OpenAPI files.

Modified to be able to specify tags for endpoints generated by Files, and to set the service name if not specified.

Example design:

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

	Method("add", func() {
		Payload(func() {
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/add/{a}/{b}")
		})
	})
	Files("/openapi.json", "./gen/http/openapi.json")
})
```

### Before:

Swagger_Editor

### After:

Swagger_Editor


raphael:
This is great, thank you! (back port to v2 as usual please :) )

PR #2388, opened 2019-11-22
v1: Prevent directory path traversal in FileHandler

christi3k:
Presently `FileHandler` allows `../` which enables directory path traversal when a `Files` endpoint is defined with a wildcard.

Example API design definition:

```
var _ = Resource("docs", func() {
	Description("Documentation endpoint")

	Origin("*", func() {
		Methods("GET", "OPTIONS")
	})

	Files("/docs/*filepath", "swagger-ui")
})
```

Example exploit:

```
curl "http://localhost/docs/..%2F..%2F..%2Fetc%2F"
```

This patch updates `FileHandler` to return `ErrNotFound` if a `../` is found in `req.URL.Path`.

I wanted to submit tests along with this but got stuck trying to make them work. If someone could provide guidance, I would be happy to make another attempt.

Existing tests do pass:

```
$ make test
Ginkgo ran 25 suites in 1m1.149089831s
Test Suite Passed
go test ./_integration_tests
ok  	github.com/goadesign/goa/_integration_tests	20.423s
```

Additionally, I've applied this patch to a local dev instance of our app and verified the changes operates as expected. A curl request such as the one does return a 404.

### DIFF
--- a/service.go
+++ b/service.go
@@ -372,6 +372,10 @@ func (ctrl *Controller) FileHandler(path, filename string) Handler {
 		}
 	}
 	return func(ctx context.Context, rw http.ResponseWriter, req *http.Request) error {
+		// prevent path traversal
+		if containsDotDot(req.URL.Path) {
+			return ErrNotFound(req.URL.Path)
+		}
 		fname := filename
 		if len(wc) > 0 {
 			if m, ok := ContextRequest(ctx).Params[wc]; ok {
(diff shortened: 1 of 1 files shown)


raphael:
Thank you for the PR, this is great! I wonder if it wouldn't be more robust to use [os.PathSeparator](https://golang.org/pkg/os/#PathSeparator) though instead of hard-coding the file path separator - this would alleviate the need to check for both types of separators which could actually result in faulty behavior (e.g. if a unix file has a backslash in its name). Another (better?) approach would be to use [IsPathSeparator](https://golang.org/pkg/os/#IsPathSeparator) instead of a custom function to splice.

christi3k:
Thanks for the suggestion to use IsPathSeparator. I've updated the isSlashRune() function to use it.

While I was at it, I also adjusted the check to allow:

- `..` in filenames (legal in *nix and Windows, I believe)
- relative paths using `..` _within_ the defined path mount point (but not outside/above)

`make tests` all passed and I tested a few different paths with a running version against our app.

raphael:
Looks great, thank you.

PR #2397, opened 2019-12-02
Adding zero to dsl

laches1sm:
*What does this PR change?*

Brings in Zero into the DSL and the gRPC transformation logic. This means that:

- if no zero value is defined, then just put the default zero value, but if one is defined, then we set that instead, i.e. I set a zero val of -10000, I get a protobuf request in, the zero value of that field defined is now -10000 (reword this once this PR is out of drafts!)

I'm still fixing up the validation, hence why this PR will not be ready to review until I get this sorted. If anyone is interested in fighting through Go templates with me, feel free to give me a shout in the comments :) this is more for tracking stuff for me :)


codegen/validation.go
```diff
@@ -477,7 +485,14 @@ if {{ if .string }}utf8.RuneCountInString({{ $target }}){{ else }}len({{ $target
 }
 {{- end }}`
 
-	requiredValTmpl = `if {{ $.target }}.{{ .attCtx.Scope.Field $.reqAtt .req true }} == nil {
-        err = goa.MergeErrors(err, goa.MissingFieldError("{{ .req }}", {{ printf "%q" $.context }}))
-}`
+	requiredValTmpl = `if {{ $.target }}.{{ .attCtx.Scope.Field $.reqAtt .req  true }} == nil {
```
laches1sm, line 488: @nitinmohan87 @raphael I hope you don't mind me tagging you, but I'm a little stumped on how to handle the validation templates. This is sort of a first try at it, going off of the validation set out in the issue about this, it's currently not working :) I haven't really messed about with Go templates before (had a look at the documentation, but that's all my experience with it) so it would be great if you could give me some pointers :D

PR #2401, opened 2019-12-04
Properly handle struct:field:type metadata

### LINKED ISSUE #2399: Code generated when using `struct:field:type` meta does not compile
The code generated when using `Meta("struct:field:type"...)` will not compile.

A project is worth a thousand words, so here's a zip file of a project containing a minimal reproduction case. Code was generated with 3.0.8.

[goabug.zip](https://github.com/goadesign/goa/files/3919398/goabug.zip)

raphael:
When generating HTTP marshaling code.

Fix #2399

### DIFF
--- a/codegen/import.go
+++ b/codegen/import.go
@@ -64,10 +64,9 @@ func (s *ImportSpec) Code() string {
 	return fmt.Sprintf(`"%s"`, s.Path)
 }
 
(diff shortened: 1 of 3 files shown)


tylerstillwater:
That was fast! Thanks!

raphael:
Thank you for the great report and repro. The issue is fixed in the head of the v3 and v2 branches.

PR #2404, opened 2019-12-05
v3: Fix decoding of queries with string slices set as default

ikawaha:
**Design**
```go
var PayloadQueryStringSliceDefaultDSL = func() {
	Service("ServiceQueryStringSliceDefault", func() {
		Method("MethodQueryStringSliceDefault", func() {
			Payload(func() {
				Attribute("q", ArrayOf(String), func() {
					Default([]string{"hello", "goodbye"}) // string slice
				})
			})
			HTTP(func() {
				GET("/")
				Param("q")
			})
		})
	})
}
```

**before**

```go
func(*http.Request) goahttp.Decoder) func(*http.Request) (interface{}, error) {
	return func(r *http.Request) (interface{}, error) {
		var (
			q []string
		)
		q = r.URL.Query()["q"]
		if q == nil {
			q = []interface{"hello", "goodbye"} // ops! r.URL.Query()["q"] returns []string
		}
		payload := NewMethodQueryStringSliceDefaultPayload(q)

		return payload, nil
	}
}
```

**after**
```go
		q = r.URL.Query()["q"]
		if q == nil {
			q = []string{"hello", "goodbye"} // correct type
		}
```

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -653,7 +653,11 @@ const requestParamsHeadersT = `{{- define "request_params_headers" }}
 		}
 		{{- else if .DefaultValue }}
 		if {{ .VarName }} == nil {
-			{{ .VarName }} = {{ printf "%#v" .DefaultValue }}
+			{{ .VarName }} = []string{
+                {{- range $i, $v := .DefaultValue }}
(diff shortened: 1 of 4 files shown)


raphael:
Thank you for the great PR!

PR #2406, opened 2019-12-06
v2: [backporting] Fix decoding of queries with string slices set as default (#2404)

ikawaha:
This is backporting of #2404

* Add tests for decoding the query
* Fix decoding of queries with string slices set as default


raphael:
Perfect, thank you!

PR #2411, opened 2019-12-08
Trim spaces for middleware.Logger

tchssk:
## Before

```
 id=XXXXXXXX req=GET /foo from=127.0.0.1
^
```

## After

```
id=XXXXXXXX req=GET /foo from=127.0.0.1
```


raphael:
Great, thank you!

PR #2419, opened 2019-12-10
Properly handle server main handler

tchssk:
Generated server main handlers have to receive only endpoints for the transport.

```
// anotherServiceEndpoints is unneeded if gRPC is defined only for serviceEndpoints.
handleGRPCServer(ctx, u, serviceEndpoints, anotherServiceEndpoints, &wg, errc, logger, *dbgF)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/codegen/example/example_server.go
+++ b/codegen/example/example_server.go
@@ -278,7 +278,7 @@ func main() {
 			} else if u.Port() == "" {
 				u.Host += ":{{ $u.Port }}"
 			}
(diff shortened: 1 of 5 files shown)


tchssk:
1f0620aeb228014c9fa676bfae66ea4c439d2150 will fail.

raphael:
Yes you're right - good catch, thank you!

PR #2424, opened 2019-12-13
v3: Fix a bug of the file server

ikawaha:
This patch fixes a bug of the file server when there is a wildcard in the root of the file path.

**Design**

```go
Service("ServiceFileServer", func() {
	Files("/{wildcard}", "/path/to/folder")
})
```

**Before**

```go
func MountPathToFolder(mux goahttp.Muxer, h http.Handler) {
	mux.Handle("GET", "/{wildcard}/", h.ServeHTTP)
	mux.Handle("GET", "/{wildcard}/*wildcard", h.ServeHTTP)
}
```

**After**
```go
func MountPathToFolder(mux goahttp.Muxer, h http.Handler) {
	mux.Handle("GET", "/", h.ServeHTTP)
	mux.Handle("GET", "/*wildcard", h.ServeHTTP)
}
```


raphael:
Great catch, thank you!

PR #2427, opened 2019-12-14
v2: Fix a bug of the file server (#2424)

ikawaha:
This is backporting of #2424 to v2.

* Add tests for file servers
* Fix a bug of the file server when there is a wildcard in the root of the file path


raphael:
Thank you!

Issue #2429, opened 2019-12-14
Required parameters inherited by Parent DSL are represented by pointers in the child


ikawaha:
Does `Parent` DSL correctly inherit path parameter properties?
When inheriting a required parameter in the parent, it is represented by a pointer in the Payload structure in a child. 

## Example design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = Service("parent", func() {
	Description("The calc service performs operations on numbers.")
	HTTP(func() {
		CanonicalMethod("parent-method")
	})
	Method("parent-method", func() {
		Payload(func() {
			Attribute("a", Int, "parent path attribute")
			Required("a")
		})
		Result(Int)
		HTTP(func() {
			GET("/hello/{a}/")
		})
	})
})

var _ = Service("child", func() {
	HTTP(func() {
		Parent("parent")
	})

	Method("child-method", func() {
		Payload(func() {
			Attribute("b")
			Required("b")
		})
		Result(Int)
		HTTP(func() {
			GET("/goodbye/{b}")
		})
	})
})
```

## Generated payloads

Parent path parameter `a` is required. So `A` is not pointer variable.
```go
type ParentMethodPayload struct {
	// parent path attribute
	A int
}
```

Child inherent path parameter `a` from Parent service. 
`A` is pointer variable.

```go
// ChildMethodPayload is the payload type of the child service child-method
// method.
type ChildMethodPayload struct {
	B string
	// parent path attribute
	A *int
}
```

ikawaha:
╭( ･ㅂ･)و ̑̑

PR #2434, opened 2019-12-15
Fix a bug that required is not inherited by Parent DSL

ikawaha:
see. #2429

### DIFF
--- a/expr/http_endpoint.go
+++ b/expr/http_endpoint.go
@@ -206,7 +206,7 @@ func (e *HTTPEndpointExpr) Prepare() {
 								if e.MethodExpr.Payload.Validation == nil {
 									e.MethodExpr.Payload.Validation = &ValidationExpr{}
 								}
-								e.MethodExpr.Payload.Validation.Required = append(e.MethodExpr.Payload.Validation.Required, name)
+								e.MethodExpr.Payload.Validation.AddRequired(name)
 							}
 							o.Set(name, att)
 						}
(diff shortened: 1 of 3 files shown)


expr/http_endpoint.go
```diff
@@ -192,18 +192,23 @@ func (e *HTTPEndpointExpr) Prepare() {
...
 						}
-						if o := AsObject(e.MethodExpr.Payload.Type); o != nil {
-							if o.Attribute(name) == nil {
-								o.Set(name, att)
+						if o := AsObject(e.MethodExpr.Payload.Type); o != nil && o.Attribute(name) == nil {
+							if c.MethodExpr.Payload.IsRequired(name) {
+								if e.MethodExpr.Payload.Validation == nil {
+									e.MethodExpr.Payload.Validation = &ValidationExpr{}
+								}
+								e.MethodExpr.Payload.Validation.Required = append(e.MethodExpr.Payload.Validation.Required, name)
```
raphael: There is a method that makes this easier, it takes care of removing duplicate entries in the `Required` field: https://github.com/goadesign/goa/blob/21aa53219da0de90202ca13f99c612bcb594c817/expr/attribute.go#L630

raphael (commented):
This is great! thank you very much! I left a comment for a small change.

ikawaha:
Thank you for your suggestion!

raphael:
Looks great, thank you!

PR #2435, opened 2019-12-16
Fix a bug that required is not inherited by Parent DSL (#2434)

ikawaha:
This is backporting of #2434 to v2.
see. #2429

* Fix a bug that required is not inherited by Parent DSL
* Add tests whether required is inherited by Parent
* Use AddRequired() instead of append()


raphael:
Thank you!

PR #2438, opened 2019-12-20
Fix usage of Debug in example http server

tchssk:
Generated example HTTP server contains the Debug middleware but request parameters are not printed. The middleware depends on httptreemux.ContextParams so it must be applied for each servers before mount them.

### DIFF
--- a/http/codegen/example_server.go
+++ b/http/codegen/example_server.go
@@ -256,6 +256,19 @@ func handleHTTPServer(ctx context.Context, u *url.URL{{ range $.Services }}{{ if
 		{{ .Service.VarName }}Server = {{ .Service.PkgName }}svr.New(nil, mux, dec, enc, eh, nil)
 		{{-  end }}
 	{{- end }}
+	{{- if .Services }}
+		if debug {
+			for _, server := range []interface {
+				Use(func(http.Handler) http.Handler)
+			}{
+				{{- range $svc := .Services }}
+				{{ .Service.VarName }}Server,
+				{{- end }}
+			} {
+				server.Use(httpmdlwr.Debug(mux, os.Stdout))
+			}
+		}
+	{{- end }}
 	}
 	// Configure the mux.
 	{{- range .Services }}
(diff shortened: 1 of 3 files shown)


http/codegen/example_server.go
```diff
@@ -256,6 +256,19 @@ func handleHTTPServer(ctx context.Context, u *url.URL{{ range $.Services }}{{ if
...
+				Use(func(http.Handler) http.Handler)
+			}{
+				{{- range $svc := .Services }}
+				{{ .Service.VarName }}Server,
+				{{- end }}
+			} {
+				server.Use(httpmdlwr.Debug(mux, os.Stdout))
+			}
+		}
+	{{- end }}
```
tchssk, line 268: It might be better to define an interface like this:

```go
type (
	Server interface {
		Use(func(http.Handler) http.Handler)
	}

	Servers []Server
)

func (s Server) Use(m func(http.Handler) http.Handler) {
	for _, v := range s {
		v.Use(m)
	}
}

Comment (raphael) [reply http/codegen/example_server.go:268]: Yes that makes sense (I guess you meant `func (s Servers)`), would be a lot cleaner.

**Comment by raphael** [comment]
This is great, thank you for the fix. I agree that we may want to introduce a new interface type to represent the set of servers.

File: http/server.go [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+import "net/http"
+
+type (
+	// Server is the HTTP server interface used to wrap the server handlers
+	// with the given middleware.
+	Server interface {
+		Use(func(http.Handler) http.Handler)
+	}
+
+	// Servers is the list of server.
```
raphael: Servers is a list of servers.

raphael:
This is great, thank you! I left a little comment nitpick.

raphael:
server**s** :) - thank you!

tchssk:
Oops 😓

raphael:
Great, thank you!

PR #2455, opened 2020-01-14
Fixes streaming endpoints with no payload 

tchssk:
This is backporting of #2246 to v2.

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -464,6 +464,8 @@ func {{ .HandlerInit }}(
 			}
 			return
 		}
+	{{- else }}
+		var err error
 	{{- end }}
 
 	{{ if .ServerStream }}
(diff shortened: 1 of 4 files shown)


http/codegen/server.go
```diff
@@ -464,6 +464,8 @@ func {{ .HandlerInit }}(
 			}
 			return
 		}
+	{{- else }}
```
raphael, line 467: This could be moved below the `{{ if .ServerStream }}` below since the other case already defines `err`:
```go
{{ if .ServerStream }}
	var err error
	var cancel context.CancelFunc
	// ...
```

raphael:
Thank you! I left a small improvement in a comment.
Edit: I didn't realize this was a backport - thank you for catching this! We should probably leave this as is so the code generated by v2 and v3 is the same. Maybe we can make the change I'm suggesting in a different set of PRs.

PR #2464, opened 2020-01-15
goa -> Goa

tchssk:
This is backporting of e329512 to v2.


raphael:
This is great, thanks!

PR #2465, opened 2020-01-15
Updated the cast for CLI args for []byte payload from string() to []byte()

tchssk:
This is backporting of #2196 to v2.


raphael:
Thank you!

PR #2470, opened 2020-01-21
Properly close gRPC client streams without result

MichaelUrman:
Previously a client stream without a result was not closed as described in [gRPC Basics](https://grpc.io/docs/tutorials/basic/go/), namely the [server side](https://grpc.io/docs/tutorials/basic/go/#client-side-streaming-rpc) did not call SendAndClose, and the [client side](https://grpc.io/docs/tutorials/basic/go/#client-side-streaming-rpc-1) did not call CloseAndRecv. Nor did the goa wrappers offer any way to call these methods.

This PR adds to the existing `client-streaming-no-result` test to verify that the Close methods are changed to call those respective methods, and then modifies the `streamCloseT` template to pass the test.

Testing this with a local client and server that has synchronous code that previously behaved asynchronously, this change appears to make it synchronous as expected.

### DIFF
--- a/grpc/codegen/service_data.go
+++ b/grpc/codegen/service_data.go
@@ -1340,11 +1340,22 @@ func (s *{{ .VarName }}) {{ .RecvName }}() ({{ .RecvRef }}, error) {
 const streamCloseT = `
 func (s *{{ .VarName }}) Close() error {
 {{- if eq .Type "client" }}
+{{- if .Endpoint.Method.Result }}
 	{{ comment "Close the send direction of the stream" }}
 	return s.stream.CloseSend()
 {{- else }}
+	{{ comment "synchronize and report any server error" }}
+	_, err := s.stream.CloseAndRecv()
+	return err
+{{- end }}
+{{- else }}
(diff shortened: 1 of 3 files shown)


MichaelUrman:
Here are two things I'm unsure of:

1. Is this a reasonable change to the template? Incrementally it's pretty trivial, but collectively the template is now getting a little unwieldy.
2. Does this enable proper handling of io.EOF from the underlying SendMsg call, [as documented](https://godoc.org/google.golang.org/grpc#ClientStream), namely is it okay to call CloseAndRecv instead of merely RecvMsg after an io.EOF? This PR does not currently expose a way for the client to call just RecvMsg.

raphael:
Thank you for the PR, this is great! My understanding is that calling `CloseAndRecv` is indeed sufficient in this case (see the comment on [NewStream](https://godoc.org/google.golang.org/grpc#ClientConn.NewStream)). Would you mind back-porting this fix to v2 as well? Thank you!.

MichaelUrman:
👍 See #2476 for backport.

PR #2471, opened 2020-01-22
V2: [OpenAPI] Convert the type name if primitive

ikawaha:
This is backporting of https://github.com/goadesign/goa/pull/2469 to v2.


raphael:
Great, thank you!

PR #2474, opened 2020-01-25
V3: Fix to use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline)

ikawaha:
```
LINTING CODE...
basic/cmd/calc-cli/grpc.go:15:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)
cellar/cmd/cellar-cli/grpc.go:15:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)
error/cmd/divider-cli/grpc.go:15:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)
security/cmd/multi_auth-cli/grpc.go:15:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)
streaming/cmd/chatter-cli/grpc.go:15:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)
tracing/cmd/calc-cli/grpc.go:26:3: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (S1038)

^ - staticcheck errors!
```

PR #2475, opened 2020-01-26
V2: Fix to use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (#2474)

ikawaha:
This is backporting of #2474 to v2.


raphael:
Thank you!

PR #2476, opened 2020-01-27
Properly close gRPC client streams without result (Backport #2470 to v2)

MichaelUrman:
Backport #2470 to v2. I just did a cherry-pick followed by the fun of remembering how to test non-go.mod.

I see a failing test (TestValidateFormat) both before and after my changes, and the expected temporary test failure between my two commits. I have only run go test; I have not tested client behavior here, but see no reason to expect difficulties.


raphael:
Thank you for all the work here, much appreciated! The build is green so I feel confident it's all good.

PR #2482, opened 2020-02-05
fix client streaming with no response

dvictor:
For client streaming that does not expect a response, the goa stream.Close in server was calling SendAndClose(nil). That was causing a marshalling error because even if the response of the endpoint has no value, it exists as a (empty) type. Clients ignoring the result of stream.Close have worked, but that hides a potential network or other errors when closing the stream.

Before backporting to goa v2, I'd like some feedback on this. Does it look ok?

### DIFF
--- a/grpc/codegen/service_data.go
+++ b/grpc/codegen/service_data.go
@@ -841,7 +841,7 @@ func buildRequestConvertData(request, payload *expr.AttributeExpr, md []*Metadat
 //
 // svr param indicates that the convert data is generated for server side.
 func buildResponseConvertData(response, result *expr.AttributeExpr, svcCtx *codegen.AttributeContext, hdrs, trlrs []*MetadataData, e *expr.GRPCEndpointExpr, sd *ServiceData, svr bool) *ConvertData {
-	if e.MethodExpr.IsStreaming() || (!svr && isEmpty(e.MethodExpr.Result.Type)) {
+	if !svr && (e.MethodExpr.IsStreaming() || isEmpty(e.MethodExpr.Result.Type)) {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


raphael:
Question: shouldn't the user code be able to specify the value to give to `SendAndClose`?

dvictor:
> Question: shouldn't the user code be able to specify the value to give to `SendAndClose`?

No, because the value is always the same: the empty struct of the response type. It has no fields.

raphael:
I understand that this is what the current code does but in general it seems that users should be able to specify the value given to `SendAndClose` no? For example https://grpc.io/docs/tutorials/basic/go/ shows the code passing a non empty struct.

dvictor:
We are talking about client streaming case with _no response from server_. At the gRPC/pb level, there's no difference between no response or some return value type. The api is the same. At the goa level though, if you don't define a response type, the interface is simpler. Instead of a SendAndClose, you only get the Close function on the stream object. Previous to Urman's fix, that function was empty. His fix called the empty s.stream.SendAndClose(nil), but that nil is not what gRPC code expects. It expects the pb message type with no fields.

MichaelUrman:
For some additional context, if the goa design specifies a Result (other than Empty), different templates are leveraged, [exposing SendAndClose(_resultType_)](https://github.com/goadesign/goa/blob/v3/grpc/codegen/testdata/streaming_code.go#L202-L209) in the goa gRPC server interface. When the result is unspecified or empty, the server [offers just Close()](https://github.com/goadesign/goa/blob/v3/grpc/codegen/testdata/streaming_code.go#L253-L257), and the client similarly hides the empty protobuf message from its API.

The nil in the latter is a problem (it marshals...poorly), and I missed it in #2470 because I hadn't wired up the server side.

raphael:
Thank you for the clarification, makes sense - I had missed the fact that this only applied for methods with empty results.

PR #2489, opened 2020-02-09
Fix extra space of paths

### LINKED ISSUE #2487: v3: Extra space is inserted when using example cli
I wrote design as followings.

```go
var _ = Service("test", func() {
	Method("add", func() {
		Payload(func() {
			Attribute("users", ArrayOf(String, func() {}))
		})

		HTTP(func() {
			GET("/add/{users}")
		})
	})
})
```

And, I generate codes by `goa gen` and `goa example`.
And I run `test-cli  test add --users '["hello", "world"]'`, But server got `["hello", "<Space>world"]`.

I think it is because users are Joined as followings in `gen/http/test/client/paths.go`

```go
fmt.Sprintf("/add/%v", strings.Join(usersSlice, ",<Space>"))
```

kitagry:
fix #2487

### DIFF
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -2559,7 +2559,7 @@ const (
 		{{- end }}
 	{{- end }}
 	return fmt.Sprintf("{{ .PathFormat }}", {{ range $i, $arg := .Args }}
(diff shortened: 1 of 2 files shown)


raphael:
This is great, thank you! Do you think you could backport this to v2 as well? Just create a branch off of v2 and cherry-pick this merge commit.

kitagry:
Thank you!
I created PR for v2. #2490

PR #2493, opened 2020-02-16
Fix reference to StatusOK in ContentType docs

sevein:


raphael:
Whoops, thank you! Would you mind back-porting to `v2` as well? Thanks!

PR #2494, opened 2020-02-18
Fix reference to StatusOK in ContentType docs

sevein:
Port of https://github.com/goadesign/goa/pull/2493 for v2.


raphael:
Thanks!

PR #2516, opened 2020-03-23
Don't set default value for required attribute in streaming payload gRPC

### LINKED ISSUE #2513: Required fields with a Default act differently for Unary and Streaming gRPC payloads
When a field of a type is [marked Required, or supplied a Default](https://github.com/MichaelUrman/goa-grpc-defaults/blob/master/design/design.go#L14-L36), the behavior depends on where it is used: unary payloads ignore Default on required fields; streaming payloads apply Default regardless.

(I haven't looked closely at the HTTP side, as my focus was backwards compatibility for gRPC APIs. But I couldn't get past missing_payload errors.)

See https://github.com/MichaelUrman/goa-grpc-defaults for a tweaked CalcSvc example that calls this out, especially the resulting difference between the [unary payload builder](https://github.com/MichaelUrman/goa-grpc-defaults/blob/master/gen/grpc/calc/server/types.go#L74-L83) and the [streaming payload builder](https://github.com/MichaelUrman/goa-grpc-defaults/blob/master/gen/grpc/calc/server/types.go#L154-L164).

nitinmohan87:
Fixes #2513

### DIFF
--- a/expr/grpc_endpoint.go
+++ b/expr/grpc_endpoint.go
@@ -319,6 +319,13 @@ func (e *GRPCEndpointExpr) Finalize() {
 	// Finalize streaming payload type if defined
 	if e.MethodExpr.StreamingPayload.Type != Empty {
 		initAttrFromDesign(e.StreamingRequest, e.MethodExpr.StreamingPayload)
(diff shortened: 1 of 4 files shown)


raphael:
Thank you! Can you please port back to v2 as well? thanks!

PR #2524, opened 2020-04-02
More flexible arguments for Response DSL

tchssk:
This patch makes it possible to accept the name of the error as second argument of the Response DSL. It allows you to align the first argument of error responses with successful responses.

```go
Response(StatusOK)
Response(StatusNotFound, "not_found")
Response(StatusInternalServerError, "internal_server_error")
```

### DIFF
--- a/dsl/response.go
+++ b/dsl/response.go
@@ -95,14 +95,12 @@ import (
 //
 func Response(val interface{}, args ...interface{}) {
 	name, ok := val.(string)
-	if !ok {
-		if len(args) > 0 {
-			name, ok = args[0].(string)
-			if ok {
-				arg := args[0]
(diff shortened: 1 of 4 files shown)


dsl/response.go
```diff
@@ -91,6 +95,16 @@ import (
 //
 func Response(val interface{}, args ...interface{}) {
 	name, ok := val.(string)
+	if !ok {
```
raphael: Maybe combine these two lines?
```go
if !ok && len(args) > 0 {
// ....
```

tchssk: I fixed it. 0e98e72

raphael:
Thank you!

Issue #1283, opened 2017-06-26
Trailing Slash in Routing removed


Ragnaroek:
Action("something", func() {
		Routing(POST("/path/with/trailing/slash/"))
})

Goa seems to remove (haven't found where) to remove the trailing slash before the path is given to
httptreemux. This is bad, because we have a client that calls our endpoint with a trailing slash and does not follow redirect. The redirect to the non slash version is automatically done by httptreemux. I need to have the reverse: the non-slash version redirects to the slash version.

raphael:
https://github.com/goadesign/goa/blob/master/design/definitions.go#L766 and https://github.com/goadesign/goa/blob/master/design/definitions.go#L781 is where the paths are  computed. The algorithm uses `path.Join` which removes the trailing slashes.  It would probably be OK if `FullPath` added back the trailing slash if there was one originally. I'd be happy to merge a PR that did that...

Ragnaroek:
Oh well, I looked at the code but didn't see that!
I will try to make a PR today.

raphael:
Thank you for the PR!, closing the issue.

PR #2530, opened 2020-04-13
v3: Fix trailing slash in routing removed

ikawaha:
The routing trailing slash problem fixed in v1 seems to be occurring again in v3.
This patch fixes it in the same way as v1.

see also. https://github.com/goadesign/goa/issues/1283

### DIFF
--- a/expr/http_endpoint.go
+++ b/expr/http_endpoint.go
@@ -887,6 +887,9 @@ func (r *RouteExpr) FullPaths() []string {
 	res := make([]string, len(bases))
 	for i, b := range bases {
 		res[i] = httppath.Clean(path.Join(b, r.Path))
+		if strings.HasSuffix(r.Path, "/") {
+			res[i] += "/"
+		}
 	}
 	return res
 }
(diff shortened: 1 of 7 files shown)


raphael:
Great, thank you! Can you please also back-port to v2, thanks!

trestletech:
Is the handling here different for POST requests? I'm seeing 

My design is:

```
		HTTP(func() {
			POST("/blah/")
			Response(StatusOK)
		})
```

but I'm seeing:

```
[srvrapi] 21:50:27 id=oEcoo0o3 req=POST /blah/ from=172.17.0.1
[srvrapi] 21:50:27 id=oEcoo0o3 status=301 bytes=0 time=143.6µs
[srvrapi] 21:50:27 id=SUOWXrMt req=GET /blah from=172.17.0.1
[srvrapi] 21:50:27 id=SUOWXrMt status=405 bytes=0 time=68.5µs
```

(where the 405 indicates that the GET call to that path is invalid). I tried updating to the latest commit using `go get -u goa.design/goa/v3@4064ab219b39e4741ec61a06f62b6ba5906ffae0` but still see this behavior.

trestletech:
Just confirmed that if I add the trailing slash back into `mux.Handle("POST", "/learnr", f)` in the generated `server.go`, it starts working for me.

Did I fail to update to the right version, or is this bug still open for POST requests?

trestletech:
Oops. I think I was forgetting to rerun generate. Once I did that, I think we're back online. Thanks for the fix and sorry for the false alarm!

PR #2531, opened 2020-04-13
v2: Fix trailing slash in routing removed

ikawaha:
This is the backporting of #2530 to v2.

* Fix trailing slash in routing removed
* Exclude when the path is the root
* Fix a test


raphael:
Thank you!

PR #2532, opened 2020-04-14
V3: Fix generated endpoints when using SkipRequestBodyEncodeDecode with BasicAuth

achow-flexera:
Fixed using incorrect type on the generated endpoints when using SkipRequestBodyEncodeDecode() with BasicAuth. 

@raphael and @nitinmohan87, 
When I added a test for JWTAuth, it returned
`HTTP endpoint request body must be empty when using SkipRequestBodyEncodeDecode but not all method payload attributes are mapped to headers and params`

Just want to verify if other authentication types are supported?

### DIFF
--- a/codegen/service/security_test.go
+++ b/codegen/service/security_test.go
@@ -78,7 +78,6 @@ func TestSecureWithSkipRequestBodyEncodeDecode(t *testing.T) {
 		Code string
 	}{
 		{"with-basicauth", testdata.EndpointWithBasicAuthAndSkipRequestBodyEncodeDecodeDSL, testdata.EndpointWithBasicAuthAndSkipRequestBodyEncodeDecodeCode},
-		//{"with-jwtauth", testdata.EndpointWithJWTAuthAndSkipRequestBodyEncodeDecodeDSL, ""},
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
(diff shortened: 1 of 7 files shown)


codegen/service/security_test.go
```diff
@@ -70,3 +70,31 @@ func TestSecureEndpoint(t *testing.T) {
...
 }
+
+func TestSecureWithSkipRequestBodyEncodeDecode(t *testing.T) {
+	cases := []struct {
+		Name string
+		DSL  func()
+		Code string
+	}{
+		{"with-basicauth", testdata.EndpointWithBasicAuthAndSkipRequestBodyEncodeDecodeDSL, testdata.EndpointWithBasicAuthAndSkipRequestBodyEncodeDecodeCode},
+		//{"with-jwtauth", testdata.EndpointWithJWTAuthAndSkipRequestBodyEncodeDecodeDSL, ""},
```
raphael: Can you please remove this commented out line? thank you!

raphael:
Looks great, thank you for the fix! Can you please rebase on top of the latest v3 so the build passes?

raphael:
Also I'm taking a look at the JWT use case - I believe Goa should be able to tell that JWT uses HTTP headers and therefore should not produce this validation error.

raphael:
Update: the issue with JWT should now be fixed. Please rebase with the latest `v3` to get the fix, thank you!

codegen/service/security_test.go
achow-flexera: Done

raphael:
Thank you! Can you please backport this to `v2` as well? thanks!

PR #2533, opened 2020-04-14
Fix build status badge to show only event=push

tchssk:
The build status badge should not be affected by pull requests.


raphael:
Makes sense - I fixed the build in v3, can you rebase this PR on top so the build passes? thank you!

raphael:
Thank you!

PR #2551, opened 2020-05-07
Fail validation if HTTP request body is non-empty in a streaming endpoint

nitinmohan87:


raphael:
I see that we still allow having both `Payload` and `StreamingPayload` for endpoints. What is the semantic? should these be exclusive?

nitinmohan87:
In websocket protocol, an initial GET request is sent from the client to initiate upgrade to websocket connection. We will be passing the attributes `Payload` in that GET request as headers/params. Once the connection is upgraded server-side, client sends the `StreamingPayload` type through the websocket.

raphael:
Makes sense, thank you! Can you please back-port this to `v2` as well? thanks!

PR #2563, opened 2020-05-21
Set bearer in client request encode

nitinmohan87:
HTTP client didn't encode Authorization header with Bearer scheme for JWT and OAuth2 by default. This is because, we computed scheme data from the Method requirements instead of endpoint requirements. We modify the endpoint scheme in the [HTTP endpoint finalize](https://github.com/goadesign/goa/blob/v3/expr/http_endpoint.go#L600-L633) to include "Authorization" scheme name.

This PR fixes how we compute requirement data for HTTP endpoints.


raphael:
Thank you! Can you please back-port the fix to v2? thanks!

PR #2566, opened 2020-06-01
Enable ContentType DSL for HTTP error responses

tchssk:
Now It seems `ContentType` DSL is disabled for HTTP error responses. This patch enables `ContentType` DSL for HTTP error responses.


raphael:
This is great, thank you! Can you please back-port to `v2` as well? thanks!

PR #2568, opened 2020-06-02
Lower camel case conversion

nitinmohan87:


raphael:
Thank you! can you please back-port to v2 as well?

PR #2596, opened 2020-06-16
"struct:error:name" and "struct:field:name" not work together

mrsndmn:
Following code 
```go
var Result = ResultType("application/vnd.goa.error", func() {
    Attribute("error", String, func() {
        Meta("struct:error:name")
        Meta("struct:field:name", "ErrorCode")
    })
    Required("error")
})
```
Generated code:
```go
type GoaError struct {
    ErrorCode string
}

// Error returns an error description.
func (e *GoaError) Error() string {
    return ""
}

// ErrorName returns "GoaError".
func (e *GoaError) ErrorName() string {
    return e.Error
}

```
This code won't be compiled:
```
cannot use e.Error (type func() string) as type string in return argument
```

See full example of DSL and generated code in tests below.

### DIFF
--- a/codegen/service/service.go
+++ b/codegen/service/service.go
@@ -173,7 +173,7 @@ func errorName(et *UserTypeData) string {
 	if obj != nil {
 		for _, att := range *obj {
 			if _, ok := att.Attribute.Meta["struct:error:name"]; ok {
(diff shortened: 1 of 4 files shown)


raphael:
Good catch! thank you. Can you please back-port to the `v2` branch as well? (simply cherry-pick the merge commit).

PR #2597, opened 2020-06-16
"struct:error:name" and "struct:field:name" not work together (v2 back-port)

mrsndmn:
Back-port of https://github.com/goadesign/goa/pull/2596 for `v2`


raphael:
Thank you!

PR #2603, opened 2020-07-05
Fix graceful shutdown

shunyy:
Fixed invalid context value  passing to `context.WithTimeout`  for graceful-shutdown.

Current generated example of graceful-shutdown is calling `context.WithTimeout` with already closed context, because of that server shutdown immediately. I don't think it is intentional behavior, so fixed with passing `context.Background` to `context.WithTimeout`

### explanation
```golang
<-ctx.Done() // Closed!!

ctx, cancel := context.WithTimeout(ctx, 30*time.Second) // already closed ctx 😢 
defer cancel()

srv.Shutdown(ctx) // does not wait for timeout as intended
```

changes to examples can be reviewed in: https://github.com/goadesign/examples/pull/71


raphael:
Whoops, thank you so much! great catch.

raphael:
Would you mind back-porting this fix to v2 please? simply cherry-pick the merge commit - thank you!

PR #2604, opened 2020-07-05
Add SIGTERM handling

shunyy:
Current generated(goa example) code states that it supports SIGINT and SIGTERM for graceful shutdown.

>	// Setup interrupt handler. This optional step configures the process so
>	// that SIGINT and SIGTERM signals cause the services to stop gracefully.

But it actually support only SIGINT, not SIGTERM. 
I think it is better to support SIGTERM for better support of kubernetes or ECS (and so on...) container lifecycle.


raphael:
Makes sense, thank you. Please back-port to v2 as well, thanks again!

PR #2610, opened 2020-07-14
Support GRPC error responses in API

nitinmohan87:


raphael:
Thank you! can you back-port to v2 as well?

PR #2614, opened 2020-07-16
Remove path wildcards in openapi v3

nitinmohan87:
Remove wildcards in path for openapi v3 like we do in openapi v2


raphael:
Thank you! could you back-port to v2 as well? thanks!

PR #2615, opened 2020-07-17
Fix omitempty tag for default

nitinmohan87:


raphael:
Looks great, thank you! please back-port to v2.

PR #2618, opened 2020-07-17
Port to v2: Fix omitempty tag for default

nitinmohan87:


raphael:
The build breaks because somehow the mongodb driver gets pulled as a dependency. We should understand why that is, really wouldn't be great if we had to have that as a dependency...

raphael:
Thank you!

PR #2620, opened 2020-07-22
Follow `errcheck` suggestion

hirosassa:
I found following errors from `golanglint-ci`:

```go
Error return value of `srv.Shutdown` is not checked (errcheck)
		srv.Shutdown(ctx)
Error return value of `w.Write` is not checked (errcheck)
		w.Write([]byte(“[” + id + “] encoding: ” + err.Error()))
```

I fixed the errors by adding "explicit-ignore" code.
Please let me know if I miss anything. 
Thank you.


raphael:
Great, thank you!

PR #2650, opened 2020-08-10
Add error message for JSON unmarshal errors in cli

NeenuAVarghese:
This PR adds logic for returning error along with the error message for JSON Unmarshal errors in CLI. 
Without the added logic, if I run the calc-cli, I get the following error from CLI:
```
$  ./calc-cli --url="grpc://localhost:8080" calc add --message '{"a": a, "b": 2}'
invalid JSON for message, example of valid JSON:
'{
      "a": 8399553735696626949,
      "b": 360622074634248926
   }'
run './calc-cli --help' for detailed usage.
```
With this change, I get to see what the error is with the JSON, which helps me quickly understand what exactly is wrong with the payload:
```
 ./calc-cli --url="grpc://localhost:8080" calc add --message '{"a": "a", "b": 2}'
invalid JSON for message,
error: json: cannot unmarshal string into Go struct field AddRequest.a of type int32,
example of valid JSON:
'{
      "a": 8399553735696626949,
      "b": 360622074634248926
   }'
run './calc-cli --help' for detailed usage.
```
Returning the error in JSON is really helpful in debugging what exactly is wrong with the message especially in case of complex payload.

### DIFF
--- a/codegen/cli/cli.go
+++ b/codegen/cli/cli.go
@@ -198,7 +198,7 @@ func BuildSubcommandData(svcName string, m *service.MethodData, buildFunction *B
 				conversion = "var err error\n" + conversion
 				conversion += "\nif err != nil {\n"
 				if flagType(m.Payload) == "JSON" {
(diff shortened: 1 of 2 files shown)


raphael:
Looks great, thank you!

PR #2666, opened 2020-08-20
Fix gRPC validation when extending types

nitinmohan87:


raphael:
Looks great, thank you! can you backport to v2 as well? thanks!

PR #2698, opened 2020-10-28
Set default value in client request

nitinmohan87:


raphael:
Thank you! looks great.

raphael:
I backported this change to v2: https://github.com/goadesign/goa/pull/2706

PR #2712, opened 2020-11-12
👷  add dependabot.yml

utamori:
Add dependabot,yml. 
This automatically creates a pull request for dependency updates.
I think it will make development a little bit easier.
[About Dependabot version updates - GitHub Docs](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/about-dependabot-version-updates)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


raphael:
Makes sense, thank you! Could we make the check weekly instead of daily? I'll merge once you make that change, thank you again!

utamori:
You're welcome.
I'm glad I was able to contribute in a small way.

raphael:
This is great thank you!

PR #2722, opened 2020-11-21
Fix code generation of http server when an array of query parameters …

ayanamizuta:
PR of issue #2721

PR #2725, opened 2020-11-23
v2: Fix code generation of http server when an array of query parameters exists

ayanamizuta:
v2 ver. of https://github.com/goadesign/goa/pull/2722


raphael:
Perfect, thank you!

Issue #2726, opened 2020-11-24
Implementing exclusiveMaximum and exclusiveMinimum validation Json


selmison:
I noted that Goa hasn't implemented **exclusiveMaximum** and **exclusiveMinimum** validation Json. Is it possible add these validations?

- Add a "exclusiveMaximum" validation to the attribute. See http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.2.3.
- Add a "exclusiveMinimum" validation to the attribute. See http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.2.5

raphael:
Can you elaborate a little bit on why `Maximum` and `Minimum` are not sufficient? these were omitted by design as they seemed redundant given that the design DSL is typically static.

selmison:
For example, a product's price or any numeric value should be greater than zero (0.0...01)

```
var ProductDTO = Type("ProductDTO", func() {
	Description("Product Type")
	Attribute("id", String)
	Attribute("name", string)
	Attribute("price", Float32, func(){
             ExclusiveMinimum(0)
         })
	Required("id", "name", "price")
})
```

raphael:
I see makes sense, there is such a thing as floats! duh :)

PR #2727, opened 2020-11-26
feature: adding ExclusiveMinimum and ExclusiveMaximum

selmison:
PR of issue #2726

### DIFF
--- a/expr/example.go
+++ b/expr/example.go
@@ -132,7 +132,10 @@ func hasMinMaxValidation(a *AttributeExpr) bool {
 	if a.Validation == nil {
 		return false
 	}
-	return a.Validation.Minimum != nil || a.Validation.Maximum != nil
+	return a.Validation.ExclusiveMinimum != nil ||
+		a.Validation.Minimum != nil ||
+		a.Validation.ExclusiveMaximum != nil ||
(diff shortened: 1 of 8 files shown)


expr/example.go
```diff
@@ -236,9 +239,15 @@ func byMinMax(a *AttributeExpr, r *Random) interface{} {
...
+	}
 	if a.Validation.Maximum != nil {
 		max = *a.Validation.Maximum
 	}
+	if a.Validation.ExclusiveMinimum != nil {
+		min = *a.Validation.ExclusiveMinimum
+	}
 	if a.Validation.Minimum != nil {
 		min = *a.Validation.Minimum
 	} else {
```
raphael, line 251: We need to test that there is no exclusive minimum as well here, something like:
```go
} else if a.Validation.ExclusiveMinimum == nil {
```

raphael (changes requested):
Thank you, this is great! Just one issue reported below.

raphael (approved):

raphael:
This is great, thank you!

raphael:
One last request: could you back-port this to the `v2` branch please? Simply cherry-pick the merge commit. Thanks again!

PR #2732, opened 2020-12-01
Add ipv6 support

ksthiele:
Hello,

this small PR makes the generated example more URI agnostic and now accepts IPv6 or future protocols that are supported by the net package.

### DIFF
--- a/codegen/example/example_server.go
+++ b/codegen/example/example_server.go
@@ -36,6 +36,7 @@ func exampleSvrMain(genpkg string, root *expr.RootExpr, svr *expr.ServerExpr) *c
 		{Path: "flag"},
 		{Path: "fmt"},
 		{Path: "log"},
+		{Path: "net"},
 		{Path: "net/url"},
 		{Path: "os"},
 		{Path: "os/signal"},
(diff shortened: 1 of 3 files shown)


codegen/example/example_server.go
```diff
@@ -274,11 +275,15 @@ func main() {
...
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "invalid URL %#v: %s\n", u.Host, err)
+					os.Exit(1)
+				}
+				u.Host = net.JoinHostPort(h, *{{ $u.Transport.Type }}PortF)
 			} else if u.Port() == "" {
-				u.Host += ":{{ $u.Port }}"
-			}
+				u.Host = net.JoinHostPort(u.Host, ":{{ $u.Port }}")
+			}
```
raphael: Nitpick: please remove the extra whitespaces

raphael (commented):
This looks great, thank you! left a small nitpick that would be nice to address.

codegen/example/example_server.go
ksthiele: done

raphael:
Thank you! one last request: could you also back-port this change to the `v2` branch as well please?

PR #2737, opened 2020-12-08
Openapi v3 response fixes

nitinmohan87:
* Sets default response description as we do in openapi v2
* Removes "content" section for empty response body (https://swagger.io/docs/specification/describing-responses/)


http/codegen/openapi/v3/builder_test.go
```diff
@@ -318,12 +318,14 @@ func matchesResponse(t *testing.T, r *ResponseRef, types map[string]*openapi.Sch
 		}
 		matchesHeader(t, h, types, exp)
 	}
-	ct, ok := v.Content["application/json"]
-	if !ok {
-		t.Error("missing response content, expected application/json")
-		return
+	if expected.Type.Type != "" {
```
lisabien, line 321: Just double-checking, is this supposed to be `expected.Type.Type` or just `expected.Type`?

nitinmohan87: Looking at the way that test struct was designed, it should be `expected.Type.Type`. We could refactor the tests to take a struct pointer instead of the actual struct so that we can check it like `expected.Type == nil`.

lisabien: Cool, thanks for the explanation.

raphael:
This looks great, thank you! could you backport this to v2 as well? thanks!

PR #2738, opened 2020-12-09
Add unicode/utf8 import used in validation to grpc cli

nitinmohan87:


raphael:
Looks great, thank you! Could you backport this to v2 as well? thanks!

PR #2748, opened 2020-12-17
Fix grpc streaming client conflicting types

### LINKED ISSUE #2719: Conflicting functions generated by gRPC
Goa generates invalid client code when a single type is used in both a StreamingPayload and a StreamingResult. This reproduces in at least v3.1.1 and v3.2.5. This occurs whether they are in the same Method, or in two different Methods.

```go
package design

var ev = Type("Event", func () {
    Description("event")
})

var _ = Service("hub", func() {
    Method("send", func() {
        StreamingPayload(ev)
        StreamingResult(ev)
        GRPC(func() {})
    })
})
```

The resulting `gen/grpc/hub/client/types.go` contains two `NewEvent` functions. (The ellipses redact irrelevant paths.)

```go
// Code generated by goa v3.2.5, DO NOT EDIT.
//
// hub gRPC client types
//
// Command:
// $ goa gen .../design

package client

import (
	hubpb ".../gen/grpc/hub/pb"
	hub ".../gen/hub"
)

func NewEvent() *hub.Event {
	result := &hub.Event{}
	return result
}

func NewEvent() *hubpb.Event {
	v := &hubpb.Event{}
	return v
}
```

Ideas for workarounds appreciated; I assume the simplest would be to create multiple types with the same layout and ensure each are used only for a single direction.

nitinmohan87:
Fixes #2719

### DIFF
--- a/expr/grpc_endpoint.go
+++ b/expr/grpc_endpoint.go
@@ -319,7 +319,14 @@ func (e *GRPCEndpointExpr) Finalize() {
 
 	// Finalize streaming payload type if defined
 	if e.MethodExpr.StreamingPayload.Type != Empty {
-		initAttrFromDesign(e.StreamingRequest, e.MethodExpr.StreamingPayload)
+		attr := e.MethodExpr.StreamingPayload
+		// If streaming payload is a user type, use the underlying attribute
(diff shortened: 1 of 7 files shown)


raphael:
This is great, thank you!

PR #2754, opened 2021-01-03
Fix validation of UUID

prijip:
Currently the UUID format accepts string like `abcdefgh-ijkl-mnop-qrst-uvqxyz012345` as valid although it contains characters that are not hex digits.

This PR makes is to verify that UUID strings contain hex digits.


raphael:
Thank you for the PR! This looks good. Could I ask you to back-port this change to v2 as well? thanks!

prijip:
Sure, I will add it to v2 as well.

PR #2758, opened 2021-01-10
Fix validation of UUID

prijip:
- Validate that UUID strings contain hex digits
- Use github.com/google/uuid for validating UUID
- More test cases for UUID


raphael:
Thank you!

Issue #2792, opened 2021-03-23
Wrong path to generated protobuf package in imports when generating examples


sablev:
I'm using [grpc_go_1_26](https://github.com/goadesign/goa/tree/grpc_go_1_26) branch.

When I try to generate examples, I get an error:
```console
project/gen/grpc/svc/client imports
        tg-service-goa/gen/grpc/svc/pb: package project/gen/grpc/svc/pb is not in GOROOT (C:\Program Files\Go\src\project\gen\grpc\svc\pb)
```
Cause of this error is that goa generate wrong import path to protobuf package:
```go
svcpb "tg-service-goa/gen/grpc/svc/pb"
```

Correct path:
```go
svcpb "tg-service-goa/gen/grpc/svc/pb/svcpb"
```

My API design definitions:
```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var (
	_ = API("example", func() {
		Title("example")
		Version("1.0")
		Description("example api description")
		Server("example", func() {
			Description("example server description")
			Services("svc")

			Host("development", func() {
				Description("Development host")
				URI("grpc://localhost:50051")
			})
		})
	})

	_ = Service("svc", func() {
		Description("svc desc")

		Method("mtd", func() {
			Payload(MtdReq)

			Result(MtdResp)
			Error("NotFound")
			Error("BadRequest")

			GRPC(func() {
				Response("NotFound", CodeNotFound)
				Response("BadRequest", CodeInvalidArgument)
			})
		})
	})

	MtdReq = Type("MtdRequest", func() {
		Description("MtdRequest desc")
		Attribute("att", ArrayOf(String), "att desc", func() {
			Example([]string{
				"R",
				"T",
			})
			Meta("rpc:tag", "1")
		})
	})

	MtdResp = Type("MtdResponse", func() {
		Description("MtdResponse desc")
		Attribute("att", MapOf(String, Boolean), "att desc", func() {
			Example(map[string]bool{
				"R": true,
				"T": true,
			})
			Meta("rpc:tag", "1")
		})
	})
)
```

raphael:
Thank you for looking into it. I believe this issue has been fixed in the head of the branch. The branch now relies on google.golang.org/protobuf/cmd/protoc-gen-go instead of github.com/golang/protobuf as it is the recommended approach going forward (see https://developers.google.com/protocol-buffers/docs/reference/go/faq). The following upgrades `protoc-gen-go` to the newest version:
```
[...]

PR #2793, opened 2021-03-23
Fix import path to generated protobuf package

sablev:
#2792 

Wrong import path to protobuf package:
`path.Join(genpkg, "grpc", svcName, pbPkgName)`

Replaced by:
`path.Join(genpkg, "grpc", svcName, pbPkgName, svcName + pbPkgName)`


raphael:
Thank you for the PR, as noted in https://github.com/goadesign/goa/issues/2792 I believe the issue can be fixed by upgrading `protoc-gen-go` to google.golang.org/protobuf/cmd/protoc-gen-go. That being said it might be desirable to maintain backwards compatibility and to have Goa adjust the generated code depending on which version of protoc-gen-go is installed. Lots of pros and cons here (complexity of implementing backwards compatibility vs. breaking change).

Also an alternative to what his PR does would be to pass the `-go_opt=paths=source_relative` flag to `protoc` which would make it generate the file in the top level folder.

PR #2797, opened 2021-03-28
Fix handling of server variables

raphael:
in generated OpenAPI v3 specification


raphael:
This PR fixes a long standing issue with the server URL generated in OpenAPI v3 spec being incorrectly expanded.

PR #2799, opened 2021-03-29
Update example in Service.go

colin-gourlay:
I've found that when using `Title` as per the example the following error is generated when I try to compile the DSL:

[design/design.go:55] invalid use of Title in service

Changing to `Description` solves the issue for me (running v3)

Description


raphael:
Thank you, `Title` was used in an older version of the DSL and the comment was never updated.

PR #2817, opened 2021-04-15
Fix custom error display that occurred in the decoder 

lev-ok:
This pull request fixes an issue related to displaying a custom error that is returned by the decoder.

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -503,6 +503,9 @@ func {{ .RequestDecoder }}(mux goahttp.Muxer, decoder func(*http.Request) goahtt
 {{- if .MultipartRequestDecoder }}
 		var payload {{ .Payload.Ref }}
 		if err := decoder(r).Decode(&payload); err != nil {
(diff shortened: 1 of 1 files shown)


raphael:
Thank you for the PR, looks good! Would it be possible to add a test that shows what was broken before and what this fixes?

raphael:
@levkoburburas ping, I'd like to merge this PR but would have liked to see an example of "before" and "after" in the form of a test. Do you think that this is something you could provide? thank you!

lev-ok:
> @levkoburburas ping, I'd like to merge this PR but would have liked to see an example of "before" and "after" in the form of a test. Do you think that this is something you could provide? thank you!

Hi @raphael, I apologize for not answering for so long. Sorry, I'm very, very busy right now and don't have enough time to write unit test for this changes. That's OK if you decide to close this PR.

Thanks for this awesome REST API server!

raphael:
I'm going to close this as I don't quite understand the use case that this addresses.

PR #2823, opened 2021-04-28
Properly handle media types with params by http.ResponseEncoder()

tchssk:
* `http.ResponseEncoder` always returns `json.Encoder` if media type contains a param.
* This pull request fixes it.


raphael:
Whoops, thank you!

PR #2825, opened 2021-05-02
Properly handle media types with params by http.ResponseEncoder()

tchssk:
This is a backporting of #2823 to v2.


raphael:
The `v2` branch is falling a bit behind and at this point it seems the vast majority of Go projects should be using Go modules. I am thinking of deprecating the `v2` branch and stopping to maintain it.

tchssk:
I agree. I'll close this.

PR #2826, opened 2021-05-02
Avoid encoder panic when v is interface{} type and it's value is nil

ikawaha:
When using `Result(Any)`, if the service method returns nil, it will cause a panic. This patch avoids this.

A parameter of type interface{} in an encoder is usually typed nil, but may be untyped nil when using `Any`.

```go
// EncodeDocumentResponse returns an encoder for responses returned by the
// asset document endpoint.
func EncodeDocumentResponse(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
		res := v.(interface{}) // ← ★ panic if v == nil
		enc := encoder(ctx, w)
		body := res
		w.WriteHeader(http.StatusOK)
		return enc.Encode(body)
	}
}
```

After applying this patch, the response will look like this:
```
$ curl -iii -XGET localhost:8080/doc
HTTP/1.1 200 OK
Content-Type: application/json
Date: Sun, 02 May 2021 06:20:24 GMT
Content-Length: 5

null
```

# Example
## Design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = Service("asset", func() {
	Method("document", func() {
		Result(Any)
		HTTP(func() {
			GET("/doc")
			Response(StatusOK)
		})
	})
})
```

## Service Method

```go
package assetapi

import (
	"context"
	"log"

	"calc/gen/asset"
)

// asset service example implementation.
// The example methods log the requests and return zero values.
type assetsrvc struct {
	logger *log.Logger
}

// NewAsset returns the asset service implementation.
func NewAsset(logger *log.Logger) asset.Service {
	return &assetsrvc{logger}
}

// Document implements document.
func (s *assetsrvc) Document(ctx context.Context) (res interface{}, err error) {
	s.logger.Print("asset.document")
	return  // ← ★ return nil
}
```

## Generated code

```go
// EncodeDocumentResponse returns an encoder for responses returned by the
// asset document endpoint.
func EncodeDocumentResponse(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder) func(context.Context, http.ResponseWriter, interface{}) error {
		res := v.(interface{}) // ← ★ panic if v == nil
		enc := encoder(ctx, w)
		body := res
		w.WriteHeader(http.StatusOK)
		return enc.Encode(body)
	}
}
```

## Error

> curl -iii -XGET localhost:8080/doc
> curl: (52) Empty reply from server
```
14:06 $ go run ./cmd/asset --http-port 8080
[assetapi] 14:06:34 HTTP "Document" mounted on GET /doc
[assetapi] 14:06:34 HTTP server listening on "localhost:8080"
[assetapi] 14:06:54 id=rxsfd5PT req=GET /doc from=127.0.0.1
[assetapi] 14:06:54 asset.document
2021/05/02 14:06:54 http: panic serving 127.0.0.1:51358: interface conversion: interface is nil, not interface {}
goroutine 24 [running]:
net/http.(*conn).serve.func1(0xc0000cb4a0)
	/usr/local/Cellar/go/1.16.3/libexec/src/net/http/server.go:1824 +0x153
panic(0x12febe0, 0xc0001642a0)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:971 +0x499
calc/gen/http/asset/server.EncodeDocumentResponse.func1(0x13b9380, 0xc000164270, 0x13b8920, 0xc0000c4460, 0x0, 0x0, 0x0, 0x0)
	/Users/ikawaha/go/src/github.com/ikawaha/calc/gen/http/asset/server/encode_decode.go:21 +0xec
... snip
```

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -1122,7 +1122,7 @@ func {{ .ResponseEncoder }}(encoder func(context.Context, http.ResponseWriter) g
 				w.Header().Set("goa-view", res.View)
 			{{- end }}
 		{{- else }}
-			res := v.({{ .Result.Ref }})
+			res, _ := v.({{ .Result.Ref }})
 		{{- end }}
 		{{- range .Result.Responses }}
 			{{- if .ContentType }}
--- a/http/codegen/server_encode_test.go
+++ b/http/codegen/server_encode_test.go
@@ -59,6 +59,7 @@ func TestEncode(t *testing.T) {
 		{"body-array-user", testdata.ResultBodyArrayUserDSL, testdata.ResultBodyArrayUserEncodeCode},
 		{"body-primitive-string", testdata.ResultBodyPrimitiveStringDSL, testdata.ResultBodyPrimitiveStringEncodeCode},
 		{"body-primitive-bool", testdata.ResultBodyPrimitiveBoolDSL, testdata.ResultBodyPrimitiveBoolEncodeCode},
+		{"body-primitive-any", testdata.ResultBodyPrimitiveAnyDSL, testdata.ResultBodyPrimitiveAnyEncodeCode},
 		{"body-primitive-array-string", testdata.ResultBodyPrimitiveArrayStringDSL, testdata.ResultBodyPrimitiveArrayStringEncodeCode},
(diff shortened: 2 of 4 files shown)


raphael:
Thank you for the great explanation and the fix, looks great.

PR #2829, opened 2021-05-03
Support Meta "struct:field:type" in gRPC

MichaelUrman:
This addresses what appears to be a real bug: if you annotate a field (either streaming or non) used for gRPC, the protobuf type transformation code in convertType failed to respect Meta `struct:field:type`, resulting in the lack of type casts between the requested type and the protobuf mapping thereof, and code that doesn't compile.

It turns out that Meta `struct:field:type` is probably closer to what I wanted for #2498. The implications deserve more thought than I've given them. For example, does a shared design work cleanly across both a client and a server? Do the requested types need to be available to both, or is there a way for one side to use the default type? (Perhaps this was already considered before `struct:field:type` was introduced.)

At a raw implementation level, I'm also unclear how this approach handles the imports on the generated files. It seemed to in my superficial testing, so I'm hoping it's magic I'm not familiar with.

### DIFF
--- a/grpc/codegen/protobuf_transform.go
+++ b/grpc/codegen/protobuf_transform.go
@@ -465,13 +465,24 @@ func convertType(source, target *expr.AttributeExpr, sourceVar string, ta *trans
 		return fmt.Sprintf("%s(%s)", transformHelperName(source, target, ta), sourceVar)
 	}
 
+	sourceType, _ := codegen.GetMetaType(source)
+	targetType, _ := codegen.GetMetaType(target)
 	if source.Type.Kind() != expr.IntKind && source.Type.Kind() != expr.UIntKind {
+		if sourceType != "" || targetType != "" {
+			if ta.proto || targetType == "" {
+				targetType = protoBufNativeGoTypeName(target.Type)
+			}
+			return fmt.Sprintf("%s(%s)", targetType, sourceVar)
+		}
 		return sourceVar
 	}
 	if ta.proto {
-		return fmt.Sprintf("%s(%s)", protoBufNativeGoTypeName(source.Type), sourceVar)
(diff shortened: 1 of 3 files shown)


grpc/codegen/protobuf_transform.go
```diff
@@ -465,13 +465,24 @@ func convertType(source, target *expr.AttributeExpr, sourceVar string, ta *trans
 		return fmt.Sprintf("%s(%s)", transformHelperName(source, target, ta), sourceVar)
 	}
 
+	sourceType, _ := codegen.GetMetaType(source)
```
raphael, line 468: Would it make sense to move the initialization of `sourceType` under the `if` below?

raphael:
This looks great, thank you for the PR. Left a comment with a nitpick. I am not sure I follow the questions above wrt sharing such type. As far as generating the import this is done by invoking [AddServiceDataMetaTypeImports](https://github.com/goadesign/goa/blob/20b02e80edc370f29814f129966e0015800146fa/codegen/service/service.go#L154). This function should already be called in all the places where the import is needed.

MichaelUrman:
Regarding the nit: yes, it could be moved. A prior state leveraged sourceType in the condition. But that condition was incorrect, and I missed that potential cleanup upon later changes.

On sharing the type, if client and server both leverage goa-generated clients, does the server wanting to use a SpecializedID instead of an Int64 mean the client also has to use a SpecializedID, thus SpecializedID shouldn't be in an internal package? Or can a client easily disregard this and use int64 if it pleases?

(My motivating use case is to use go's type system to disambiguate various Int64 IDs. My client leverages the proto directly&mdash;it's not even go, much less goa&mdash;so it doesn't care about go types.)

PR #2830, opened 2021-05-05
Add a DSL Redirect

tchssk:
This pull request adds a new DSL to define a HTTP redirect.

The DSL can be used like below:

```go
	Method("method", func() {
		HTTP(func() {
			GET("/resources")
			Redirect("/redirect/dest", StatusMovedPermanently)
		})
	})
})

	Files("/file.json", "/path/to/file.json", func() {
		Redirect("/redirect/dest", StatusMovedPermanently)
	})
})
```

The generated server handler initalizer for the HTTP endpoint redirect is:

```go
func NewMethodHandler(
	endpoint goa.Endpoint,
	mux goahttp.Muxer,
	decoder func(*http.Request) goahttp.Decoder,
	encoder func(context.Context, http.ResponseWriter) goahttp.Encoder,
	errhandler func(context.Context, http.ResponseWriter, error),
	formatter func(err error) goahttp.Statuser,
) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		ctx := context.WithValue(r.Context(), goahttp.AcceptTypeKey, r.Header.Get("Accept"))
		ctx = context.WithValue(ctx, goa.MethodKey, "Method")
		ctx = context.WithValue(ctx, goa.ServiceKey, "Service")
		http.Redirect(w, r, "/redirect/dest", http.StatusMovedPermanently)
	})
}
```

and the generated server mounter for the file server redirect is:

```go
func Mount(mux goahttp.Muxer) {
	MountPathToFileJSON(mux, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		http.Redirect(w, r, "/redirect/dest", http.StatusMovedPermanently)
	}))
}
```

### DIFF
--- a/dsl/http_redirect.go
+++ b/dsl/http_redirect.go
@@ -1,0 +1,56 @@
+package dsl
+
+import (
+	"goa.design/goa/v3/eval"
+	"goa.design/goa/v3/expr"
+)
+
+// Redirect indicates that HTTP requests reply to the request with a redirect.
+// The logic is the same as the standard http package Redirect function.
+//
+// Redirect must appear in a HTTP endpoint expression or a HTTP file server
+// expression.
+//
+// Redirect accepts 2 arguments and an optional DSL. The first argument is the
+// url that redirected to. The second argument is the HTTP status code.
+//
+// Example:
+//
+//    var _ = Service("service", func() {
+//        Method("method", func() {
+//            HTTP(func() {
+//                GET("/resources")
+//                Redirect("/redirect/dest", StatusMovedPermanently)
+//            })
+//        })
+//    })
+//
+//    var _ = Service("service", func() {
+//        Files("/file.json", "/path/to/file.json", func() {
+//            Redirect("/redirect/dest", StatusMovedPermanently)
+//        })
+//    })
+//
+func Redirect(url string, code int, fns ...func()) {
+	if len(fns) > 1 {
+		eval.ReportError("too many arguments given to Redirect")
+		return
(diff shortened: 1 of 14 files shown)


dsl/http_redirect.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Redirect indicates that HTTP requests reply to the request with a redirect.
+// The logic is the same as the standard http package Redirect function.
+//
+// Redirect must appear in a HTTP endpoint expression or a HTTP file server
+// expression.
+//
+// Redirect accepts 2 arguments and an optional DSL. The first argument is the
+// url that redirected to. The second argument is the HTTP status code.
```
raphael: URL that is being redirected to.

raphael (commented):
This looks great, thank you for the PR! Just one question: it seems the DSL func argument to `Redirect` is only there to allow for using `Meta`. And it doesn't like any metadata attached to redirects would be used anywhere at this point. Would it make sense to simplify and remove the DSL argument and the `Meta` field or do you have a use for it?

dsl/http_redirect.go
tchssk: I added `Meta` with reference to other existing DSLs and expressions, but it was not needed.

* d94aeee disables `Meta` for `Redirect`.
* 76d7dda fixes godocs.

raphael: Looks great, thank you for the changes!

PR #2849, opened 2021-05-17
Disallow response body for HEAD method

nitinmohan87:


raphael:
This is great! thank you.

PR #2851, opened 2021-05-18
Fix code gen for empty error response bodies

nitinmohan87:


http/codegen/server.go
```diff
@@ -1324,7 +1324,7 @@ const responseT = `{{ define "response" -}}
 	{{- end }}
 
 	{{- if .ErrorHeader }}
-	w.Header().Set("goa-error", {{ printf "%q" .ErrorHeader }})
+	w.Header().Set("goa-error", res.ErrorName())
```
nitinmohan87, line 1327: I had to make this change so that we always use this `res` variable that we initialize in the error encoder. There could be a situation where the `res` variable is initialized and never used causing compile issues.

codegen/funcs.go
```diff
@@ -258,7 +258,7 @@ func WrapText(text string, maxChars int) string {
 
 // InitStructFields produces Go code to initialize a struct and its fields from
 // the given init arguments.
-func InitStructFields(args []*InitArgData, name, targetVar, sourcePkg, targetPkg string, mustInit bool) (string, []*TransformFunctionData, error) {
```
nitinmohan87, line 261: These args were never used in the function :facepalm:

raphael:
This is great, thank you!

PR #2853, opened 2021-05-21
Always encode all result attributes

nitinmohan87:
This PR changes the HTTP encoding of responses to guarantee that all the result type attributes are transmitted. It does that by writing any attribute not already written to the response body, headers or cookies to special headers prefixed with `Goa-`. This is especially convenient for transmitting error details back to requests using HTTP verbs that do not permit writing a response body (HEAD).

For example:

```go
var Result = Type("Result", func() {
    Attribute("foo")
    Attribute("bar")
})

    Method("head", func() {
        Result(Result)
    })
    HTTP(func() {
        HEAD("/")
        Response(StatusOK, func() {
            Body(Empty) // Body is empty, attributes "foo" and "bar" will be transmitted via HTTP headers,
        })
    })
})
```

Note that it is possible to avoid having all the result type attributes being transmitted by using views (e.g. defining a default view explicitly that omits certain attributes.). This is especially useful when using `Tag` to define multiple possible response status code (the tagged attribute used to select the status code can be omitted from the response).


raphael:
This is great, thank you!

PR #2857, opened 2021-05-25
Map unmapped attrs ErrorResult

nitinmohan87:
* Only map attributes that are unmapped if type is ErrorResult (Undoes many of the changes in #2853 )


raphael:
Thank you, looks great!

PR #2858, opened 2021-05-26
Allow to specify examples in OpenAPI when skip encode decode body set

nitinmohan87:
Allows to specify examples for HTTP endpoints with skip request/response encode/decode body as below
```
                Method("test", func() {
                        Payload(Any, func() {
                               Example("my example")
                        })
                        Result(Any, func() {
                                Example("my result")
                        })
			HTTP(func() {
				POST("/")
				SkipRequestBodyEncodeDecode()
				SkipResponseBodyEncodeDecode()
			})
		})
```


http/codegen/openapi/v3/types.go
```diff
@@ -74,6 +74,12 @@ func buildBodyTypes(api *expr.APIExpr) (map[string]map[string]*EndpointBodies, m
 			}
 
 			req := sf.schemafy(e.Body)
+			if e.SkipRequestBodyEncodeDecode {
+				// it is possible for design to have request Body as Empty and set
+				// SkipRequestBodyEncodeDecode. In this case, we set the request body
+				// example from the Payload.
+				req = sf.schemafy(e.MethodExpr.Payload)
+			}
```
raphael, line 82: Should the code remove any header, cookie or param from the payload for the body example?

http/codegen/openapi/v3/types.go
```diff
@@ -115,6 +121,12 @@ func buildBodyTypes(api *expr.APIExpr) (map[string]map[string]*EndpointBodies, m
 					body.Type = rt
 				}
 				js := sf.schemafy(body)
+				if e.SkipResponseBodyEncodeDecode {
+					// it is possible for design to have request Body as Empty and set
+					// SkipRequestBodyEncodeDecode. In this case, we set the request body
+					// example from the Payload.
+					js = sf.schemafy(e.MethodExpr.Result)
```
raphael, line 128: Sames as above

codegen/service/example_svc.go
```diff
@@ -106,10 +106,10 @@ func basicEndpointSection(m *expr.MethodExpr, svcData *Data) *codegen.SectionTem
 		MethodData:     md,
 		ServiceVarName: svcData.VarName,
 	}
-	if m.Payload.Type != expr.Empty {
+	if md.Payload != "" {
```
raphael, line 109: Can you explain in which cases these two tests are different?

codegen/service/service_data.go
```diff
@@ -785,6 +785,25 @@ func buildMethodData(m *expr.MethodExpr, svcPkgName string, service *expr.Servic
...
+			// if endpoint defines SkipXX and only a body then we ignore the payload
+			// and use it only to generate examples.
+			payloadName = ""
+			payloadRef = ""
+		}
+		bodyOnly := true
+		for _, r := range httpMet.Responses {
+			if !r.BodyOnly() {
+				// one of the response defines header/cookie.
+				bodyOnly = false
```
raphael: could add a `break` here

http/codegen/openapi/v3/builder.go
```diff
@@ -210,21 +210,37 @@ func buildOperation(key string, r *expr.RouteExpr, bodies *EndpointBodies, rand
...
+				Example: e.Body.Example(rand),
+			}
+			requestBody = &RequestBodyRef{Value: &RequestBody{
+				Description: e.Body.Description,
+				Required:    e.Body.Type != expr.Empty,
+				Content:     map[string]*MediaType{ct: mt},
+				Extensions:  openapi.ExtensionsFromExpr(e.Body.Meta),
+			}}
+		} else if e.SkipRequestBodyEncodeDecode {
+			// it is possible for design to have request Body as Empty and set
```
raphael: I don't understand this case. If SkipXXX is used then it seems that the OpenAPI request body schema should be computed by looking up all the attributes of the payload that are not used in headers, cookies and params? If the body type is set to empty then maybe we should honor that in the OpenAPI schema?

http/codegen/openapi/v3/response.go
```diff
@@ -43,6 +43,19 @@ func responseFromExpr(r *expr.HTTPResponseExpr, bodies map[int][]*openapi.Schema
 				Example:    r.Body.Example(rand),
 				Extensions: openapi.ExtensionsFromExpr(r.Body.Meta),
 			}
+		} else {
+			e := r.Parent.(*expr.HTTPEndpointExpr)
+			if e.SkipResponseBodyEncodeDecode {
+				// it is possible for design to have response Body as Empty and set
+				// SkipResponseBodyEncodeDecode. In this case, we set the response body
+				// example from the Result.
+				content = make(map[string]*MediaType)
```
raphael: Same as above

http/codegen/testdata/openapi_dsls.go
```diff
@@ -314,6 +314,24 @@ var ExtensionDSL = func() {
...
+				Example(`{"foo":"bar"}`)
+			})
+			HTTP(func() {
+				POST("/")
+				SkipRequestBodyEncodeDecode()
+				SkipResponseBodyEncodeDecode()
+			})
+		})
+	})
+}
```
raphael, line 333: Do you think we also need to test for the cases where the body has multiple attributes and some are used by other elements than the body (e.g. headers, params etc.)?

raphael (commented):
Looks good, thank you - I left a few comments and questions

codegen/service/example_svc.go
nitinmohan87: Method Payload can be Empty when SkipRequestEncodeDecode is set. In that case, we should generate the example service correctly without the payload argument.

raphael: That makes sense, thank you. Would it make sense/be clearer to use `mustInitPayload` instead?

codegen/service/service_data.go
```diff
@@ -816,6 +816,40 @@ func buildMethodData(m *expr.MethodExpr, svcPkgName string, service *expr.Servic
...
+	if m.Result.Type == expr.Empty {
+		return false
+	}
+
+	var httpMet *expr.HTTPEndpointExpr
+	if httpSvc := expr.Root.HTTPService(m.Service.Name); httpSvc != nil {
+		httpMet = httpSvc.Endpoint(m.Name)
+		for _, r := range httpMet.Responses {
+			if httpMet.SkipResponseBodyEncodeDecode && r.HasBodyOnly() {
+				return false
```
raphael, line 846: Should this function only return `false` if *all* responses only define a body?

expr/http_endpoint.go
```diff
@@ -423,8 +391,6 @@ func (e *HTTPEndpointExpr) Validate() error {
 				if r.Body != nil && r.Body.Type == Empty {
 					verr.Add(r, "Response body empty but endpoint defines streaming WebSocket response.")
 				}
-			} else if successResp && e.SkipResponseBodyEncodeDecode {
```
raphael, line 426: The idea with this test was that it's not possible for Goa to know what response status to generate if there is no result to the service level function (and therefore no field to compare with the `Tag` value defined in the design). So maybe the test should also make sure that there is only a body defined (in which case there won't be a result) but in that case it seems to be still needed?

expr/http_endpoint.go
```diff
@@ -846,6 +810,72 @@ func (e *HTTPEndpointExpr) validateHeadersAndCookies() *eval.ValidationErrors {
...
+	}
+
+	successResp := false
+	for _, r := range e.Responses {
+		if r.Body != nil {
+			verr.Add(r, "Cannot define a response body when endpoint uses SkipResponseBodyEncodeDecode.")
+		}
+		// with SkipResponseBodyEncodeDecode at most one success response can be defined
+		if r.StatusCode < 400 {
+			if successResp && e.SkipResponseBodyEncodeDecode {
```
raphael, line 868: This a pre-existing bug but it seems the test should also make sure that there is only a body defined (in which case there won't be a result and thus no way for Goa to know what status code to use because no field to test for the `Tag` value). If there are headers defined then I believe it's OK to have multiple responses.

expr/http_endpoint.go
```diff
@@ -536,9 +499,6 @@ func (e *HTTPEndpointExpr) Validate() error {
 				verr.Add(e, "Payload type is array but HTTP endpoint defines both a body and headers. At most one of these must be defined and it must be an array.")
 			}
 		}
-		if !hasParams && !hasHeaders && e.SkipRequestBodyEncodeDecode {
```
raphael, line 539: I don't see this case being covered in the `SkipXXX` validation functions.

expr/http_endpoint.go
```diff
@@ -573,9 +533,6 @@ func (e *HTTPEndpointExpr) Validate() error {
 				verr.Add(e, "Payload type is map but HTTP endpoint defines both a body and route or query string parameters. At most one of these must be defined and it must be a map.")
 			}
 		}
-		if !hasParams && e.SkipRequestBodyEncodeDecode {
-			verr.Add(e, "Payload type is map but HTTP endpoint uses SkipRequestBodyEncodeDecode and does not define headers.")
```
raphael, line 577: Same as above, I don't see this validation being tested below.

expr/http_endpoint.go
```diff
@@ -609,9 +566,6 @@ func (e *HTTPEndpointExpr) Validate() error {
 	}
 
 	body := httpRequestBody(e)
-	if e.SkipRequestBodyEncodeDecode && body.Type != Empty {
```
raphael, line 612: Same as above

http/codegen/service_data.go
```diff
@@ -1400,7 +1400,16 @@ func buildResultData(e *expr.HTTPEndpointExpr, sd *ServiceData) *ResultData {
...
+		hasResult := result.Type != expr.Empty
+		if hasResult {
+			for _, r := range e.Responses {
+				if e.SkipResponseBodyEncodeDecode && r.HasBodyOnly() {
+					hasResult = false
+					break
+				}
+			}
+		}
+		if hasResult {
```
raphael, line 1412: Could that just be in the same block as above?

http/codegen/service_data.go
```diff
@@ -1539,7 +1548,7 @@ func buildResponses(e *expr.HTTPEndpointExpr, result *expr.AttributeExpr, viewed
 						break
 					}
 				}
-				if needInit(result.Type) {
+				if needInit(result.Type) && !(e.SkipResponseBodyEncodeDecode && resp.HasBodyOnly()) {
```
raphael, line 1551: Are there cases where we call `needInit` without testing for `SkipResponseBodyEncodeDecode`? If not maybe `needInit` could be refactored to do that test as well?

codegen/service/service_data.go
nitinmohan87: With SkipResponseXXX there can be only be one successful response. Otherwise validation would fail https://github.com/goadesign/goa/blob/v3/expr/http_endpoint.go#L426-L428

PR #2859, opened 2021-05-27
Validate result types received via client grpc stream

nitinmohan87:
This PR fixes an issue where the generated gRPC client would not run validations on result payloads.


raphael:
Thank you!

PR #2863, opened 2021-06-05
Stop processing if a type is specified for 'Error' DSL in API

ikawaha:
issue: https://github.com/goadesign/goa/issues/2867 case 1

API level Error can be defined only with default ErrorResult.
(because result types are generated for each service but API level Error is not inherited to services)
This PR adds a validation to not allow Error with custom result type.

### Design
```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = API("calc", func() {
	Error("DivByZero", String)
	HTTP(func() {
		Response("DivByZero", StatusInternalServerError)
	})
})
```

### goa gen
```
$ goa gen calc/design
exit status 1
[design/design.go:8] error type cannot be specified in API calc
```

### DIFF
--- a/dsl/error.go
+++ b/dsl/error.go
@@ -46,6 +46,9 @@ func Error(name string, args ...interface{}) {
 	erro := &expr.ErrorExpr{AttributeExpr: att, Name: name}
 	switch actual := eval.Current().(type) {
 	case *expr.APIExpr:
+		if erro.Type != expr.ErrorResult {
+			eval.ReportError("error type cannot be specified")
+		}
 		expr.Root.Errors = append(expr.Root.Errors, erro)
 	case *expr.ServiceExpr:
 		actual.Errors = append(actual.Errors, erro)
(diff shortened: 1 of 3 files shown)


raphael:
Thank you for the PR! I am not sure I understand why API level error definition couldn't use custom types? Clearly there is a bug as it doesn't work today but it seems like that should be possible? The reference at the `Service` or `Method` level should inherit from the API level type unless they override it.

```go
var _ = API("my_api", func() {
    Error("Custom", String)
    HTTP(func() {
        Response("Custom", StatusInternalError)
    })
    GRPC(func() {
        Response("Custom", CodeInternal)
    })
})

    Method("meth", func() {
        Error("Custom") // type is String
        HTTP(func() { GET("/") }) // Response status code for error "Custom" is StatusInternalError
        GRPC(func() { }) // Response code for error "Custom" is CodeInternal
    })
})

var _ = Service("service2", func() {
    Error("Custom", ErrorResult) // Overrides type defined at the API level
   // ...
})
```

I could be missing something...

ikawaha:
API level errors are not inherited automatically, but can be inherited if explicitly stated in a method or service.

So it seems that there is a bug in the current behavior.

The premise was wrong, I'm going to close this PR.
Thanks for the review!

PR #2866, opened 2021-06-05
Stop processing if the Error definitions in service and method are inconsistent

ikawaha:
issue: https://github.com/goadesign/goa/issues/2867 case 3

When Errors of the same name with different ResultType are defined in Service and Method, only the error defined as Service will be valid.

### design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

	Error("DivByZero", String)  // ← ★
	HTTP(func() {
		Response("DivByZero", StatusInternalServerError)
	})
	Method("div", func() {
		Error("DivByZero", Int) // ← ★
		Payload(func() {
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/div/{a}/{b}")
			Response("DivByZero", StatusBadRequest)
		})
	})
})
```

### goa gen

```go
$ goa gen calc/design
exit status 1
[design/design.go:13] inconsistent definition in service and method: DivByZero in service "calc" method "div"
```

### DIFF
--- a/dsl/error.go
+++ b/dsl/error.go
@@ -50,6 +50,11 @@ func Error(name string, args ...interface{}) {
 	case *expr.ServiceExpr:
 		actual.Errors = append(actual.Errors, erro)
 	case *expr.MethodExpr:
+		for _, v := range actual.Service.Errors {
+			if v.Name == erro.Name && v.Type != erro.Type {
+				eval.ReportError("inconsistent definition in service and method: %s", v.Name)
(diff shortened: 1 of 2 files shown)


raphael:
Thank you for the PR! This fixes the issue but I'm wondering if we wouldn't want the method level error definition override the service level method definition? this would be more consistent with for example how security works where methods can override the default specified at the service level.

PR #2868, opened 2021-06-07
Remove the extra colon in port parameters

ikawaha:
Do not add a prefix colon to the port specification in `net.JoinHostPort()`.

https://play.golang.org/p/RkH9iV-OWgF

```go
package main

import (
	"fmt"
	"net"
)

func main() {
	fmt.Printf("host:port = %s\n", net.JoinHostPort("example.com", ":80"))
}
```

output
```console
host:port = example.com::80
```


raphael:
Whoops, thank you!

Issue #2867, opened 2021-06-05
The `Error` DSL has some problems


ikawaha:
The `Error` DSL seems to have a few problems.

# TL;DR;

1. API level `Error` generates invalid code if `ResultType` is set.
    - only references are allowed(https://github.com/goadesign/goa/pull/2555).
2. If error responses with the same name but different responses are specified in `Service` and `Method`, invalid code will be generated.
3. When `Error`s of the same name with different `ResultType` are defined in Service and Method, only the error defined as Service will be valid.

# Case 1

API level `Error` generates invalid code if `ResultType` is set.

-> PR:https://github.com/goadesign/goa/pull/2863

## Design example

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = API("calc", func() {
	Error("DivByZero", String)
	HTTP(func() {
		Response("DivByZero", StatusBadRequest)
	})
})

var _ = Service("calc", func() {
	Method("div", func() {
		Error("DivByZero")
		Payload(func() {
			Field(1, "a", Int, "Left operand")
			Field(2, "b", Int, "Right operand")
			Required("a", "b")
		})
		Result(Int)
		HTTP(func() {
			GET("/div/{a}/{b}")
		})
	})
})
```

## Generated code

```go: gen/http/server/encode_decode.go
func EncodeDivError(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder, formatter func(err error) goahttp.Statuser) func(context.Context, http.ResponseWriter, error) error {
	encodeError := goahttp.ErrorEncoder(encoder, formatter)
	return func(ctx context.Context, w http.ResponseWriter, v error) error {
		en, ok := v.(ErrorNamer)
		if !ok {
			return encodeError(ctx, w, v)
		}
		switch en.ErrorName() {
		case "DivByZero":
			res := v.(string)        // ← Impossible type assertion: `string` does not implement 'error'
			enc := encoder(ctx, w)
			body := res
			w.Header().Set("goa-error", res.ErrorName())
			w.WriteHeader(http.StatusBadRequest)
			return enc.Encode(body)
		default:
			return encodeError(ctx, w, v)
		}
	}
}

```

# Case 2

If error responses with the same name but different responses are specified in `Service` and `Method`, invalid code will be generated.

-> PR:https://github.com/goadesign/goa/pull/2865

## Design example

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = Service("calc", func() {
[...]

PR #2869, opened 2021-06-09
Inherit API level errors to each service

tchssk:
Related to #2555 #2867 

I think it's useful to be able to inherit API level errors to each service.
This pull request makes it possible.

This is a part of improving Error DSL.
I'll open new ones to improve codegen of http/grpc.


raphael:
Thank you for the PR, the idea with API level errors was that they could be referenced but not automatically inherited. So you can define transport level mappings at that level and not have to reproduce them in each service, for example:

```go
var _ = API("my_api", func() {
    Error("Custom")
    HTTP(func() {
        Response("Custom", StatusInternalError)
    })
    GRPC(func() {
        Response("Custom", CodeInvalid)
    })
})

    Method("meth", func() {
        Error("Custom")
        HTTP(func() {
            GET("/")
            // No need to define response for error "Custom"
        })
        GRPC(func() {
            // No need to define response for error "Custom"
        })
    })
    Method("meth2", func() {
        HTTP(func() {
            GET("/meth2")
            // Does not return error "Custom"
        })
        GRPC(func() {
            // Does not return error "Custom"
        })
    })
})
 
var _ = Service("service2", func() {
    Error("Custom")
     // No need to define response for error "Custom"
     // ....
})
```

I don't think we can change this behavior because it would break existing designs.

tchssk:
Oh I understand it.

> I don't think we can change this behavior because it would break existing designs.

You mean if we use the automatic inheritance, errors will be inherited to also unwanted services, don't you?

It may be possible to realize it by extending `Meta` DSL like below:

```go
var _ = API("my_api", func() {
	Error("Custom", func() {
		Meta("error:inherit:all", true)  // This indicates the error will be inherited to all services.
	})
	HTTP(func() {
		Response("Custom", StatusInternalError)
	})
	GRPC(func() {
		Response("Custom", CodeInvalid)
	})
})

	Method("meth", func() {
		// No need to define error "Custom"
		HTTP(func() {
			GET("/")
			// No need to define response for error "Custom"
		})
		GRPC(func() {
			// No need to define response for error "Custom"
		})
	})
	Method("meth2", func() {
		HTTP(func() {
			GET("/meth2")
			// Does not return error "Custom"
		})
		GRPC(func() {
			// Does not return error "Custom"
		})
	})
})

var _ = Service("service2", func() {
	// No need to define error "Custom"
	// No need to define response for error "Custom"
	// ....
})
```

or using `Services` DSL:

```go
var _ = API("my_api", func() {
	Error("Custom", func() {
		Services("service1", "service2")  // This indicates the error will be inherited to service1 and service2.
	})
	// ...
})
```

raphael:
Yes that's exactly what I meant. The `Services` DSL is a great idea! It would be consistent with the use in `Server`.

tchssk:
The Services DSL was not bad idea but it has no new feature, just another syntax to behaves like the current API Error.

I'll close this. Thank you for a discussion.

PR #2875, opened 2021-06-18
Use formatting parameter instead of embedding var

MichaelUrman:
While unlikely, os.Args[0] could contain % characters, resulting in
wonky formatting to the usage message. This isn't particularly important
in and of itself, but it created noise in some heuristic linting.


raphael:
This is great, thanks for the cleanup!

PR #2839, opened 2021-05-14
Support for http.FileSystem in HTTP file servers

tchssk:
This pull request makes it possible to change the file system of a HTTP file server.

It modifies a server to use `http.FileServer()` instead of `http.ServeFile()` , and change the server initializer to receive a file system.

```go
func New(
	e *servicefileserver.Endpoints,
	mux goahttp.Muxer,
	decoder func(*http.Request) goahttp.Decoder,
	encoder func(context.Context, http.ResponseWriter) goahttp.Encoder,
	errhandler func(context.Context, http.ResponseWriter, error),
	formatter func(err error) goahttp.Statuser,
	fileSystemPathToFile1JSON http.FileSystem,
) *Server {
	if fileSystemPathToFile1JSON == nil {
		fileSystemPathToFile1JSON = http.Dir(".")
	}
	return &Server{
		Mounts: []*MountPoint{
			{"/path/to/file1.json", "GET", "/server_file_server/file1.json"},
		},
		PathToFile1JSON: http.FileServer(fileSystemPathToFile1JSON),
	}
}
```

This and #2838 have same purpose but different approach. #2338 behaves based on `design` but this is not affected by `design` and more simple.

PR #2892, opened 2021-07-23
Seems like the Mount api has changed?

TechplexEngineer:


raphael:
Yes it did in https://github.com/goadesign/goa/pull/2839, great catch!

PR #2910, opened 2021-08-17
Fix validate method of ResultType with Type Enum

### LINKED ISSUE #2907: Error when use Enum Type in ResultType
I write as following

```go
var _ = Service("test", func() {
	Method("test", func() {
		Result(TestResult)
		HTTP(func() {
			GET("/")
			Response(StatusOK)
		})
	})
})

var TestResult = ResultType("test_result", func() {
	Attributes(func() {
		Attribute("test", ArrayOf(TestEnum))
	})
})

var TestEnum = Type("test_enum", String, func() {
	Enum("test1", "test2")
})
```

And, I got the following error.

```
❯ go test ./...
# github.com/kitagry/goa-err/gen/test/views
gen/test/views/view.go:65:12: invalid operation: result != nil (mismatched types
 TestEnumView and nil)
gen/test/views/view.go:66:8: invalid indirect of result (type TestEnumView)
gen/test/views/view.go:67:67: invalid indirect of result (type TestEnumView)
```

generated `view.go` is 

```go
// TestEnumView is a type that runs validations on a projected type.
type TestEnumView string

// ValidateTestEnumView runs the validations defined on TestEnumView.
func ValidateTestEnumView(result TestEnumView) (err error) {
	if result != nil {
		if !(*result == "test1" || *result == "test2") {
			err = goa.MergeErrors(err, goa.InvalidEnumValueError("result", *result, []interface{}{"test1", "test2"}))
		}
	}
	return
}
```

Thank you!

kitagry:
fix #2907

### DIFF
--- a/codegen/service/service_data.go
+++ b/codegen/service/service_data.go
@@ -619,9 +619,7 @@ func (d ServicesData) analyze(service *expr.ServiceExpr) *Data {
 		}
 	}
 
-	var (
-		desc string
-	)
+	var desc string
 	{
 		desc = service.Description
 		if desc == "" {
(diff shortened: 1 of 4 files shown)


raphael:
Whoops I just fixed it as well :) I used a different approach though and didn't have tests. If you want to create a PR with the test you wrote that would be great!

PR #2911, opened 2021-08-17
Add test for #2907

kitagry:
relate #2907


raphael:
This is great, thank you!

PR #2921, opened 2021-09-13
Add . to tmpdir for loading correct module path

### LINKED ISSUE #2918: Goa gen error
Received the following error when we run `goa gen` 
`go get: malformed module path "goa865448624": missing dot in first path element`. Seems like this has got to do with the release of v3.5.1

nitinmohan87:
Fixes #2918


raphael:
Thank you!

PR #2936, opened 2021-10-14
Not skip tests for yaml files of openapi v2

tchssk:
The tests for yaml files of openapi v2 were disabled by 168bb0926deb9982367cf2ab5329059ac157607a but I think they should be enabled.
They were caused by [`validateSwagger()`](https://github.com/goadesign/goa/blob/1f18b5aac2d9ce6676b932e12ded288987e8f56f/http/codegen/openapi/v2/files_test.go#L259-L268) so I changed to only disable the function against yaml.


raphael:
That's great, thank you!

PR #2940, opened 2021-10-19
Simplify examples in testdata of openapi

tchssk:
I'm improving `http/codegen/openapi`.
This is for preparation for that.


raphael:
Great, thanks

PR #2941, opened 2021-10-20
More detailed additionalProperties for openapi dictionaries

tchssk:
Goa is currently generating an OpenAPI dictionary as a free-form object.
This patch improves the code generator to generate more detailed `additionalProperties`.

https://swagger.io/docs/specification/data-models/dictionaries/
https://swagger.io/specification/v2/


raphael:
This is great! thank you.

PR #2952, opened 2021-11-24
Return encode decode errors grpc

nitinmohan87:
Without this change, goa returns a "fault" error from server even when the error was client side (e.g., goa request validation failed).

This PR makes sure that we return any validation errors during encode/decode as it is without changing them to grpc errors. I have also exported Goa's encode/decode error names so that it can be used in the design for mapping error name to a status code of choice.

NOTE: This is a **breaking** change because
* it changes the error name from "fault" to the actual error name that happened during encode/decode.
* it changes the grpc status code to "Unknown" instead of "InvalidArgument" unless the corresponding encode/decode error is mapped to a code of choice in the design (see example below)

```
Error(goa.InvalidLength)
GRPC(func() {
    Response(goa.InvalidLength, CodeInvalidArgument)
})
```

### DIFF
--- a/dsl/error.go
+++ b/dsl/error.go
@@ -3,6 +3,53 @@ package dsl
 import (
 	"goa.design/goa/v3/eval"
 	"goa.design/goa/v3/expr"
+	pkg "goa.design/goa/v3/pkg"
+)
+
+const (
+	// The constants below make it possible for the service specific code to
+	// return error names that are consistent with the names used by the
+	// generated request and response payload validation code.
+	//
+	// Usage:
+	//
+	// var _ = Service("divider", func() {
+	//     Error(MissingField)
+	//     Error(InvalidRange)
+	//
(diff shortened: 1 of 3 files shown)


raphael:
Thank you for the PR, looks great! Just one suggestion: could we also add the error constants to the DSL package so that they can be properly documented? The definitions could alias the ones in the `pkg` package. So in `dsl/error.go` there could be:

```go
const (
	// The constants below make it possible for the service specific code to
	// return error names that are consistent with the names used by the
	// generated request and response payload validation code.
	// Usage:
	// [Provide meaningful example here]

	// InvalidFieldType is the error name for invalid field type errors.
	InvalidFieldType = pkg.InvalidFieldType
	// MissingField is the error name for missing field errors.
	MissingField = ...
)
```

raphael:
Thank you!

PR #2956, opened 2021-11-30
Properly handle hijacked connection in ResponseCapture middleware

raphael:


raphael:
The change in this PR is not quite right as the `WriteHeader` and `Write` methods should not be called in the first place for hijacked connections.

PR #2974, opened 2022-01-09
Make it easier to mount multiple servers at once

ernesto-jimenez:
When initialising multiple servers, we found that it was pretty easy to setup a middleware across multiple servers with `(goahttp.Server).Use`.

However, mounting multiple servers into a muxer was clunky, having to call the package-level `Mount` with each of the servers.

This aims to allow the same simplicity have to use middlewares with mounting.

Example:

```go
// Doing
s := goahttp.Server{s1, s2, s3}
s.Use(httpmiddleware.RequestID())
s.Mount(mux)

// Instead of
s := goahttp.Server{s1, s2, s3}
s.Use(httpmiddleware.RequestID())

server1.Mount(mux, s1)
server2.Mount(mux, s2)
server3.Mount(mux, s3)
```

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -13,6 +13,12 @@ type (
 		Use(func(http.Handler) http.Handler)
 	}
 
+	// Mounter is the interface for servers that allow mounting their endpoints
+	// into a mutex.
+	Mounter interface {
+		Mount(Muxer)
+	}
+
 	// Servers is a list of servers.
 	Servers []Server
 )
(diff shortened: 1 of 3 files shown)


http/server.go
```diff
@@ -13,6 +13,12 @@ type (
 		Use(func(http.Handler) http.Handler)
 	}
 
+	// Mounter is the interface for servers that allow mounting their endpoints
+	// into a mutex.
```
raphael: Should that read `muxer` instead of `mutex`?

raphael (commented):
Thank you, this looks great - just a small typo that might be good to fix.

http/server.go
ernesto-jimenez: It certainly should have been `muxer`. I'll blame my internal autocorrect 😅

```suggestion
	// into a muxer.
```

ernesto-jimenez:
@raphael updated. Thank you!

PR #2984, opened 2022-01-25
Prefix proto filenames

### LINKED ISSUE #2949: proto: file "error.proto" is already registered
The proto world requires you to use distinct names, package-qualified. The message names can be the same, but you should use different package names to keep them isolated. Currently `"goa.design/goa/v3/grpc/pb"` use error package and if another dependency use the same package the following error occurs - `proto: file "error.proto" is already registered`

MichaelUrman:
To avoid name clashes with other simple protobuf names, goa can namespace by path or filename. This PR chooses to do so by filename prefix as it's a small change.

- Goa's own files use prefix with `goadesign_goa_`: `error.go` becomes `goadesign_goa_error.proto`
- Goa generates service proto files with prefix `goadesign_goagen_`, e.g. `goadesign_goagen_calc.proto`

While this causes some source-control churn, including lots of ripple in the unexported identifiers used in the .pb.go file, I believe it introduces no incompatibilities. With this in place, the identified conflict from github.com/googleapis/gax-go/v2's internal error.proto no longer conflicts.

Fixes #2949


raphael:
This is great, thank you!

PR #2988, opened 2022-02-04
Implement service `PathName` where snake_cased `VarName` was used before

antichris:
Add new `PathName` field to the `service.Data` struct, and replace all uses of the `VarName` field in path context with that.

Out of the 37 references to the `VarName` field, 31 were as an argument for `codegen.SnakeCase()` to be used in a file or import path. There are **only 4** instances where the `VarName` field value is actually used to name a service type variable.

This would also come in handy for resolving goadesign/plugins#116.

*Since `PathName` might not be the best name for this field, I'm open to suggestions.*

### DIFF
--- a/codegen/example/example_server.go
+++ b/codegen/example/example_server.go
@@ -54,7 +54,7 @@ func exampleSvrMain(genpkg string, root *expr.RootExpr, svr *expr.ServerExpr) *c
 		sd := service.Services.Get(svc)
 		svcData[i] = sd
 		specs = append(specs, &codegen.ImportSpec{
-			Path: path.Join(genpkg, codegen.SnakeCase(sd.VarName)),
+			Path: path.Join(genpkg, sd.PathName),
 			Name: scope.Unique(sd.PkgName),
 		})
 	}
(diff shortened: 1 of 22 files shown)


raphael:
This is great, how about `Filename` instead of `PathName`? Or `Directory`?

antichris:
> how about `Filename` instead of `PathName`? Or `Directory`?

Yeah, I considered both of those. `Directory` or `DirName` seemed the best fit, but, because there is this one instance where it is used to derive a file name, and it seemed to me a bit awkward naming that field something directory-related because of it, I stuck with the little less specific `PathName`: https://github.com/goadesign/goa/blob/071a6b972cb24c13f3a1897ee366708334944f80/codegen/service/example_svc.go#L63-L64

`PathName` also fits the other service `Name` fields, `StructName`, `VarName` and `PkgName`, that all mean the name of the service to be used in the specific context (in path, for its struct, variable or package, respectively).

But, if you think none of this is a reason enough, I'd be happy to change it.

raphael:
OK makes sense, thank you for the explanation.

PR #2989, opened 2022-02-04
Add to Multiply

JohnRoesler:


antichris:
Why?

README.md
```diff
@@ -113,8 +113,8 @@ var _ = Service("calc", func() {
 ```
 
 This file contains the design for a `calc` service which accepts HTTP GET
-requests to `/add/{a}/{b}` where `{a}` and `{b}` are placeholders for integer
-values. The API returns the sum of `a` and `b` in the HTTP response body.
+requests to `/multiply/{a}/{b}` where `{a}` and `{b}` are placeholders for integer
+values. The API returns result of `a` multiplied by `b` in the HTTP response body.
```
raphael: missing "the" (The API returns the result of...)

README.md
```diff
@@ -183,15 +183,15 @@ is intended to generate just that: an *example*, in particular it is not
...
-`calc.go` and change the code of the `add` function as follows:
+request so all we have to do is to perform the actual multiplication. Edit the file
+`calc.go` and change the code of the `multiply` function as follows:
 
 ```go
-// Add returns the sum of attributes a and b of p.
-func (s *calcsrvc) Add(ctx context.Context, p *calc.AddPayload) (res int, err error) {
+// Multiply returns the multiplied value of attributes a and b of p.
+func (s *calcsrvc) Multiply(ctx context.Context, p *calc.MultiplyPayload) (res int, err error) {
         return p.A + p.B, nil
```
raphael: Please change implementation to `p.A * p.B`

raphael (changes requested):
This is great, just two small changes needed

README.md
antichris: Isn't it "the product"?

raphael: Yes that would be even better "The API returns the product of ..."

JohnRoesler:
```suggestion
values. The API returns the product of `a` multiplied by `b` in the HTTP response body.
```

README.md
JohnRoesler:
```suggestion
        return p.A * p.B, nil
```

README.md
JohnRoesler: done 👍

README.md
JohnRoesler: done 👍

raphael (approved):

PR #2999, opened 2022-02-14
Remove blank spaces when snake casing

nitinmohan87:


raphael:
Thanks! I'm wondering if replacing spaces with underscores wouldn't result in better results? Seems strange that `"aa AA aa"` would result in `"aa_a_aaa"` - I would have expected that to produce `"aa_aa_aa"`.

nitinmohan87:
What about leading and trailing blank spaces? "  aa aa  "" would produce "\_\_aa_aa\_\_". So kebab case would return "--aa--aa--". I am ok with changing to underscores. Maybe we can trim the leading and trailing blank spaces but any spaces in the middle we change to underscores?

raphael:
Makes sense, I guess we could also replace any consecutive white spaces with a single underscore (after trimming).

Issue #3009, opened 2022-02-24
Default value breakage


maxmcd:
I think this is a similar issue to: https://github.com/goadesign/goa/issues/2733

We're seeing compiler errors in generated code after upgrading from 3.1.3 to any future version.

We have an attribute with a custom type:

```go
Attribute("exdates", func() {
	Default([]time.Time{})
	Meta("struct:field:type", "[]time.Time", "time")
	Meta("struct:tag:json", "-")
})
```

When the marshal function is generated it contains the following lines:

```go
{
	var zero string
	if res.Exdates == zero {
		res.Exdates = []time.Time{}
	}
}
```

And I get the following compiler error:
```
[snip]/encode_decode.go:2450:18: invalid operation: res.Exdates == zero (mismatched types []time.Time and string)
```

If I remove the `Default([]time.Time{})` then goa tries to use type `*[]time.Time` instead of `[]time.Time` which causes errors with our code that expects the generation output with `[]time.Time`:
```
data/rrule.go:161:17: cannot range over rule.Exdates (type *[]time.Time)
```

Open to suggestions on fixing, or consideration as a bug. Ideally we would not use a pointer to a slice, but understand if that's suggested at this point.

raphael:
Just to make sure, I did fix something that could be related recently so have you tried with the latest HEAD version of Goa? (not v3.5.5 but latest Git SHA).

maxmcd:
Thanks for the quick response. Just tried with `goa.design/goa/v3 v3.5.6-0.20220224213728-60194cb9a2f6` and I'm getting the same error when compiling:
```
encode_decode.go:1440:18: invalid operation: res.Exdates == zero (mismatched types []time.Time and string)
encode_decode.go:1637:18: invalid operation: res.Exdates == zero (mismatched types []time.Time and string)
encode_decode.go:2420:18: invalid operation: res.Exdates == zero (mismatched types []time.Time and string)
encode_decode.go:2450:18: invalid operation: res.Exdates == zero (mismatched types []time.Time and string)
```

raphael:
Thank you for checking. The declaration for the `zero` value should be able to honor the custom type definition so this feels like a bug.

maxmcd:
This seems to produce a failing test that is then fixed, but not sure how to handle:

1. Extending the test fixtures.
2. Handling checking for the meta type and then validating that enough attributes are present.
3. General "where things go" and how to extend. 

[...]

PR #3010, opened 2022-02-25
Fix default handling of custom type during Go transformation

maxmcd:
For https://github.com/goadesign/goa/issues/3009

@raphael moving this here for easier discussion.

This change breaks the grpc test. The result expects something like this:
```go
if source.TimeSlice == "" {
    target.TimeSlice = []time.Time{}
}
```
Which is invalid, and I imagine grpc has no business supporting `time.Time`, but maybe I'm not thinking of valid use-cases for custom types within grpc? 

Would it be better to create a new test type to just test this kind of type with `go_transform`, or try and fix the grpc stuff, or something else...?


raphael:
My inclination would be to try and make it work for gRPC as well since protobuf does support extensions via include which could mean new types that are no built-in into Goa. That being said `time.Time` wouldn't work as is because of the way github.com/golang/protobuf/ptypes/timestamp works (the code would have to generate a `AsTime()` call. So, if it ends up being too difficult to implement for gRPC it's probably OK to change the scope of the tests for now.

maxmcd:
@raphael added code changes with attempt at supporting protobuf. 

Now just one weird issue. 

This test failure is rendering `string(source.TimeSlice)` for the conversion:
```
--- FAIL: TestProtoBufTransform (0.04s)
    --- FAIL: TestProtoBufTransform/to-protobuf-type (0.02s)
        --- FAIL: TestProtoBufTransform/to-protobuf-type/defaults-to-defaults (0.00s)
            protobuf_transform_test.go:147: invalid code, got:
                func transform() {
                	target := &WithDefaults{
                		Int:            int32(source.Int),
                		TimeSlice:      string(source.TimeSlice),
                		RequiredInt:    int32(source.RequiredInt),
```

The code that is rendering that here: https://github.com/maxmcd/goa/blob/32f3808786c79abb0a8cfb4d5738d3c89f60a3b8/grpc/codegen/protobuf_transform.go#L495-L502

Strangely (to me) `targetType` is equal to `""` once we get here, so it default back to the string type. I might be misunderstanding a little, but I think I would expect the `Meta` to be populated for the target `expr.AttributeExpr`, but it is `nil`. 

Any thoughts?

maxmcd:
Ah, and should I replace `[]time.Time` with maybe `[]json.RawMessage` as a more representative test case? (Or maybe not important that it is a slice).

maxmcd:
Ok, this somewhat invasive change fixes the generated code (and causes no other test failures), but not sure if this is taking things a little too far.

```diff
diff --git a/grpc/codegen/protobuf.go b/grpc/codegen/protobuf.go
index 4f852f77..f5994b73 100644
--- a/grpc/codegen/protobuf.go
+++ b/grpc/codegen/protobuf.go
@@ -213,7 +213,7 @@ func protoBufGoFullTypeName(att *expr.AttributeExpr, pkg string, s *codegen.Name
 	case expr.UserType, expr.CompositeExpr:
 		return protoBufFullMessageName(att, pkg, s)
 	case expr.Primitive:
-		return protoBufNativeGoTypeName(actual)
+		return protoBufNativeGoTypeName(att)
 	case *expr.Array:
 		return "[]" + protoBufGoFullTypeRef(actual.ElemType, pkg, s)
 	case *expr.Map:
@@ -377,8 +377,12 @@ func protoBufNativeMessageTypeName(t expr.DataType) string {
 // protoBufNativeGoTypeName returns the Go type corresponding to the given
 // primitive type generated by the protocol buffer compiler after compiling
 // the ".proto" file (in *.pb.go).
-func protoBufNativeGoTypeName(t expr.DataType) string {
-	switch t.Kind() {
+func protoBufNativeGoTypeName(att *expr.AttributeExpr) string {
+	typeName, _ := codegen.GetMetaType(att)
+	if typeName != "" {
+		return typeName
+	}
+	switch att.Type.Kind() {
 	case expr.BooleanKind:
 		return "bool"
 	case expr.IntKind:
@@ -402,7 +406,7 @@ func protoBufNativeGoTypeName(t expr.DataType) string {
 	case expr.BytesKind:
 		return "[]byte"
 	default:
-		panic(fmt.Sprintf("cannot compute native protocol buffer type for %T", t)) // bug
+		panic(fmt.Sprintf("cannot compute native protocol buffer type for %T", att.Type)) // bug
 	}
 }

diff --git a/grpc/codegen/protobuf_transform.go b/grpc/codegen/protobuf_transform.go
index 4d346bbb..4aad94e4 100644
--- a/grpc/codegen/protobuf_transform.go
+++ b/grpc/codegen/protobuf_transform.go
@@ -102,7 +102,9 @@ func protoBufTransform(source, target *expr.AttributeExpr, sourceVar, targetVar
 // generating protobuf types (as protogen itself won't honor these overrides).
 func removeMeta(att *expr.AttributeExpr) {
 	codegen.Walk(att, func(a *expr.AttributeExpr) error {
-		a.Meta = nil
+		delete(a.Meta, "struct:field:name")
+		delete(a.Meta, "struct:field:external")
+		// a.Meta = nil
 		return nil
 	})
 }
@@ -493,14 +495,14 @@ func convertType(source, target *expr.AttributeExpr, sourceVar string, ta *trans
 	if source.Type.Kind() != expr.IntKind && source.Type.Kind() != expr.UIntKind {
 		if sourceType != "" || targetType != "" {
 			if ta.proto || targetType == "" {
-				targetType = protoBufNativeGoTypeName(target.Type)
+				targetType = protoBufNativeGoTypeName(target)
 			}
 			return fmt.Sprintf("%s(%s)", targetType, sourceVar)
 		}
 		return sourceVar
 	}
 	if ta.proto {
-		return fmt.Sprintf("%s(%s)", protoBufNativeGoTypeName(target.Type), sourceVar)
+		return fmt.Sprintf("%s(%s)", protoBufNativeGoTypeName(target), sourceVar)
 	}
 	if targetType == "" {
 		targetType = codegen.GoNativeTypeName(target.Type)
```

raphael:
The change above makes sense! It may allow for using protobuf extensions that add support for specific types so seems to be warranted. Thank you for the great contribution!

grpc/codegen/protobuf_transform_test.go
```diff
@@ -258,11 +258,11 @@ const (
...
-		RequiredString: string(source.RequiredString),
-		DefaultBool:    bool(source.DefaultBool),
+		RequiredString: tdtypes.CustomString(source.RequiredString),
+		DefaultBool:    tdtypes.CustomBool(source.DefaultBool),
 	}
 	if source.Integer != nil {
-		target.Integer = int32(*source.Integer)
+		target.Integer = tdtypes.CustomInt(*source.Integer)
 	}
 }
```
maxmcd, line 267: I did end up having to change this test. I think this is a fix!? Not sure...

maxmcd:
Nice! I think this is ready for review. `make` passes for me locally with the exception of one test:
```
--- FAIL: TestPkgImport (0.00s)
    --- FAIL: TestPkgImport/vendored-pkg (0.00s)
        convert_test.go:85: got goa.design/goa/vendor/github.com/some/pkg, expected github.com/some/pkg
FAIL
```

I think that's unrelated to my changes, and maybe (hopefully) is just due to some local environment difference.

raphael:
Thank you!

PR #3011, opened 2022-02-27
Add support for "struct:pkg:path" Meta

raphael:
"struct:pkg:path" overrides where the Go type generated for the enclosing
user type definition is generated. Both the file location and Go package name
are overridden. The file location is computed by appending the location of
the gen folder with the string given as second argument to Meta. So for
example Meta("struct:pkg:path", "foo/bar") will generate the Go type in the
file "gen/foo/bar/name.go" where "name" is the name of the type.
Additionally the package name is computed by taking the base path of the
provided location ("bar" in the prior example). The example below causes the
Go type declaration for MyType to live in the file gen/types/mytype.go with
the package name "types":

    package design

    var MyType = Type("MyType", func() {
        Attribute("name")
        Meta("struct:pkg:path", "types")
    })


lawrencejones:
Oh my, this is going to be a game changer for us!

lawrencejones:
I've just tried this out, unfortunately I'm seeing goa generate duplicates of a type into the same file.

Adding a struct tag to a common type like User:
```go
var User = Type("User", func() {
	Meta("struct:pkg:path", "common")
	Attribute("id", String, "Unique identifier of the user", func() {
		Example("01FCNDV6P870EA6S7TK1DSYDG0")
	})
	// ...
})
```

Causes an error because the file comes out with two Go definitions:
```go
// Code generated by goa v3.5.5, DO NOT EDIT.
//
// User types
//
// Command:
// $ goa gen github.com/incident-io/core/server/api/design -o api

package common

type User struct {
        // Unique identifier of the user
        ID string
        // ID of the user's organisation
        OrganisationID string
        // Slack ID of the user
        SlackUserID string
        // Name of the user
        Name string
        // Email of the user
        Email *string
        // Role of the user
        Role string
        // URL of user avatar image
        AvatarURL string
        // Job title of the user
        JobTitle string
}
// Code generated by goa v3.5.5, DO NOT EDIT.
//
// User types
//
// Command:
// $ goa gen github.com/incident-io/core/server/api/design -o api

package common


type User struct {
        // Unique identifier of the user
        ID string
        // ID of the user's organisation
        OrganisationID string
        // Slack ID of the user
        SlackUserID string
        // Name of the user
        Name string
        // Email of the user
        Email *string
        // Role of the user
        Role string
        // URL of user avatar image
        AvatarURL string
        // Job title of the user
        JobTitle string
}
```

I suspect there's some duplication issue somewhere, possibly when many services require the same struct, but thought I'd post here before digging too deep in case this is an obvious issue.

raphael:
Ah whoops yes it is an obvious issue :) Fix coming shortly.

lawrencejones:
https://github.com/goadesign/goa/pull/3014

I've opened this PR, which is one way to 'fix' the problem. You could also fix this by having the codegen package detect duplicates and be a bit more defensive around it, though you'll be much better at identifying which approach we should go with :)

raphael:
Thank you for the PR, I ended up fixing the issue in #3015 by properly detecting dupes.

PR #3015, opened 2022-02-28
Correctly handle multiple user types with common path

raphael:


lawrencejones:
Nice, this is what I thought a proper fix might look like.

PR #3017, opened 2022-03-01
Failing test for struct:pkg:path bug

lawrencejones:
Where generated code fails to reference user defined types correctly,
assuming they exist in the current package instead of their overriden
import path.

```
go test goa.design/goa/v3/codegen/service
--- FAIL: TestStructPkgPath (0.01s)
    --- FAIL: TestStructPkgPath/multiple (0.00s)
        service_test.go:132: got
            
            // Service is the MultiplePkgPathMethod service interface.
            type Service interface {
                // A implements A.
                A(context.Context, *bar.Bar) (res *bar.Bar, err error)
                // B implements B.
                B(context.Context, *baz.Baz) (res *baz.Baz, err error)
                // EnvelopedB implements EnvelopedB.
                EnvelopedB(context.Context, *EnvelopedBPayload) (res *EnvelopedBResult, err error)
            }
            
            // ServiceName is the name of the service as defined in the design. This is the
            // same value that is set in the endpoint request contexts under the ServiceKey
            // key.
            const ServiceName = "MultiplePkgPathMethod"
            
            // MethodNames lists the service method names as defined in the design. These
            // are the same values that are set in the endpoint request contexts under the
            // MethodKey key.
            var MethodNames = [3]string{"A", "B", "EnvelopedB"}
            
            // EnvelopedBPayload is the payload type of the MultiplePkgPathMethod service
            // EnvelopedB method.
            type EnvelopedBPayload struct {
                Baz *Baz
            }
            
            // EnvelopedBResult is the result type of the MultiplePkgPathMethod service
            // EnvelopedB method.
            type EnvelopedBResult struct {
                Baz *Baz
            }
            
            got vs. expected:
            25c25
            <  ␉ Baz *Baz
            ---
            >     Baz *baz.Baz
            31c31
            <  ␉ Baz *Baz
            ---
            >     Baz *baz.Baz
FAIL
FAIL    goa.design/goa/v3/codegen/service       0.244s
FAIL
```

### DIFF
--- a/codegen/service/testdata/service_code.go
+++ b/codegen/service/testdata/service_code.go
@@ -2399,6 +2399,8 @@ type Service interface {
 	A(context.Context, *bar.Bar) (res *bar.Bar, err error)
 	// B implements B.
 	B(context.Context, *baz.Baz) (res *baz.Baz, err error)
+	// EnvelopedB implements EnvelopedB.
+	EnvelopedB(context.Context, *EnvelopedBPayload) (res *EnvelopedBResult, err error)
 }
 
 // ServiceName is the name of the service as defined in the design. This is the
@@ -2409,7 +2411,19 @@ const ServiceName = "MultiplePkgPathMethod"
 // MethodNames lists the service method names as defined in the design. These
 // are the same values that are set in the endpoint request contexts under the
 // MethodKey key.
-var MethodNames = [2]string{"A", "B"}
+var MethodNames = [3]string{"A", "B", "EnvelopedB"}
+
+// EnvelopedBPayload is the payload type of the MultiplePkgPathMethod service
+// EnvelopedB method.
+type EnvelopedBPayload struct {
+	Baz *baz.Baz
+}
+
+// EnvelopedBResult is the result type of the MultiplePkgPathMethod service
+// EnvelopedB method.
+type EnvelopedBResult struct {
+	Baz *baz.Baz
+}
 `
 
 const PkgPathFoo = `// Foo is the payload type of the PkgPathMethod service A method.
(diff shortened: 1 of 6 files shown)


lawrencejones:
CC @raphael going to dig into this myself now, shout if you know an obvious fix :)

codegen/service/testdata/service_code.go
```diff
@@ -2409,7 +2411,19 @@ const ServiceName = "MultiplePkgPathMethod"
 // MethodNames lists the service method names as defined in the design. These
 // are the same values that are set in the endpoint request contexts under the
 // MethodKey key.
-var MethodNames = [2]string{"A", "B"}
+var MethodNames = [3]string{"A", "B", "EnvelopedB"}
+
+// EnvelopedBPayload is the payload type of the MultiplePkgPathMethod service
+// EnvelopedB method.
+type EnvelopedBPayload struct {
+    Baz *baz.Baz
```
lawrencejones: This is what we should get, but instead we get `Baz *Baz`, where the package is excluded.

lawrencejones:
Ok, I _think_ this fixes the bug, and the tests would appear to suggest it does.

Going to try running this for our real codebase and see if we get something that compiles 🤞

lawrencejones:
Ok, our HTTP code is still broken, but this appears to have fixed the service-level stuff.

lawrencejones:
The last commit I pushed did fix the HTTP generated code, in that the marshal methods produced the right types (ie, stuff in custom packages got the right package prefix).

There's now another issue where if a custom type, say `Parent`, includes a field that is also placed in the same package. At that point, you get:

```go
// api/gen/common/parent.go
package common

type Parent struct {
  Child *common.Child
}
```

As we'll render the package name for the child struct field too 😭 

I don't want to put any more time into this when you'll know far better than I about how to tackle it, so we're going to do something really gross and `sed -i 's/common.//g' api/gen/common/*.go` in our makefile for now, which gets us by.

raphael:
Thank you for looking into this, seems that we may need to introduce the concept of "current package" to the named scope and compare that with the package given to the `FullXXX` methods to determine whether or not to prefix the type. I'll try this approach.

raphael:
Ended up being simpler, see #3018  - closing this in favor of that PR.

PR #3022, opened 2022-03-04
pkg:path types that reference their own package types

lawrencejones:
If you have a user type placed in a package, ie. common, and it has a
field of another type in package common, we codegen a type reference of
common.Thing which will fail, as that code is already in the common
package.


lawrencejones:
@raphael this is a failing test for what we were discussing in Slack!

lawrencejones:
Linking here for those that find it: https://gophers.slack.com/archives/C0FK8EV28/p1656434471212069?thread_ts=1655929395.006179&cid=C0FK8EV28

tl;dr, this PR did not achieve what it needs to, and you'll still get redundant package prefixes to types into an external package.

PR #3031, opened 2022-03-10
Added nil check in case packages.Load can't get module info

### LINKED ISSUE #3029: Unhelpful error (panic) from `goa gen`
I have no idea what I'm doing wrong that is causing this. Of course, it's particularly difficult to correct my mistake if the program panics instead of providing a helpful error message. Looking at the goa code, it appears that it doesn't like something about how I'm specifying my package name, but I have no idea what to do to correct this other than just randomly try stuff.

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x7502ba]

goroutine 1 [running]:
main.NewGenerator({0x7ffe7892f35a, 0x3}, {0x7ffe7892f35e, 0x22}, {0x7e7480, 0x1})
	/home/wmoran2901/go/pkg/mod/goa.design/goa/v3@v3.5.5/cmd/goa/gen.go:63 +0x19a
main.generate({0x7ffe7892f35a, 0x3}, {0x7ffe7892f35e, 0x22}, {0x7e7480, 0x1}, 0x0)
	/home/wmoran2901/go/pkg/mod/goa.design/goa/v3@v3.5.5/cmd/goa/main.go:82 +0xc8
main.main()
	/home/wmoran2901/go/pkg/mod/goa.design/goa/v3@v3.5.5/cmd/goa/main.go:62 +0x451

davepgreene:
This PR adds a `nil` check in case `packages.Load` doesn't return any module info.

Fixes #3029.

### DIFF
--- a/cmd/goa/gen.go
+++ b/cmd/goa/gen.go
@@ -60,8 +60,11 @@ func NewGenerator(cmd string, path, output string) *Generator {
 		fset := token.NewFileSet()
 		p := regexp.MustCompile(`goa.design/goa/v(\d+)/dsl`)
 		for _, pkg := range pkgs {
-			if _, err := os.Stat(filepath.Join(pkg.Module.Dir, "vendor")); !os.IsNotExist(err) {
-				hasVendorDirectory = true
+			// Nil check in case packages.Load can't get module info
+			if pkg.Module != nil {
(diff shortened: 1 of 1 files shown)


raphael:
Thank you!

PR #3042, opened 2022-03-30
Fix conversion of a 64-bit integer to a lower bit size type int without an upper bound check.

NeenuAVarghese:
Reference: https://codeql.github.com/codeql-query-help/go/go-incorrect-integer-conversion/


raphael:
Very cool, thank you!

Issue #3058, opened 2022-04-30
Proposal: How about allowing ServiceError to wrap error?


ikawaha:
## Motivation

The motivation for this proposal comes from the error handling in the middleware.

I tried to create a middleware that only logs specific errors, but I had to parse the error message to distinguish whether the "internal_server_error" was a DB error or some other unexpected error, and so on.

I am thinking that if the original error is embedded in ServiceError, it would be possible to distinguish the error by error type without having to parse the error message.

## TL;DR;

* A field is added to ServiceError to hold the error and the Unwrap interface is implemented.
* The generated helper functions (MakeXXXX) wraps the error.
* The original error is embedded or attached in ServiceError, making it easier to handle errors with errors.Is() and errors.As().

## PoC

* PoC 1: Add a field to the ServiceError structure to hold the error 
    * #3059
* PoC 2: Keep existing MakeXXXXX and also generate MakeXXXWrapped that can wrap errors
    * #3061

ikawaha:
Thanks a lot for all the advice.

I have written another PoC to see if I could handle this by adding new helper functions without modifying the core Goa structures. This method does not affect existing user code. 

#3061 

Please tell me what you think.

PR #3059, opened 2022-04-30
PoC 1: ServiceError wraps error

ikawaha:
This is a PoC for issue https://github.com/goadesign/goa/issues/3058

When using the default error type, Goa generates a corresponding helper function such as MakeXXXX. This helper function takes an error its argument, but uses only the message of the error.

```
// MakeDivByZero builds a goa.ServiceError from an error.
func MakeDivByZero(err error) *goa.ServiceError {
    return &goa.ServiceError{
        Name:    "DivByZero",
        ID:      goa.NewErrorID(),
        Message: err.Error(),             // ← using only message of the error.
     }
}
```

The suggestion is to wrap this error in ServiceError.

```
	ServiceError struct {
		// Name is a name for that class of errors.
		Name string
		// ID is a unique value for each occurrence of the error.
		ID string
		// Pointer to the field that caused this error, if appropriate
		Field *string
		// Message contains the specific error details.
		Message string
		// Is the error a timeout?
		Timeout bool
		// Is the error temporary?
		Temporary bool
		// Is the error a server-side fault?
		Fault bool
		// History tracks all the individual errors that were built into this error, should
		// this error have been merged.
		history []ServiceError
		// Err is a wrapped error, if exists.
		err error                                                // ← ★ embed original error.
	}
```

Wrapping the original error makes it easier to handle errors without having to create custom errors. The error is embedded as a private field in ServiceError, so this change should not confuse users.

```
// MakeDivByZero builds a goa.ServiceError from an error.
func MakeDivByZero(err error) *goa.ServiceError {
        return goa.NewServiceErrorWrapped(err, &goa.ServiceErrorElement{  // ← ★ returns a service error that wraps the error.

                Name:    "DivByZero",
                ID:      goa.NewErrorID(),
                Message: err.Error(),
        })
}
```

### DIFF
--- a/codegen/service/service.go
+++ b/codegen/service/service.go
@@ -399,7 +399,11 @@ const unionValueMethodT = `func ({{ .TypeRef }}) {{ .Name }}() {}
 // input: map[string]{"Type": TypeData, "Error": ErrorData}
 const errorInitT = `{{ printf "%s builds a %s from an error." .Name .TypeName |  comment }}
 func {{ .Name }}(err error) {{ .TypeRef }} {
-	return &{{ .TypeName }}{
+	return {{if (eq "goa.ServiceError" .TypeName) -}}
+               goa.NewServiceErrorWrapped(err, &goa.ServiceErrorElement{ 
+           {{- else -}}
+               &{{ .TypeName }}{
+           {{- end }}
 		Name: {{ printf "%q" .ErrName }},
 		ID: goa.NewErrorID(),
 		Message: err.Error(),
@@ -412,7 +416,7 @@ func {{ .Name }}(err error) {{ .TypeRef }} {
 	{{- if .Fault }}
 		Fault: true,
 	{{- end }}
-	}
+	}{{ if (eq "goa.ServiceError" .TypeName) -}} ) {{- end }}
 }
 `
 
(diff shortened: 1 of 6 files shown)


raphael:
That makes sense! One challenge though is that we also need to consider the client side. One of the ideas behind Goa is that the errors are encoded/decoded transparently so that the client logic doesn't need to worry about differences between transports or even if the call was made in memory directly through the endpoint server method. The code that unmarshals errors would have to re-create the `error` field, I don't know if that's always possible though. Right now Goa encodes the error name in the HTTP `goa-error` header or gRPC metadata so that it can set it when creating the error client side (see https://github.com/goadesign/goa/blob/63cf0bcec4ce584d63c4f83d3788c5e2f937c4eb/http/codegen/client.go#L710) but it doesn't know the actually underlying error type and in general it's probably not possible to recreate it. This means that the server side would have the field set but not the client side which is usually where error checking occurs.

ikawaha:
Thank you for your comment. As you pointed out, this change only embeds the original error in ServiceError, so it does not any affect the behavior of either the server or the client.

The motivation for this proposal comes from the error handling in the middleware.

I tried to create a middleware that only logs specific errors, but I had to parse the error message to distinguish whether the "internal_server_error" was a DB error or some other unexpected error, and so on.

I am thinking that if the original error is embedded in ServiceError, it would be possible to distinguish the error by error type without having to parse the error message.

raphael:
I see, yes that makes sense and is a good use case for it. OK maybe then we could have a clear comment on the field explaining that it is only available in the process that created the error and doesn't get marshaled.

PR #3076, opened 2022-06-10
Add error handling for server shutting down

kitagry:
The code generated by `goa example` ignores an error. So I added a log for it.


raphael:
Looks great, thank you.

PR #3081, opened 2022-06-13
fix: example server code, exit on invalid host error

### LINKED ISSUE #3077: invalid host argument doesn't exit the process
With the default auto-generated server code when providing an invalid host argument the server is not automatically exiting, only the error message is printed.

Example
  - code: https://github.com/goadesign/goa/blob/e1bc1170dc3668e2096147254e3db57ad3cef411/codegen/example/testdata/example_server_code.go#L112-L114
  - cli argument: `--host test.com`

I think the process is up waiting for a cancellation signal that is not triggered for this scenario, a `os.Exit(1)` after the print should stop the process and resolve the issue.

nya1:
Closes https://github.com/goadesign/goa/issues/3077

Updated example http server code, when an invalid host is provided the process will exit with a non-zero exit code.

Used `logger.Fatalf` even thought `fmt.Fprintf` + `os.Exit(1)` is often used in `example_server_code.go`, the ending result is the same but I can change it.

### DIFF
--- a/codegen/example/example_server.go
+++ b/codegen/example/example_server.go
@@ -258,16 +258,14 @@ func main() {
 						}
 					}
 					if !{{ .VarName }}Seen {
-						fmt.Fprintf(os.Stderr, "invalid value for URL '{{ .Name }}' variable: %q (valid values: {{ join .Values "," }})\n", *{{ .VarName }}F)
-						os.Exit(1)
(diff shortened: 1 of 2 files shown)


raphael:
Thank you for the PR! You're right that the use of `logger.Fatalf` feels a bit inconsistent. I would suggest changing the 3 other instances of `fmt.Fprintf`+`os.Exit(1)` to `logger.Fatalf` as well - one line is better than two :)

nya1:
I see, I've now updated example_server.go and reflected the changes on the tests

raphael:
Looks great, thank you!

PR #3082, opened 2022-06-18
Fix a parsing bug for user defined imports

ikawaha:
A bug in the recursion of the safelyGetMetaTypeImports function may cause user-defined imports to be missing.

# Example Design

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

	Method("foo", func() {
		Payload(func() {
			Attribute("x", MapOf(String, String, func() {
				Elem(func() {
					Meta("struct:field:type", "json.RawMessage", "encoding/json")
				})
			}))
		})
		Result(Empty)
		HTTP(func() {
			GET("/foo")
		})
	})
})

```

# Details 

The "encoding/json" must be included in the imports of `gen/calc/service.go` file, but is not.

```console
$ goa version
Goa version v3.7.6
$ goa gen calc/design
gen/calc/client.go
gen/calc/endpoints.go
gen/calc/service.go
... snip
$ cat gen/calc/service.go |head -20
// Code generated by goa v3.7.6, DO NOT EDIT.
//
// calc service
//
// Command:
// $ goa gen calc/design

package calc

import (
	"context"     // ← ★
)

// Service is the calc service interface.
type Service interface {
	// Foo implements foo.
	Foo(context.Context, *FooPayload) (err error)
}
```

### DIFF
--- a/codegen/import.go
+++ b/codegen/import.go
@@ -177,7 +177,7 @@ func safelyGetMetaTypeImports(att *expr.AttributeExpr, seen map[string]struct{})
 		uniqueImports[*im] = struct{}{}
 	}
 	for imp := range uniqueImports {
-		// Copy loop variable into body so next iteration doesnt overwrite its address https://stackoverflow.com/questions/27610039/golang-appending-leaves-only-last-element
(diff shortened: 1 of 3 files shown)


raphael:
This is great, thank you!

PR #3084, opened 2022-06-28
Fixes #3083: properly generate attributes with meta `struct:pkg:path`

### LINKED ISSUE #3083: Missing package when using `struct:pkg:path` inside an attribute
We were extracting some types that we use as attributes into their own package using `Meta("struct:pkg:path", "types")`.

However, when we try to do so, the struct with the generated attribute generates code looking for the struct in the current package rather than the imported `types`.

Example:
```go
var PkgPathPayloadAttributeDSL = func() {
		Attribute("IntField", Int)
		Meta("struct:pkg:path", "foo")
	})
	var Bar = Type("Bar", func() {
		Attribute("Foo", Foo)
	})

	Service("PkgPathPayloadAttributeDSL", func() {
		Method("Foo", func() {
			Payload(Bar)
			Result(Bar)
		})
	})
}
```

We would expect generated code for `Bar` to be something like:
```go
type Bar struct {
    Foo *types.Foo
}
```

However, it generates this instead:
```go
type Bar struct {
    Foo Foo
}
```

ernesto-jimenez:
### What was the issue?

What was meant to be a check on the parentPkg to avoid recursive
definitions wasn't actually checking the parent package, but the package
of the field itself.

It was a bug that got introduced in #3022 and never worked as expected.

### Solution:

Injecting the parent package for the initial definition seems to address
this.


codegen/scope.go
```diff
@@ -138,13 +144,11 @@ func (s *NameScope) goTypeDef(att *expr.AttributeExpr, ptr, useDefault bool, pkg
...
-					if UserTypeLocation(ut) != nil {
-						parentPkg = UserTypeLocation(ut).PackageName()
-					}
-				}
-				tdef = s.goTypeDef(at, ptr, useDefault, parentPkg)
+				tdef = s.goTypeDef(at, ptr, useDefault, pkg)
+				// if name == "Foo" {
+				// 	tdef += "XXX"
+				// 	// panic(at)
+				// }
```
ernesto-jimenez:
```suggestion
```

ernesto-jimenez:
@raphael this fixes part of the issue in #3083

The `gen/$svc/service.go` generated is valid, however, the following files are invalid:
* `gen/http/$svc/client/encode_decode.go`
* `gen/http/$svc/server/encode_decode.go`

Instead of using the type from `struct:pkg:path`, it is still importing it from `gen/$svc `

Any suggestions on how to address the second part?

raphael:
Thank you for this, will take a look

ernesto-jimenez:
@raphael I have added a failing test to illustrate the remaining issue.

I reckon the issue might be with `codegen.GoTransform`. I just saw it contains no tests for types with `Meta("struct:pkg:path")`

PR #3114, opened 2022-08-05
Fix gRPC Meta("struct:field:type")

MichaelUrman:
The three commits increase the coverage of gRPC handling of `Meta("struct:field:type")`, fixing regressions since #2829. Without this fix, a design similar to the new test case generates code with mismatched types.

The first commit adds StructMetaTypeDSL to cover a Method with a Payload and Result using defined types, including a rough outline of what the expected output should be. The second fixes a regression in type handling, and I believe properly updates the expectations of the test case that was added at the same time. The third fixes both implementation and expected output for defaults handling when multiple fields have to be checked against a zero value.

As for the alleged regression, I believe that #3010 introduced flawed handling of protobuf types. Namely it used the `Meta("struct:field:type")` types for values in the protobuf struct, rather than properly casting them back to their underlying types. To answer @maxmcd's [question](https://github.com/goadesign/goa/pull/3010/files#r815349101) six months late:

> ```diff
> 	customSvcToCustomProtoCode = `func transform() {
> 	target := &CustomTypes{
> -		RequiredString: string(source.RequiredString),
> -		DefaultBool:    bool(source.DefaultBool),
> +		RequiredString: tdtypes.CustomString(source.RequiredString),
> +		DefaultBool:    tdtypes.CustomBool(source.DefaultBool),
> 	}
> 	if source.Integer != nil {
> -		target.Integer = int32(*source.Integer)
> +		target.Integer = tdtypes.CustomInt(*source.Integer)
> 	}
>  }
> ```
> I did end up having to change this test. I think this is a fix!? Not sure...

I can now confidently answer that the original test case was and still is correct. Since I reverted the entire change to `protoBufNativeGoTypeName`, I would appreciate if @maxmcd could verify this does not break his desired use case.


raphael:
My bad for not catching that earlier. Custom field types should be used for types generated at the service level - not transport level and so `customSvcToCustomProtoCode` should indeed not have been modified to consider custom types. Thank you for adding a test and for the great PR!

Issue #2221, opened 2019-07-22
v2/v3: Proposal to support custom zero values (a.k.a Zero DSL) for gRPC


nitinmohan87:
~~NOTE: This will be a **breaking** change.~~

Due to [protocol buffer's handling of missing values](https://developers.google.com/protocol-buffers/docs/proto3#default) it is impossible to figure out if the default value is set by the application or protobuf. As a consequence, there is a limitation for required attributes to never fail validation because such attributes are always initialized to a default value by protobuf. To workaround this situation and provide parity with how Goa treats required attributes in HTTP, we propose a `Zero` DSL which can be set to a custom value on an `Attribute` in the design to denote its zero-value.

## The `Zero` DSL
```
Service("my_service", func() {
    Method("my_method", func() {
        Payload(func() {
            Field(1, "my_string", String, "My zero is different", func() {
                Zero("this_is_my_zero_value")
            })
            Field(2, "no_zero_string", String, "My zero is not different")
            Required("my_string", "no_zero_string")
        })

        Result(func() {
            Field(1, "my_integer", Int, "My zero is also different", func() {
                Zero(-199999999)
                Minimum(0)
            })
            Field(2, "no_zero_integer", Int, "My zero is not different")
        })

        GRPC(func() {})
    })
})
```
### Validations
* If `my_string == ""` in the message then goa-generated validation will fail (since `my_string` is required). If `my_string ==  "this_is_my_zero_value"`, validation succeeds.
* If `no_zero_string == ""` in the message then goa-generated validation WILL NOT fail (there is no way to know who set the empty string).
### Transforming Protobuf Request Message to Goa Payload type
* `my_string` in goa payload will be automatically set to empty string.
* `no_zero_string` in goa payload will have the same value as the protobuf message.

### Validations
* If `my_integer == -199999999`, goa-generated validation won't fail even though `Minimum` is set to `0` on the attribute.
### Transforming Protobuf Response Message to Goa Result
* If `my_integer == 0` in the message then goa result will have `my_integer` set to `nil` because it is an optional field. If `my_integer == -199999999`, then goa result will have `my_integer` set to a pointer to `0`.
* If `no_zero_integer == 0`  in the message then goa result will have `no_zero_integer` set to a pointer to `0` (no way to know who set the zero value) because it is an optional field.

The `Zero` DSL will have no effect for HTTP endpoints.

smessier:
[...]

PR #3119, opened 2022-08-08
Omit setting grpc fields to zero defaults

MichaelUrman:
The expected behavior for zero-valued defaults bothered me; it needlessly compared the proto message fields against zeros only to set the goa fields to zeros. This PR explicitly tests a variety of built-in types and avoids the compare and set.

It's possible the isNonZero would be better written with reflect, to reduce its code. Or moved to, say, expr.AttributeType, so that we could instead write something like this in protobuf_transform:

```diff
- if tdef := tgtc.DefaultValue; tdef != nil && ta.TargetCtx.UseDefault && isNonZero(tdef) {
+ if tgtc.HasNonzeroDefault() && ta.TargetCtx.UseDefault {
+ 	tdef := tgtc.DefaultValue
```

### DIFF
--- a/grpc/codegen/client_types_test.go
+++ b/grpc/codegen/client_types_test.go
@@ -23,6 +23,7 @@ func TestClientTypeFiles(t *testing.T) {
 		{"with-errors", testdata.UnaryRPCWithErrorsDSL, testdata.WithErrorsClientTypeCode},
 		{"bidirectional-streaming-same-type", testdata.BidirectionalStreamingRPCSameTypeDSL, testdata.BidirectionalStreamingRPCSameTypeClientTypeCode},
 		{"struct-meta-type", testdata.StructMetaTypeDSL, testdata.StructMetaTypeTypeCode},
+		{"default-fields", testdata.DefaultFieldsDSL, testdata.DefaultFieldsTypeCode},
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
--- a/grpc/codegen/proto_test.go
+++ b/grpc/codegen/proto_test.go
@@ -26,6 +26,7 @@ func TestProtoFiles(t *testing.T) {
 		{"method-with-acronym", testdata.MethodWithAcronymDSL, testdata.MethodWithAcronymProtoCode},
 		{"custom-package-name", testdata.ServiceWithPackageDSL, testdata.ServiceWithPackageCode},
 		{"struct-meta-type", testdata.StructMetaTypeDSL, testdata.StructMetaTypePackageCode},
+		{"default-fields", testdata.DefaultFieldsDSL, testdata.DefaultFieldsPackageCode},
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
--- a/grpc/codegen/protobuf_transform.go
+++ b/grpc/codegen/protobuf_transform.go
@@ -323,7 +323,7 @@ func transformObject(source, target *expr.AttributeExpr, sourceVar, targetVar st
 		// Default value handling. We need to handle default values if the target
 		// type uses default values (i.e. attributes with default values are
 		// non-pointers) and has a default value set.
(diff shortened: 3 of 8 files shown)


MichaelUrman:
I was slightly surprised to learn the following code issues an error. Should I file a bug report, and/or attempt to address it?

```go
    		Field(14, "typed", Int64, func() {
			Meta("struct:field:type", "time.Duration", "time")
			Default(0 * time.Second)
		})
```

> testing.go:18: [testdata/dsls.go:932] default value 0 is incompatible with attribute of type int64 in attribute

Namely it appears that a default of `time.Duration(0)` is not accepted, despite the field type being overridden to time.Duration. (In the meantime, it seems that isNonZero doesn't have to verify underlying types.)

MichaelUrman:
While writing this I had several misgivings over the new-ish defaults behavior in gRPC at all. On the one hand, it means that setting a nonzero default makes it impossible to send a zero value over gRPC. On the other hand, without it a nonzero default silently does something you may not expect. I don't have a good resolution to offer here, but suspect we should err towards allowing it, and just clarify the documentation for dsl.Default(). If there's enough context to prove an unexpected interaction, perhaps issue a warning, or create a linter for this.

I'm assuming any of that would be better off in a separate issue or PR. Let me know if you think it should be here.

raphael:
The PR looks good - thank you. Could you please rebase on the latest v3?

As you probably know the challenge with gRPC is that the encoding doesn't support the notion of "missing" vs "has default value" since primitive values used in a gRPC message cannot be nil. We had a proposal to tackle this explicitly: https://github.com/goadesign/goa/issues/2221 - however this never got implemented. The proposal allows designers to define the desired behavior (i.e., allow or prevent a primitive field to have a zero value).

MichaelUrman:
Rebased.

Yes, the missing means zero, zero means default, etc. as covered on 2221 is why I was surprised to see missing mapped to defaults in v3. I must have missed the fanfare when it was introduced. :)

Any thoughts on the inability to specify a default value of the type specified by Meta("struct:field:type")? Since the error seems to come from Default(), which presumably can't see a later Meta call, I expect at best a non-trivial fix.

raphael:
Wrt specifying default values for types equipped with `Meta("struct:field:type")` - Maybe there's a way to improve `IsCompatible` to look for that metadata and use reflection to figure out that the name of the Go literal value matches the name provided to `Meta`. See for example https://github.com/goadesign/goa/blob/951acc00173014741291e9522fb8ce524529903b/expr/types.go#L324 (every type implements this method).

MichaelUrman:
I don't think IsCompatible has enough information. Maybe you can use `eval.Current().(*expr.AttributeExpr)` like its caller `dsls.Default`, but either location only works if `Meta("struct:field:type", ...)` precedes `Default(...)`. To become order-independent, we have to defer verification, say until after `dsl.Attribute`'s call to `eval.Execute`. This alters error reporting, at least for imported types. (I chose to still validate defaults of builtin types in the original location.)

```diff
- testing.go:18: [testdata/dsls.go:935] default value "" is incompatible with attribute of type int64 in attribute
+ testing.go:18: [../../dsl/attribute.go:236] default value _ is incompatible with attribute of type int64 in attribute
```

The attempt did find one [existing error worth fixing](https://github.com/goadesign/goa/commit/f4b5a44e79031d8092be3d19c75aea6482b19409#diff-8cf71849629683f5678c540aef44d9c718315219dffcc02f6947b8f7e378d4e8R199). I'm not satisfied with this approach yet so have kept it off the PR for now: https://github.com/goadesign/goa/commit/f4b5a44e79031d8092be3d19c75aea6482b19409

raphael:
Yeah the type compatibility check for default values would have to be moved to the validation step of the DSL execution, probably here: https://github.com/goadesign/goa/blob/v3/expr/attribute.go#L189 (and it probably should have been there from the get go). Thank you again for this PR, merged!

PR #3123, opened 2022-08-19
Add `operationId` templating

plunkettscott:
This PR adds template support for the `operationId` generated from the Goa DSL.

## Usage

The `operationId` may be modified at three levels, the `API()`, `Service()`, or `Method()`. The closest template to the operation will be used. For example, a custom template in `API()` will be used globally except in cases where an operation's `Service()` or `Method()` defines a template.

To specify a template in any of the supported blocks, use `Meta`.

```go
// This is the default template
Meta("openapi:operationId", "{service}#{method}(#{routeIndex})")
```

## Variables

The following replacement variables are available:

### `{service}`

This variable is replaced with the name of the Service that the operation belongs to.

### `{method}`

This variable is replaced with the name of the Method documented by the operation.

### `(#{routeIndex})`

This variable looks a bit gnarly but can be broken down into two distinct parts:

The `{routeIndex}` component is replaced with the index of the route in cases where more than one path matches the same method.

The outer component, `(# )`, indicates that the `{routeIndex}` should only be replaced with the actual *index* if its greater than 0. The `#` character, which is the default, may be replaced with any content that you wish to render as a spacer between preceding tokens and the `{routeIndex}` replacement.

The following are all valid:

- `(#{routeIndex})` - The default behavior (eg. `#1`, `#2`, etc)
- `(.{routeIndex})` - Places a `.` before the index
- `(someWords{routeIndex})` - Places `someWords` before the index
- `({routeIndex})` - No separating content between preceding tokens and the index.
- `(spaces are also supported{routeIndex})` - Places `spaces are also supported` before the index.

**Important note!** If the `routeIndex` variable is not provided and multiple paths are defined on a single operation, the template will act as if `(#{routeIndex})` were appended to the end of the `operationId` template.

## Static Content

In addition the the above replacement variables, you may also specify static content which will be rendered literally if no replacement variables are found. For example, the following would be acceptable:

```go
// operationId will be: listPosts
Meta("openapi:operationId", "listPosts")
```

### DIFF
--- a/http/codegen/openapi/v3/builder.go
+++ b/http/codegen/openapi/v3/builder.go
@@ -194,9 +194,7 @@ func buildOperation(key string, r *expr.RouteExpr, bodies *EndpointBodies, rand 
 
 	// OpenAPI summary
 	var summary string
-	var operationIDFormat string
-
-	setTemplates := func(meta expr.MetaExpr) {
+	setSummary := func(meta expr.MetaExpr) {
 		for n, mdata := range meta {
 			if (n == "openapi:summary" || n == "swagger:summary") && len(mdata) > 0 {
 				if mdata[0] == "{path}" {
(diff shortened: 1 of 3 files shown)


raphael:
Awesome PR, thank you!

PR #3126, opened 2022-08-20
Fix union transform

### LINKED ISSUE #3122: Build errors when using oneOf and nested types in StreamingResult
Since v3.8.0 (https://github.com/goadesign/goa/issues/3113), I've been able to use `OneOf` in `StreamingResult` when using result types. Thank you for the fix, @raphael!

However, a similar problem appeared when I tried to nest another result type.

```
# github.com/sevein/oneof/gen/chatter
gen/chatter/service.go:102:17: res.Payload.Item undefined (type interface{payloadVal()} has no field or method Item)
gen/chatter/service.go:126:18: vres.Payload.Item undefined (type interface{payloadVal()} has no field or method Item)
```

See [failed build](https://github.com/sevein/oneof/runs/7885866162?check_suite_focus=true) for more details.

Same repo for this issue: https://github.com/sevein/oneof.
Design change: https://github.com/sevein/oneof/commit/72089390edf4baec76bef413e3889a39aa2a0683.

Thank you in advance!

raphael:
When union type has fields that need initialization

Fixes #3122

### DIFF
--- a/codegen/go_transform.go
+++ b/codegen/go_transform.go
@@ -52,7 +52,6 @@ func init() {
 // newVar if true initializes a target variable with the generated Go code
 // using `:=` operator. If false, it assigns Go code to the target variable
 // using `=`.
-//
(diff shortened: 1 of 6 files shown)


sevein:
Thanks, this is working now!

PR #3133, opened 2022-08-26
Fix types with different names are output as the same type in OpenAPI

najeira:


najeira:
sorry. I made a mistake. This hasn't been tested yet.

PR #3143, opened 2022-09-08
Replace any with interface{} - supporting older go installs

McDeez:
Switching to interface{} to allow for older Go installs to still work.


raphael:
Makes sense, thank you.

Issue #3129, opened 2022-08-25
Union attributes are not described by the OpenAPI specification


sevein:
Types that are only used as union attributes are not being described by Goa in the OpenAPI specification.
That makes sense given that they map as JSON-encoded strings, but it would be useful to have a mechanism that can be used to force a type to appear in the spec. I have tried both `type:generate:force` and `openapi:generate` with no luck.

Link to [conversation in Slack](https://gophers.slack.com/team/U043DU7LM).

Thank you!

raphael:
#3142 makes the OpenAPI v3 spec generator generate schemas for "forced types" (i.e. types that have the `"generate:type:force"` metadata).

PR #3144, opened 2022-09-08
Generate OpenAPI spec for forced result types

sevein:
Similar to https://github.com/goadesign/goa/pull/3142 but focused on forced result types.

Relates to #3129.


raphael:
Good catch, thank you!

PR #3147, opened 2022-09-22
Closes #3105 - Add GoaErrorName and deprecate ErrorName

### LINKED ISSUE #3105: Add `GoaErrorName` as a replacement for `ErrorName`
### What do we want to do?

We have a middleware that captures unexpected errors and sends them to our error reporting infrastructure.

So far, we had been using `errors.As(err, *goa.ServiceError)`. However, we are now defining some custom error types and we would like to avoid a massive type switch.

We would prefer a type check with an interface, such:

```go
type isGoaError interface {
  IsGoaError() bool
}

func middleware(...) ... {
  // ...
  if _, ok := err.(isGoaError); !ok {
    report(error)
  }
}
```

### Proposal

- Add `func (*goa.ServiceError) IsGoaError() bool`
- Generate the `IsGoaError` method for types with `Meta("struct:error:name")` (maybe `struct:error:type` instead?)
- Add some documentation. (where would be the right place?)

ernesto-jimenez:
Implemented what we discussed in #3105

1. Add `GoaErrorName` methods and `GoaErrorNamer` interface alongside `ErrorName` and `ErrorNamer`.
2. Deprecate `ErrorName` and `ErrorNamer`.
3. Add adapters to support both `ErrorNamer` and `GoaErrorNamer`.

### DIFF
--- a/codegen/service/service.go
+++ b/codegen/service/service.go
@@ -390,7 +390,14 @@ func (e {{ .Ref }}) Error() string {
 }
 
 // ErrorName returns {{ printf "%q" .Name }}.
+//
+// Deprecated: Use GoaErrorName - https://github.com/goadesign/goa/issues/3105
 func (e {{ .Ref }}) ErrorName() string {
+	return e.GoaErrorName()
+}
+
+// GoaErrorName returns {{ printf "%q" .Name }}.
+func (e {{ .Ref }}) GoaErrorName() string {
 	return {{ errorName . }}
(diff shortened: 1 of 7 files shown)


ernesto-jimenez:
@raphael are you OK with adding the deprecation in the comment? some people's linters will let them know they need to switch they method they use.

raphael:
Yes looks great, thank you.

PR #3149, opened 2022-09-22
Add MethodNames() to http server

ernesto-jimenez:
@raphael based on the discussion from a few weeks ago, suggesting this.

I noticed that some tests are coupled to the order of the sections, hence why I added this at the end.

It would be good to refactor the test so they are agnostic of section order.

### DIFF
--- a/http/codegen/handler_test.go
+++ b/http/codegen/handler_test.go
@@ -1,9 +1,11 @@
 package codegen
 
 import (
+	"path/filepath"
 	"testing"
 
 	"goa.design/goa/v3/codegen"
+	"goa.design/goa/v3/codegen/codegentest"
 	"goa.design/goa/v3/expr"
 	"goa.design/goa/v3/http/codegen/testdata"
 )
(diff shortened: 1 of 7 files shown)


ernesto-jimenez:
@raphael how would you want to test this?

raphael:
Thank you for doing this. Could we change the tests so they rely on the name of the sections instead of the their index?

ernesto-jimenez:
Took a stab at it for the tests that would fail if I placed this after server use. Didn't want this to become a massive PR refactoring of all tests. I do have some ideas on how to improve tests if you are interested, though.

Now I'm puzzled about why tests are passing in Linux but not Windows 🤔

http/codegen/handler_test.go
```diff
@@ -4,38 +4,35 @@ import (
...
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
 			RunHTTPDSL(t, c.DSL)
 			fs := ServerFiles(genpkg, expr.Root)
-			if len(fs) != c.FileCount {
-				t.Fatalf("got %d files, expected %d", len(fs), c.FileCount)
+			sections := codegentest.Sections(fs, "/server.go", "server-handler-init")
+			if len(sections) == 0 {
+				t.Errorf("section not found")
```
ernesto-jimenez:
```suggestion
				t.Fatalf("section not found")
```

http/codegen/server_handler_test.go
```diff
@@ -4,35 +4,31 @@ import (
...
 	}
 	for _, c := range cases {
 		t.Run(c.Name, func(t *testing.T) {
 			RunHTTPDSL(t, c.DSL)
 			fs := ServerFiles(genpkg, expr.Root)
-			if len(fs) != c.FileCount {
-				t.Fatalf("got %d files, expected %d", len(fs), c.FileCount)
+			sections := codegentest.Sections(fs, "/server.go", "server-handler")
+			if len(sections) == 0 {
+				t.Errorf("section not found")
```
ernesto-jimenez:
```suggestion
				t.Fatalf("section not found")
```

raphael:
Could it be because the path used to find the section needs to be `\server.go` on Windows? Maybe the path should be constructed using `os.PathSeparator`?

ernesto-jimenez:
@raphael that was it :)

raphael:
Looks great!

q: does the `lo` package require Go 1.18+? Might be good to avoid needing to force users to upgrade.

ernesto-jimenez:
@raphael removed lo. We are using it constantly internally, so didn't even think about the version :)

Sorry about that.

Tangential: What's your version support policy for `goa`? In the past, I did "two last minor versions", which gave people at least 6 months to upgrade to a new Go version, which seemed fair given the backwards compatibility guarantees the language offers.

raphael:
Thank you for the great PR, love the refactor around tests. There is no official policy around Go version support but that might be good to think about. There are users that require support for older versions (e.g. https://github.com/goadesign/goa/pull/3143) hence my question. If I understand your proposal Goa would currently support both `1.19` and `1.18`? Seems like it might not be enough for some users. Taking this discussion to Slack.

PR #3154, opened 2022-10-01
Move ErrorNamer and GoaErrorNamer to Goa library

raphael:
instead of re-generating the data structures in every transport package.


raphael:
@ernesto-jimenez FYI

PR #3155, opened 2022-10-03
Implement Unwrap interface to AdaptErrorNamer

ikawaha:
**This is a fix to HEAD, not v3.9.0.** (v3.9.0 has same problem)

AdptErrorNamer wraps goa.ServiceError but does not implement the Unwrap interface. Therefore, errors cannot be retrieved with errors.As() and panic occurs in formatter.

### Example generated code:
```
func EncodeDivError(encoder func(context.Context, http.ResponseWriter) goahttp.Encoder, formatter func(ctx context.Context, err error) goahttp.Statuser) func(context.Context, http.ResponseWriter, error) error {
	encodeError := goahttp.ErrorEncoder(encoder, formatter)
	return func(ctx context.Context, w http.ResponseWriter, v error) error {
		var deprecated goa.ErrorNamer
		if errors.As(v, &deprecated) {
			v = goa.AdaptErrorNamer{deprecated} //★ ←  wrap *goa.ServiceError
		}
		var en goa.GoaErrorNamer
		if !errors.As(v, &en) {
			return encodeError(ctx, w, v)
		}
		switch en.GoaErrorName() {
		case "zero_division":
			var res *goa.ServiceError
			errors.As(v, &res) //★ ← cannot unwrap *goa.ServiceError from AdaptError
			enc := encoder(ctx, w)
			var body interface{}
			if formatter != nil {
				body = formatter(ctx, res)
			} else {
				body = NewDivZeroDivisionResponseBody(res)
			}
			w.Header().Set("goa-error", res.GoaErrorName())
			w.WriteHeader(http.StatusBadRequest)
			return enc.Encode(body)
		default:
			return encodeError(ctx, w, v)
		}
	}
}
```

### DIFF
--- a/pkg/error.go
+++ b/pkg/error.go
@@ -272,6 +272,7 @@ func (e *ServiceError) ErrorName() string { return e.Name }
 // GoaErrorName returns the error name.
 func (e *ServiceError) GoaErrorName() string { return e.ErrorName() }
 
+// Unwrap implements Wrapper interface.
 func (e *ServiceError) Unwrap() error { return e.err }
 
 // GoaErrorName returns the error name as defined in the design.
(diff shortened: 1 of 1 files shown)


raphael:
Thank you, good catch! I'm closing this in favor of https://github.com/goadesign/goa/pull/3157 which removes the adapter entirely.

PR #3191, opened 2022-11-08
Add dsl.Randomiser that lets you pass your own example randomiser

isaacseymour:
1. Rename `expr.Random` to `expr.FakerRandom`
2. Add an interface `expr.Randomiser` that covers the methods we need
3. Split out `expr.ExampleGenerator` that handles the 'do we already have an example for this' logic from the 'generate examples' stuff
4. Change `(expr.APIExpr).random` to `(expr.APIExpr).ExampleGenerator`, and initialise it with a faker-based random example generator
5. Add `dsl.Randomiser` which overrides `(expr.APIExpr).ExampleGenerator` with whatever randomiser you want to pass in, wrapped in the non-customisable 'only generate one example for each thing' logic.

I've also added an `ArrayLength` method to the randomiser interface, and updated places that previously would call `r.Int()` to generate an array length to use that.

The motivation for this PR is that we have a large design and many people working on it, and the diffs we get on OpenAPI schema files can be quite hard to read with most of the changes coming from changes to the random examples due to the order of example generation when we add new types (for example).

We'd therefore like to disable all random example generation and use hard-coded default examples, and define manual examples with `dsl.Example` where we'd like them.

### DIFF
--- a/dsl/api.go
+++ b/dsl/api.go
@@ -150,14 +150,14 @@ func License(fn func()) {
 	eval.IncompatibleDSL()
 }
 
-// Randomiser sets the API example randomiser.
+// Randomizer sets the API example randomizer.
 //
-// Randomiser must appear in a API expression.
+// Randomizer must appear in a API expression.
 //
-// Randomiser takes a single argument which is an implementation of
-// expr.Randomiser.
+// Randomizer takes a single argument which is an implementation of
+// expr.Randomizer.
 //
-// The default randomiser uses the API name as the seed, to get consistent
+// The default randomizer uses the API name as the seed, to get consistent
 // random examples.
 //
 // Example:
@@ -166,9 +166,9 @@ func License(fn func()) {
 //        Random(expr.NewRandom("different seed"))
 //    })
 //
-func Randomiser(randomiser expr.Randomiser) {
+func Randomizer(randomizer expr.Randomizer) {
 	if s, ok := eval.Current().(*expr.APIExpr); ok {
-		s.ExampleGenerator = &expr.ExampleGenerator{Randomiser: randomiser}
+		s.ExampleGenerator = &expr.ExampleGenerator{Randomizer: randomizer}
 		return
 	}
 	eval.IncompatibleDSL()
--- a/dsl/api.go
+++ b/dsl/api.go
@@ -163,9 +163,17 @@ func License(fn func()) {
 // Example:
 //
 //    var _ = API("divider", func() {
-//        Random(expr.NewRandom("different seed"))
+//        Randomizer(expr.NewFakerRandomizer("different seed"))
 //    })
 //
+// There's also a deterministic randomizer which will only generate one example
+// for each type, so all strings are "abc123", all ints are 1, etc.
+//
+// Example:
+//
+//    var _ = API("divider", func() {
+//        Randomizer(expr.NewDeterministicRandomizer())
+//    })
 func Randomizer(randomizer expr.Randomizer) {
 	if s, ok := eval.Current().(*expr.APIExpr); ok {
 		s.ExampleGenerator = &expr.ExampleGenerator{Randomizer: randomizer}
--- a/dsl/api.go
+++ b/dsl/api.go
@@ -150,6 +150,30 @@ func License(fn func()) {
 	eval.IncompatibleDSL()
 }
 
+// Randomiser sets the API example randomiser.
+//
+// Randomiser must appear in a API expression.
+//
+// Randomiser takes a single argument which is an implementation of
+// expr.Randomiser.
+//
+// The default randomiser uses the API name as the seed, to get consistent
+// random examples.
+//
+// Example:
+//
+//    var _ = API("divider", func() {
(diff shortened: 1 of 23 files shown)


dsl/api.go
```diff
@@ -150,6 +150,30 @@ func License(fn func()) {
...
+// Randomiser takes a single argument which is an implementation of
+// expr.Randomiser.
+//
+// The default randomiser uses the API name as the seed, to get consistent
+// random examples.
+//
+// Example:
+//
+//    var _ = API("divider", func() {
+//        Random(expr.NewRandom("different seed"))
```
raphael: Should that be `Randomizer`?

dsl/api.go
```diff
@@ -150,6 +150,30 @@ func License(fn func()) {
 	eval.IncompatibleDSL()
 }
 
+// Randomiser sets the API example randomiser.
```
raphael: Just to be consistent with the rest of the code base (which uses the American based syntax) could we use `Randomizer`?

raphael (commented):
This looks great - the functionality makes sense. There is a small breaking change in that the public `NewRandom` function signature is changing but that's probably OK.  I added a couple of small suggestions to the PR.
I'm also wondering how you are intending on using this functionality. In particular, would it make sense to include whatever implementation of the `Randomizer` interface you are going to be building as an alternative built-in into Goa for other people that have the same desire of not having examples change when the design change?

dsl/api.go
isaacseymour: 🙈 🇬🇧 sorry!

isaacseymour:
I've `s/Randomiser/Randomizer`'d everywhere - force of habit 🙈 

> There is a small breaking change in that the public NewRandom function signature is changing but that's probably OK.

I've removed this breaking change - so `NewRandom` is now the function to build "something which does example generation with the right defaults", as it was before, although its return type has changed. The function for building the faker-backed randomizer is now `NewFakerRandomizer`, which seems clearer than `NewRandom` to me!

> would it make sense to include whatever implementation of the Randomizer interface you are going to be building as an alternative built-in into Goa for other people that have the same desire of not having examples change when the design change?

I've included that in an extra commit ([2b0bd65](https://github.com/goadesign/goa/pull/3191/commits/2b0bd65262a45aaaa34398f8d676261662baf11d)). Very happy to have it included in Goa if you like it, but can also rebase it away 😄

raphael:
Great, the last thing to do is to make sure the tests pass. You can try locally by running `make`. The deterministic example generator is great. Maybe it should be added to the `Randomizer` DSL function header comment as an example since the default uses the faker-based randomizer anyway.

isaacseymour:
I _think_ the tests should pass now - there's one protobuf test I can't get to run locally on master either, but the other ones are now fixed 💪

dsl/api.go
```diff
@@ -150,6 +150,30 @@ func License(fn func()) {
...
+// Randomizer takes a single argument which is an implementation of
+// expr.Randomizer.
+//
+// The default randomizer uses the API name as the seed, to get consistent
+// random examples.
+//
+// Example:
+//
+//    var _ = API("divider", func() {
+//        Random(expr.NewRandom("different seed"))
```
raphael: The example doesn't seem quite right... Also, would be nice to showcase the deterministic randomizer since that's probably the main use case for `Randomizer`.

go.mod
```diff
@@ -3,6 +3,7 @@ module goa.design/goa/v3
 go 1.19
 
 require (
+	github.com/davecgh/go-spew v1.1.1
```
raphael: Debug leftover?

expr/random.go
```diff
@@ -4,25 +4,89 @@ import (
...
-	Seed  string
-	Seen  map[string]*interface{}
-	faker *faker.Faker
-	rand  *rand.Rand
+// Randomizer generates consistent random values of different types given a seed.
+//
+// The random values should be consistent in that given the same seed the same
+// random values get generated.
+type Randomizer interface {
+	// ArrayLength decides how long an example array will be
```
raphael, line 18: Looks like this isn't being actually used with the latest commit. The deterministic randomizer is probably broken with that last change.

raphael (commented):
Looks great, left a few comments inline.

go.mod
isaacseymour: Late night coding 🙈

expr/random.go
isaacseymour: Oops. I got a bit desperate trying to figure out what part of my change had broken the tests and reverted too much of it. [9a5a6f9](https://github.com/goadesign/goa/pull/3191/commits/9a5a6f9cdfcba01a7a76ce2f5f0a510ad579f76d) brings back the use of ArrayLength, and I've tested this out on our design and it has the desired effect.

dsl/api.go
isaacseymour: I'm not 100% sure on the magic comment format for godoc, but I've got two examples here now, I *hope*!

raphael:

PR #3194, opened 2022-11-13
Added detail about Service dsl function to godoc comment

c-reeder:
Added a note in the documentation for the `Service()` DSL function which states that you can call it multiple times and that all those definitions will be merged.


raphael:
Thank you!

PR #3199, opened 2022-11-19
Fixed UUID generation in openapi examples

c-reeder:
The OpenAPI generation currently produces invalid UUIDs in the examples. This is problematic for any tool that pulls those in for testing (ex. Postman). 

Using the regular expression to generate the UUIDs was only acceptable on occasion because it failed to take into account the different UUID variants (reserved, backwards compatible, etc).

This change makes us use the same library dependency that we were _already_ using for checking/inspecting an incoming UUID in order to generate one.

Also the generic fallback was invalid because it had one character too many.

### DIFF
--- a/expr/example.go
+++ b/expr/example.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"time"
 
+	googleuuid "github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


raphael:
This is great, thank you!

PR #3203, opened 2022-11-26
OpenAPI v3 type names refer to openapi:typename in meta

### LINKED ISSUE #3169: OpenAPI Generated Type Names
The names for the types of the Request bodies seem to be solely generated off of the names of the goa method.
Ex. If I have `	Method("create", func() {`, then it will yield `CreateRequestBody`.

The problem with this is that there can be many duplicates.

If I have 5 different services and each one has a `create` method like above, then I will end up with the unfortunately named `CreateRequestBody`, `CreateRequestBody2`, `CreateRequestBody3`, `CreateRequestBody4`, and `CreateRequestBody5` and there is no way to avoid that.

There are a few solutions that come to mind:

- Prefix the name of the service in front of the generated name. Ex. a "create" method inside the "Users" service would yield `UsersCreateRequestBody` or something similar
- Simply make the name of the goa type for that payload the same as the name that is output in the OpenAPI. Thus giving you a way to set the name from code.
- The `TypeName()` function in the DSL currently only affects the generated go type names. This could be altered to produce the same names in the OpenAPI output.
- Some other new DSL function that specifically serves the purpose of letting you set the name of the generated OpenAPI generated type name (whether in an inline defined payload or a separately defined `Type()`.

I'm sure there maybe other similarly helpful solutions, but I think that the main goal here is that we provide:

-  _Some_ better default system for selecting names that makes it much less common for numbers to need to be appended to the end of those generated OpenAPI type names
-  OR some way to override the generated OpenAPI type names
- Or both

Thanks for reading and I'm helping to provide any further information that would enrich this description! :)

najeira:
Fixes: #3169 #3202


raphael (commented):
This looks great. It would be good to document this new feature in the `Meta` function of the `dsl` package. Something like:
```go
//    - "openapi:typename" overrides the name of the type generated in the OpenAPI specification.
//    Applicable to types (including embedded Payload and Result definitions).
//
//    package design
//
//    var Foo = Type("Foo", func() {
//        Attribute("name", String)
//        Meta("openapi:typename", "Bar")
//    })
```

najeira:
Thank you for your review. I will add the comment.

This fix works when defining Meta to Payload or Result, but not works when defining Meta to Type.

I will try to fix that.

najeira:
At generating OpenAPI files, Meta seems to be empty if we set the predefined Type to Payload or Result.

works:

```go
Payload(func() {
	Meta("openapi:typename", "Bar")
	Attribute("name", String)
})
```

not works:

```go
	Meta("openapi:typename", "Bar")
	Attribute("name", String)
})
Payload(Foo)
```

Please let me know if you have any ideas or how to fix it.

najeira:
I force pushed another implementaion.

This supports to the definition of Meta in Type and ResultType.

raphael:

PR #3232, opened 2023-01-23
ClientError: Add original error

RadekDvorak:
Storing the original error on `ClientError` struct enables checking the root cause:  for example
```golang
errors.Is(err, syscall.ECONNREFUSED)
```


raphael:
Looks great, thank you!

PR #3236, opened 2023-01-27
Update log.go

mcauto:
The error is always nil, so no return type is required.

The following message always occurs in gosec:
 `G104 (CWE-703): Errors unhandled.`


raphael:
Hello, thank you for the PR. Yes the standard library logger adapter always returns `nil` but that might not be the case for all implementations of the interface. For example the Go kit logger returns an error: https://github.com/go-kit/log/blob/main/log.go#L151. This would also be a breaking change so probably not something we can merge as is.

PR #3239, opened 2023-02-01
Remove typo in documented example for CollectionOf DSL

sbchapin:
The documentation for the `CollectionOf` DSL contains a typo of "reminder" when it should be "remainder".

This is used in context of the division/calc example.  If you were to use this documentation example without modification, it would provide example data that does not function with its expected view.


raphael:
Whoops, thank you!

PR #3251, opened 2023-02-27
Fix syntax of dsl.Meta("struct:field:external") from dots to colons

tchssk:
I think it's just a typo but this fix is a breaking change.
It may be better to leave both new and old syntax for a while.

Note: Only one place was already colons.

https://github.com/goadesign/goa/blob/ed6dd56f56945c2d526b77835b6e20ab897f5ce1/grpc/codegen/protobuf_transform.go#L113-L119


raphael:
Good catch, thank you. Yes, I suggest we keep supporting the old syntax but only document the new one.

tchssk:
I restored the deprecated syntax. 31d7fed3e2b94beaac43b2bb85fca291b6f2756b

raphael:
Thank you!

PR #3263, opened 2023-03-18
Rewrite `interface{}` to `any`

ikawaha:
Now that Goa requires Go version 1.20 or higher, I think it is time to rewrite `interface{}` with `any `.

* Replace `interface{}` to `any`
* Not applicable for the following:
    * grpc/pb/goadesign_goa_error.pb.go
* Adjust tests
    * 679a041a27ef469d80bb9bcc195734df832045b9,  91735ff4817971fc14b39302f6d5e04872d6f992


raphael:
Yes that makes sense, thank you for doing this!

Issue #3216, opened 2022-12-23
Required list attributes in Result are return as null if the value is empty


khaledez:
Let's say I have this design:
```go
import . "goa.design/goa/v3/dsl"

var TodosResult = Type("TodosResult", func() {
	Attribute("user", String, func() {
		Description("user ID")
	})

	Attribute("todos", ArrayOf(String), func() {
		Description("List of todos")
	})

	Required("user", "todos") // NOTE HERE todos IS REQUIRED
})

var _ = Service("todo-service", func() {
	Description("Simple Todo Service")
	Error("NotFound")

	Method("get-todos", func() {
		Description("Return a list of todos for user")
		Payload(String)
		Result(TodosResult)
		HTTP(func() {
			GET("/todos/{userID}")
			Response("NotFound", StatusNotFound)
		})
	})
})
```

The above design, generates this service endpoint:
```go
type Service interface {
	// Return a list of todos for user
	GetTodos(context.Context, string) (res *TodosResult, err error)
}
```

And If implement it like this:
```go
type TodosService struct {
}

func (service *TodosService) GetTodos(ctx context.Context, userID string) (res *api.TodosResult, err error) {
	var todos []string // empty list
	return &api.TodosResult{
		User:  userID,
		Todos: todos,
	}, nil
}
```

As you can see, the empty list has no memory allocated, so the returned http response is like this:
```json
{"user":"USER_ID","todos":null}
```

But the expected result to be like this:
```json
{"user":"USER_ID","todos":[]}
```

I checked the generated response decoder:
```go
// NewGetTodosResponseBody builds the HTTP response body from the result of the
// "get-todos" endpoint of the "todo-service" service.
func NewGetTodosResponseBody(res *todoservice.TodosResult) *GetTodosResponseBody {
	body := &GetTodosResponseBody{
		User: res.User,
	}
	if res.Todos != nil {
		body.Todos = make([]string, len(res.Todos))
		for i, val := range res.Todos {
			body.Todos[i] = val
		}
	}
	return body
}
```
It won't allocate an array if the value coming is nil.
I believe the generated code in this case should be like this, specially when the attribute is marked as `Required`:

```go
// NewGetTodosResponseBody builds the HTTP response body from the result of the
// "get-todos" endpoint of the "todo-service" service.
func NewGetTodosResponseBody(res *todoservice.TodosResult) *GetTodosResponseBody {
	body := &GetTodosResponseBody{
		User: res.User,
	}
	if res.Todos != nil {
[...]

PR #3272, opened 2023-04-04
Generate empty array for required types in response

khaledez:
As mentioned in #3216. This code change makes sure that an empty array is initialized in the response if the field is required


raphael:
This looks great, thank you

raphael:
Thank you, this looks great.

PR #3275, opened 2023-04-07
Support custom examples in ErrorResult

nitinmohan87:
With this PR https://github.com/goadesign/goa/pull/3038 we lost the ability to set custom examples for ErrorResult. This PR fixes this issue so that any custom examples for ErrorResult is respected.


raphael:
Looks great, thank you!

Issue #3278, opened 2023-04-14
Something strange happens when using ArrayOf(Int32) with MultipartRequest() in a Method()!


aenawi:
Hi
I'm trying to create an Update method inside a Service that can update multiple records at once. The design script file as follows:

```
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = Service("organization", func() {
	Description("Organization service manages CRUD operations for all Organization information.")

	Error("unauthorized", String, "Credentials are invalid")

	HTTP(func() {
		Path("/org")
		Response("unauthorized", StatusUnauthorized)
	})

	// Update method handles Organization UPDATE functionality
	Method("update", func() {
		Description("Update a single Organization or multi Organizations record(s).")

		Security(JWTAuth, func() {
			Scope("api:read")
			Scope("api:write")
		})

		Payload(func() {
			Field(1, "ids", ArrayOf(Int32), "IDs of Organization to be updated")
			Field(2, "organizations", ArrayOf(Organization), "Array of Organization info that matches the ids attribute")
			TokenField(3, "token", String, func() {
				Description("JWT token used for authentication")
				Example("eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWV9.TJVA95OrM7E2cBab30RMHrHDcEfxjoYZgeFONFh7HgQ")
			})
			Required("ids", "organizations", "token")
		})

		Result(ArrayOf(Int32)) // returns the updated IDs

		Error("not_found", NotFound, "Organizations ID(s) not found")
		Error("invalid-scopes", String, "Token scopes are invalid")

		HTTP(func() {
			PUT("/")
			Param("ids")
			MultipartRequest() // TODO: check CORS configs in API
			Header("token:Authorization", String, "Auth token", func() {
				Pattern("^Bearer [^ ]+$")
			})
			Response("invalid-scopes", StatusForbidden)
		})

		GRPC(func() {
			Response(CodeOK)
			Response("invalid-scopes", CodeUnauthenticated)
		})
	})

})

```
What happens is that when I use 'ArrayOf(Int32)' for IDs field, the output generated by goa inside this file (gen/http/organization/server/encode_decode.go) for the function NewOrganizationUpdateDecoder has this code:
```
idsRaw := r.URL.Query()["ids"]
if idsRaw == nil {
[...]

PR #3280, opened 2023-04-14
Properly handle query string slice decoding

raphael:
When the request is multipart. In this case the code should only return a single value. Make both cases work by following the common pattern of setting the error and not returning right away.


raphael:
Fix https://github.com/goadesign/goa/issues/3278

PR #3315, opened 2023-07-11
Add acronym "SDK" to commonInitialisms

dalv0911:
Hello, I have an issue that the acronym `SDK` is not generated with UPPERCASE.

So I would like to add it as an initialisms which stand for Software Development Kit to commonInitialisms.

Please take a look.


raphael:
Agreed, thank you!

PR #3319, opened 2023-07-24
Handle reader errors in SkipResponseBodyEncodeDecode

MichaelUrman:
Once we've started writing the body, it's too late for normal error handling. Instead issue the sentinal panic(http.ErrAbortHandler) to interrupt the stream.

Since the interrupted stream is much less meaningful than the normal error handling, and servers using SkipResponseBodyEncodeDecode may issue errors early in the stream, read a chunk of the stream looking for an error. If encountered, use normal error handling. Otherwise write the chunk plus whatever remains.

For context, I'm attempting to use Transfer-Encoding: Chunked. Go does this for us, automatically, so long as Content-Length hasn't been specified. But that also means there's no way for it to know if the writes were complete. See related documentation at https://stackoverflow.com/questions/40291861/how-to-break-a-chunked-in-go-using-net-http and https://github.com/golang/go/issues/17790.

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -45,6 +45,7 @@ func serverFile(genpkg string, svc *expr.HTTPServiceExpr) *codegen.File {
 	}
 	sections := []*codegen.SectionTemplate{
 		codegen.Header(title, "server", []*codegen.ImportSpec{
+			{Path: "bytes"},
 			{Path: "context"},
 			{Path: "fmt"},
 			{Path: "io"},
@@ -511,6 +512,14 @@ func {{ .HandlerInit }}(
 	{{- if .Method.SkipResponseBodyEncodeDecode }}
 		o := res.(*{{ .ServicePkgName }}.{{ .Method.ResponseStruct }})
 		defer o.Body.Close()
+		// handle immediate read error like a returned error
+		buf := &bytes.Buffer{}
+		if _, err = buf.ReadFrom(io.LimitReader(o.Body, 8*1024)); err != nil{
+			if err := encodeError(ctx, w, err); err != nil {
+				errhandler(ctx, w, err)
+			}
+			return
+		}
 	{{- end }}
 	{{- if not (or .Redirect (isWebSocketEndpoint .)) }}
 		if err := encodeResponse(ctx, w, {{ if and .Method.SkipResponseBodyEncodeDecode .Result.Ref }}o.Result{{ else }}res{{ end }}); err != nil {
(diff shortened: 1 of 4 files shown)


http/codegen/server.go
```diff
@@ -511,6 +512,14 @@ func {{ .HandlerInit }}(
 	{{- if .Method.SkipResponseBodyEncodeDecode }}
 		o := res.(*{{ .ServicePkgName }}.{{ .Method.ResponseStruct }})
 		defer o.Body.Close()
+		// handle immediate read error like a returned error
+		buf := &bytes.Buffer{}
+		if _, err = buf.ReadFrom(io.LimitReader(o.Body, 8*1024)); err != nil{
```
MichaelUrman: Should we have a real constant, or a configurable peek value instead of the magic 8*1024?

http/codegen/server.go
```diff
@@ -521,10 +530,8 @@ func {{ .HandlerInit }}(
 		}
 	{{- end }}
 	{{- if .Method.SkipResponseBodyEncodeDecode }}
-		if _, err := io.Copy(w, o.Body); err != nil {
-			if err := encodeError(ctx, w, err); err != nil {
-				errhandler(ctx, w, err)
-			}
+		if _, err := io.Copy(w, io.MultiReader(buf, o.Body)); err != nil {
```
MichaelUrman: Is it permissible to read from o.Body again if it returned an io.EOF when reading into buf?

MichaelUrman (commented):
What tests can we add for this? I've performed some manual testing, but was surprised the codegen tests didn't flag this, and assume nothing is verifying the resulting run-time behavior.

http/codegen/server.go
raphael: My vote would be a real constant until we see a real need for more configuration which might not be trivial to add in a backwards compatible way...

http/codegen/server.go
raphael: I believe this should behave correctly but haven't tested that myself...

raphael:
Thank you for raising this issue and doing the discovery work! didn't know about the special panic value for breaking chunked responses. This looks good to me, I agree a test would be really nice, maybe a new test case in `server_encode_test.go`? Alternatively it could be a brand new test file `skip_encode_decode_test.go`. The test would generate the code and compare it to golden files content like `server_encode_test.go` does.

http/codegen/server.go
MichaelUrman: Switching to bufio lets us read 1 byte with impunity. Shouldn't need a constant for 1 :)

http/codegen/server.go
MichaelUrman: Switching to bufio avoids the need for a multireader, and thus the edge case I worried about.

MichaelUrman:
It looks like handler_test was a better option for the Handler function. Since the behavior is so finicky, I'd feel a lot safer with some actual tests of that. Does goa have any run-time tests I can pattern off of? And/or do you prefer to stick with just golden output tests?

raphael:
There isn't any runtime test at the moment but not opposed to adding some. An alternative could be to add an example to github.com/goadesign/example that illustrates this use case as the examples do get tested on CI. The last commit looks great.

raphael:
Merging this before I release the next version, we can add more tests later.

PR #3322, opened 2023-08-01
Fix returning correct status code on not hijacked streaming endpoints

### LINKED ISSUE #3054: Streaming: unable to write error status code
We noticed a change in behavior with https://github.com/goadesign/goa/pull/3001: if no results are streamed back but an error is returned, Goa is unable to write the correct status code.

The new call to `w.Write(nil)` has a [side-effect](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/server.go;l=118-121) of writing out the "200 OK" status code:
https://github.com/goadesign/goa/blob/7698ee44c9111e26eb0ce4f3b264af44530caf0c/http/codegen/server.go#L498-L506

So for the non-hijacked case, the subsequent call to `encodeError` will be unable to write out the correct status code, printing an error:
```
http: superfluous response.WriteHeader call
```
The rest of the response body can still be written, fortunately, but the status code will be `200`.

cubic3d:
Fixes #3054

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -496,7 +496,7 @@ func {{ .HandlerInit }}(
 	{{- if not .Redirect }}
 		if err != nil {
 			{{- if isWebSocketEndpoint . }}
-			if _, werr := w.Write(nil); werr == http.ErrHijacked {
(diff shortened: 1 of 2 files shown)


raphael:
Thank you!

PR #3326, opened 2023-08-12
Replace goregen dependency

phroggyy:
The original goregen library has imports that can
cause issues with go mod tidy. There is a new fork of the library as of this year which has removed
superfluous dependencies of goregen.

The new package also has proper versioning and a
go.mod file.


raphael:
This is great, thanks for the PR!

Issue #3316, opened 2023-07-11
Meta openapi:example still generates example if put at API level


antipopp:
As the title says. According to documentation this meta should work at APIExpr level to disable examples for all the attributes. So far this has been working only when declared in an attribute (disabling the specific attribute example).

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = API("calc", func() {
	Title("Calculator Service")
	Description("Service for multiplying numbers, a Goa teaser")
	Server("calc", func() {
		Host("localhost", func() {
			URI("http://localhost:8000")
			URI("grpc://localhost:8080")
		})
	})

	Meta("openapi:example", "false")
})

var _ = Service("calc", func() {
	Description("The calc service performs operations on numbers.")
	Method("multiply", func() {
		Payload(func() {
			Attribute("a", Int, "Left operand")
			Attribute("b", Int, "Right operand")
			Required("a", "b")
		})

		Result(Int)

		HTTP(func() {
			POST("/")
		})
	})

	Files("/openapi.json", "./gen/http/openapi.json")
})
```

Results in
```yaml
openapi: 3.0.3
info:
    title: Calculator Service
    description: Service for multiplying numbers, a Goa teaser
    version: "1.0"
servers:
    - url: http://localhost:8000
paths:
    /:
        post:
            tags:
                - calc
            summary: multiply calc
            operationId: calc#multiply
            requestBody:
                required: true
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/MultiplyRequestBody'
                        example:
                            a: 6825552331577586910
                            b: 1595097090304680186
            responses:
                "200":
                    description: OK response.
                    content:
                        application/json:
                            schema:
                                type: integer
                                example: 8399553735696626949
                                format: int64
                            example: 360622074634248926
    /openapi.json:
        get:
            tags:
                - calc
            summary: Download ./gen/http/openapi.json
            operationId: calc#/openapi.json
            responses:
                "200":
                    description: File downloaded
components:
    schemas:
        MultiplyRequestBody:
            type: object
            properties:
                a:
                    type: integer
                    description: Left operand
                    example: 6322633713974661021
                    format: int64
                b:
                    type: integer
                    description: Right operand
                    example: 3793862871819669726
                    format: int64
            required:
                - a
                - b
tags:
    - name: calc
      description: The calc service performs operations on numbers.
```

antipopp:
fixed in #3330

PR #3330, opened 2023-08-20
meta openapi:example work at APIExpr level too

antipopp:
As mentioned in my issue #3316 this meta doesn't work as the documentation says.

```
// - "swagger:example" DEPRECATED, use "openapi:example" instead
//
// - "openapi:example" specifies whether to generate random example. Defaults to
// true. Applicable to API (applies to all attributes) or individual attributes.
//
//	var _ = API("MyAPI", func() {
//	    Meta("openapi:example", "false")
//	})
```

I've looked into the code for a while and really couldn't find a better way than this. The problem is that examples are generated on the Attribute expression, and the meta will work only there. My solution checks for the API meta during the generation of the openapi schema and sets the random generator seed to nil, and then adding the check to the `Example` method that created the example.

I found the codebase kinda confusing and I won't be surprised if there's a better, cleaner way, but since nobody was looking into it I tried myself.

### DIFF
--- a/expr/random.go
+++ b/expr/random.go
@@ -14,6 +14,8 @@ import (
 //
 // The random values should be consistent in that given the same seed the same
 // random values get generated.
+//
+// Setting the randomizer to nil disables example generation.
 type Randomizer interface {
 	// ArrayLength decides how long an example array will be
 	ArrayLength() int
--- a/http/codegen/openapi/v3/builder.go
+++ b/http/codegen/openapi/v3/builder.go
@@ -30,6 +30,14 @@ func New(root *expr.RootExpr) *OpenAPI {
 		return nil
 	}
 
+	m, ok := root.API.Meta.Last("openapi:example")
+	if !ok {
+		m, ok = root.API.Meta.Last("swagger:example")
+	}
+	if ok && m == "false" {
+		root.API.ExampleGenerator.Randomizer = nil
+	}
+
 	var (
 		bodies, types = buildBodyTypes(root.API)
 
--- a/expr/example.go
+++ b/expr/example.go
@@ -26,6 +26,10 @@ func (a *AttributeExpr) Example(r *ExampleGenerator) any {
 		return ex[len(ex)-1].Value
 	}
 
+	if r.Randomizer == nil {
+		return nil
+	}
+
 	value, ok := a.Meta.Last("openapi:example")
 	if !ok {
 		value, ok = a.Meta.Last("swagger:example")


raphael (commented):
Great first stab! A few things that would make this a bit better:
1. It would suffice to change the generator just once in the `New` function of `builder.go`.
2. The metadata check could simply call `MustGenerate(api.Meta)` instead of testing inline
3. I would suggest setting the `Randomizer` of the `ExampleGenerator` to `nil` instead of the generator itself. The test to skip generation would be changed to `if r.Randomizer == nil`. We could document the behavior in `expr/random.go` (there are currently two possible randomizer types)

antipopp:
Let me know for other suggestions, somehow I totally missed the `New` function while I was traversing the whole flow when looking into it :)

About the inline check, there is already a `MustGenerate` function that checks if the whole schema needs to be generated. I made a new function that checks only for the example meta instead of mixing the two things. Hopefully you meant it in this way.

expr/random.go
```diff
@@ -14,6 +14,9 @@ import (
 //
 // The random values should be consistent in that given the same seed the same
 // random values get generated.
+//
+// The randomizer is set to nil when the user has disabled example generation
```
raphael: Maybe the comment could be tweaked to simply read:
```go
// Setting the randomizer to nil disables example generation.
```

http/codegen/openapi/v3/builder.go
```diff
@@ -30,6 +30,10 @@ func New(root *expr.RootExpr) *OpenAPI {
 		return nil
 	}
 
+	if !mustGenerateRandomExample(root.API.Meta) {
```
raphael: I don't understand why this couldn't just re-use the existing `mustGenerate` function? what am I missing :)

raphael (commented):
Looks good - left one question!

expr/random.go
antipopp: Well that's just what Copilot somehow prompted as I typed "set to nil" haha - I'll update it.

http/codegen/openapi/v3/builder.go
antipopp: I mean, how would you structure it? `mustGenerate` takes any meta expression and checks for the `openapi:generate' tag. Adding an other check for the `openapi:example` tag would break the behaviour in case the example is false and generate is true.

Like this
```go
func mustGenerate(meta expr.MetaExpr) bool {
	m, ok := meta.Last("openapi:generate")
	if !ok {
		m, ok = meta.Last("swagger:generate")
	}
	if ok && m == "false" {
		return false
	}

	example, ok := meta.Last("openapi:example")
	if !ok {
		example, ok = meta.Last("swagger:example")
	}
	if ok && example == "false" {
		return false
	}

	return true
}
```

The only way out I can think is refactoring the function as `mustGenerate(tag string) bool` so you can specify what tag you're referring to.

raphael: Oh duh sorry somehow I managed to miss that `mustGenerate` is not about examples 🤦 . Just in-lining the check in `New` makes sense then.

antipopp: No worries, I've pushed the changes.

raphael:
Thank you!

PR #3344, opened 2023-09-08
chore: change ci process to use matrix for versioning

kevinrobayna:
This PR changes the CI to use the matrix API from GH. (It's currently failing on my fork because I don't have the token for the coverage)

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,66 +1,35 @@
 name: build
 on: [push, pull_request]
 jobs:
-
-  Linux:
-    name: Linux
-    runs-on: ubuntu-latest
+  ci:
+    strategy:
+      fail-fast: true
+      matrix:
+        go: ['1.20', '1.21']
+        os: ['ubuntu-latest', 'windows-latest']
+    runs-on: ${{ matrix.os }}
     steps:
-
-    - name: Set up Go 1.20
-      uses: actions/setup-go@v4.1.0
-      with:
-        go-version: '1.20'
-      id: go
-
-    - name: Check out code into the Go module directory
-      uses: actions/checkout@v3    
-
-    - name: Extract branch name
-      run: echo "::set-output name=branch::$(echo ${GITHUB_REF#refs/heads/})"
-      id: extract_branch
-
-    - name: Add $GOPATH/bin to PATH
-      run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
-      id: setup_path
-
-    - name: Build
-      run: make ci
-      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,66 +1,35 @@
 name: build
 on: [push, pull_request]
 jobs:
-
-  Linux:
-    name: Linux
-    runs-on: ubuntu-latest
+  ci:
+    strategy:
+      fail-fast: true
```
kevinrobayna, line 6: This is something I've taken the liberty to add. This will mean that if any of the N tests you're running fails fast it probably means the others will fail too so there's probably no reason why running them all (this should save computation time and therefore $$ to GOA)

.github/workflows/ci.yml
```diff
@@ -1,66 +1,35 @@
...
-      env:
-        GOA_BRANCH: ${{ steps.extract_branch.outputs.branch }}
\ No newline at end of file
+      - name: Set up Go ${{ matrix.go }}
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go }}
+        id: go
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v4
```
kevinrobayna, line 18: Upgraded to checkout@v4 since i was here.

.github/workflows/ci.yml
```diff
@@ -1,66 +1,35 @@
...
-      run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
-      id: setup_path
-
-    - name: Build
-      run: make ci
-      env:
-        GOA_BRANCH: ${{ steps.extract_branch.outputs.branch }}
\ No newline at end of file
+      - name: Set up Go ${{ matrix.go }}
+        uses: actions/setup-go@v4
```
kevinrobayna, line 13: changed this to use the latest version of setup-go@v4

.github/workflows/ci.yml
```diff
@@ -1,66 +1,35 @@
...
+        id: extract_branch
+      - name: Add $GOPATH/bin to PATH
+        run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+        id: setup_path
+      - name: Build
+        run: make ci
+        env:
+          GOA_BRANCH: ${{ steps.extract_branch.outputs.branch }}
+      - name: Report test-coverage to DeepSource
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
kevinrobayna, line 30: I noticed you didn't do this on windows so I added the check to avoid it

raphael:
Thank you, this is great!

PR #3346, opened 2023-09-09
Switch to chi as default Goa mux

raphael:
chi mux offers a few benefits as compared to `httpmux` including a more active community and the ability to retrieve the pattern that was used to match a route. Various alternatives were considered including Gorilla mux, https://github.com/julienschmidt/httprouter and https://bunrouter.uptrace.dev/guide/golang-router.html.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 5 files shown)


raphael:
After discussions on the slack channel and considering various alternatives [chi](https://github.com/go-chi/chi) seems like a better fit. Closing this PR in favor of https://github.com/goadesign/goa/pull/3346.

PR #3364, opened 2023-09-25
Unescape HTTP path parameters

tchssk:
This is a fix about the change from httptreemux to chi.

httptreemux unescapes http path parameters but chi does not.
This fix restores the previous httptreemux-like behavior.

https://github.com/dimfeld/httptreemux/blob/v5.5.0/unescape_18.go

### DIFF
--- a/http/mux.go
+++ b/http/mux.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"regexp"
 
 	chi "github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


tchssk:
FYI: Due to the change from httptremux to chi, requests with a trailing slash are no longer handled as same as v3.12.4 or earier. I created a  plugin that reproduces the previous httptreemux-like behavior. https://github.com/tchssk/goaplugins/tree/master/autotrailingslash

raphael:
This is great, thank you! I wonder if we should mount this middleware by default: https://pkg.go.dev/github.com/go-chi/chi/middleware#RedirectSlashes

tchssk:
The RedirectSlashes middleware looks good to me.

PR #3384, opened 2023-10-05
Bug fix SkipRequestBodyEncodeDecode

### LINKED ISSUE #3383: Method with SkipRequestBodyEncodeDecode and without payload generates invalid client
### Problem
As title states, if you design a method where you do not want to define a payload but only want the `req io.ReadCloser` in the controller, then `goa gen` generates an invalid client. 

### Example design
```
	Description("The calc service performs operations on numbers")

	// Method describes a service method (endpoint)
	Method("multiply", func() {
		Result(Int)
		HTTP(func() {
			SkipRequestBodyEncodeDecode()
			POST("/")
		})
	})
})
```

### Expected outcome
```
// Multiply calls the "multiply" endpoint of the "calc" service.
func (c *Client) Multiply(ctx context.Context, req io.ReadCloser) (res int, err error) {
	var ires any
	ires, err = c.MultiplyEndpoint(ctx, &MultiplyRequestData{ Body: req })
	if err != nil {
		return
	}
	return ires.(int), nil
}
```

### Actual outcome
```
// Multiply calls the "multiply" endpoint of the "calc" service.
func (c *Client) Multiply(ctx context.Context, , req io.ReadCloser) (res int, err error) {
	var ires any
	ires, err = c.MultiplyEndpoint(ctx, &MultiplyRequestData{ Body: req })
	if err != nil {
		return
	}
	return ires.(int), nil
}
```

As you can see, it contains an additional `, ` in the method parameters.

emilkor1:
This commit fixes a bug in SkipRequestBodyEncodeDecode which occurs in a design without payload.

This commit fixes #3383


raphael:
This is great, thank you!

PR #3366, opened 2023-09-26
Use RedirectSlashes middleware by default Goa mux 

tchssk:
https://github.com/goadesign/goa/pull/3364#issuecomment-1734770313

PR #3385, opened 2023-10-06
Replace RedirectSlashes with smartRedirectSlashes

tchssk:
Related to #3366

RedirectSlashes middleware was not suitable for Goa.

RedirectSlashes redirects a request with a trailing slash to the same path without slash.
It also works for patterns with a trailing slash, so they are no longer accesible.
(@ikawaha told me that. Thank you so much.)

I replaced the middleware with a new one.
It redirects both paths with and withohut a trailing slash according to defined patterns.

### DIFF
--- a/http/mux.go
+++ b/http/mux.go
@@ -8,7 +8,6 @@ import (
 	"regexp"
 
 	chi "github.com/go-chi/chi/v5"
-	"github.com/go-chi/chi/v5/middleware"
 )
 
 type (
@@ -80,9 +79,8 @@ type (
 // NewMuxer returns a Muxer implementation based on a Chi router.
 func NewMuxer() ResolverMuxer {
 	r := chi.NewRouter()
-	// RedirectSlashes must be mounted at the top level of the router.
-	// See. https://github.com/go-chi/chi/blob/1129e362d6cce6e3805e3bc8dfbaeb34b5129789/middleware/strip_test.go#L105-L107
-	r.Use(middleware.RedirectSlashes)
+	// smartRedirectSlashes must be mounted at the top level of the router.
+	r.Use(smartRedirectSlashes)
 	r.NotFound(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		ctx := context.WithValue(req.Context(), AcceptTypeKey, req.Header.Get("Accept"))
 		enc := ResponseEncoder(ctx, w)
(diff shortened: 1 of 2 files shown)


ikawaha:
FYI: @tchssk 


case 1: mux router
```
router := httptreemux.New()
router.GET("/about", pageHandler)
router.GET("/posts/", postIndexHandler)
router.POST("/posts", postFormHandler)
```

case 2: chi router
```
router := chi.NewRouter()
router.Get("/about", pageHandler)
router.Get("/posts/", postIndexHandler)
router.Post("/posts", postFormHandler)
```

case 3: chi + RedirectSlashes
```
router := chi.NewRouter()
router.Use(middleware.RedirectSlashes)
router.Get("/about", pageHandler)
router.Get("/posts/", postIndexHandler)
router.Post("/posts", postFormHandler)
```

| mux  | GET:/about | --- | GET:/posts/ | --- | POST:/posts | --- |
|:---|:---|:---|:---|:---|:---|:---|
| access | GET:/about | GET:/about/ | GET:/posts/| GET:/posts | POST:/posts |POST:/posts/|
| treemux| /about | /about | /posts/ | /posts/ | /posts/ | /posts/ |
| chi | /about | 404 | /posts/ | 405 | /posts| 405 |
| chi+RedirectSlashes| /about | /about | 301 | 301 | /posts | /posts|

What would be the accessibility of this `chi+smrtRedirectSlashes` combination?

tchssk:
If patterns are defined in design as

```go
GET("/about")
GET("/posts/")
POST("/posts")
```

the routings would be as follows

| Request | GET /about | GET /about/ | GET /posts/ | GET /posts | POST /posts | POST /posts/ |
| --- | --- | --- | --- | --- | --- | --- |
| httptreemux | /about | 301 (/about) | /posts/ | 301 (/posts/) | 301 (/posts/) | /posts/ |
| chi | /about | 404 | /posts/ | 405 | /posts| 405 |
| chi+RedirectSlashes | /about | 301 (/about) | 301 (/posts) | 405 | /posts | 301 (/posts) |
| chi+smartRedirectSlashes | /about | 301 (/about) | /posts/ | 301 (/posts/) | /posts | 301 (/posts) |

`POST /posts` and `POST /posts/` are not compatible between httptreemux and smartRedirectSlashes.

In httphtreemux, if a pattern with a trailing slash is added for a method , all other methods for that pattern will also be considered to have a trailing slash. This is not covered by smartRedirectSlashes because it's about routing and needs to be resolved when registering the handlers.

ikawaha:
LGTM, if backward compatibility with treemux is important. 
It may be an option not to use such a redirection at all, since chi's router does only strict routing.

raphael:
Great discussion!
>In httphtreemux, if a pattern with a trailing slash is added for a method , all other methods for that pattern will also be considered to have a trailing slash.

Maybe we have an opportunity to remove this implicit behavior as we move to chi. We do need to consider backwards compatibility though. Here are my thoughts:

* We add the `SmartRedirectSlashes` as a public and well documented function in the Goa `http` package. In particular the header comment could explain that it exists to make the behavior similar to `httptreemux` and that the middleware should be mounted on the mux.
* We do not mount that middleware by default, that is by default the Goa mux uses strict routing.

This way the new behavior is less surprising but users that rely on the implicit redirect can mount the middleware.

tchssk:
That makes sense.
I'll move the middleware under `http` package.

tchssk:
I created new two pull requests.

* #3392
* #3393

PR #3401, opened 2023-10-22
Add CookieSameSite DSL

jaredLunde:
Adds support for the [SameSite](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site-00) cookie attribute available in `net/http` [since Go 1.11](https://github.com/golang/go/commit/3d5703babe9c5344252db3fb8e96f20cd036535a).

I think this covers everything, but let me know if there's anything else I can/need to do.

### DIFF
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -1353,7 +1353,7 @@ const responseT = `{{ define "response" -}}
 			HttpOnly: true,
 			{{- end }}
 			{{- if .SameSite }}
-			SameSite: {{ printf "%q" .SameSite }},
+			SameSite: {{ .SameSite }},
 			{{- end }}
 		})
 		{{- if or $checkNil $initDef }}
--- a/http/codegen/server.go
+++ b/http/codegen/server.go
@@ -1352,6 +1352,9 @@ const responseT = `{{ define "response" -}}
 			{{- if .HTTPOnly }}
 			HttpOnly: true,
 			{{- end }}
+			{{- if .SameSite }}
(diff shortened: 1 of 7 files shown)


raphael (commented):
This looks great! The convention has been to redefine the constants in the DSL package so that the documentation is self-contained, see [formats](https://github.com/goadesign/goa/blob/b65be5cd0b076907179fac071921450c9e6b01e5/dsl/validation.go#L12C1-L54), [HTTP ](https://github.com/goadesign/goa/blob/b65be5cd0b076907179fac071921450c9e6b01e5/dsl/http.go#L12C1-L77), gRPC status codes, or data types for examples. I'll merge once these are added, thank you!

http/codegen/server.go
```diff
@@ -1352,6 +1352,9 @@ const responseT = `{{ define "response" -}}
 			{{- if .HTTPOnly }}
 			HttpOnly: true,
 			{{- end }}
+			{{- if .SameSite }}
+			SameSite: {{ printf "%q" .SameSite }},
```
jaredLunde: Will these generated lines print the enum from the `net/http` package e.g. `http.SameSiteLaxMode`? It just occurred to me that this could be a problem, as `http.SameSite___` are iotas.

jaredLunde: Actually there are a couple of issues with this including the `"%q"` 😅 Will fix. Where can I add tests for the codegen piece?

raphael:
Looks great, thank you!

PR #3409, opened 2023-11-05
Fix generated validation code for gRPC 

tchssk:
The protobuf types do not consider Meta, so the validation code should not too.


raphael:
This is really great, thank you for the nice fix and tests!

PR #3412, opened 2023-11-07
Add openapi:typename Meta support to openapi v2

tchssk:
Related to https://github.com/goadesign/goa/pull/3203


raphael:
This is great, thank you!

PR #3417, opened 2023-11-22
Use sync.Map for http.mux.wildcards

tchssk:
https://github.com/goadesign/goa/pull/3415#issuecomment-1822213529


raphael:
Great fix, thank you!

PR #3419, opened 2023-11-27
Lazy initialization for default Goa mux

tchssk:
https://github.com/goadesign/goa/pull/3415#issuecomment-1821211567

* I tried to implement it using only `sync.Mutex`.
    * `middlewares` is changed at the first call of `Handle()`, so it will be `nil` at the second call.
* `wildcards` is now accessed in the mutex lock, so I reverted #3417.

### DIFF
--- a/http/mux.go
+++ b/http/mux.go
@@ -77,7 +77,7 @@ type (
 		middlewares []func(http.Handler) http.Handler
 		// wildcards maps a method and a pattern to the name of the wildcard
 		// this is needed because chi does not expose the name of the wildcard
-		wildcards sync.Map
+		wildcards map[string]string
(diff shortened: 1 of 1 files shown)


raphael:
This looks great thank you! Reverting the change makes sense. It does mean that the reads are not protected against concurrent writes but that's probably OK given that writes should only occur during initialization and having a lock shared across all requests wouldn't be great from a performance point of view.

tchssk:
Yes. Now `Handle()` is protected with a mutex, so it's impossible that concurrent access to the map inside the method.

tchssk:
Finally it's completed. Would you please release a new version? @raphael

PR #3447, opened 2024-01-12
Properly generate format for primitive types in OpenAPI specs

tchssk:
This fix is for types that do not have format validation, but have some other validations.


raphael:
This is great, thank you!

PR #3437, opened 2023-12-29
Add openapi:generate Meta support to dsl.Attribute

tchssk:
Related to #3405 #3406

PR #3452, opened 2024-01-17
Fix openapi:generate Meta to affect required validations

tchssk:
Related to https://github.com/goadesign/goa/pull/3437


raphael:
Good catch, thank you.

PR #3454, opened 2024-01-21
Generate valid OpenAPI specifications

raphael:
When security requirements are empty (needs to be empty array and not nil). Also take into consideration metadata set at service level to generate summary.


raphael:
Note: DeepSource is incorrect in stating that
```go
scopes := []string{}
```
is equivalent to:
```go
var scopes []string
```
In the former case the slice is initialized and when serialized to JSON produces `[]` while in the latter case it's not and when serialized to JSON produces `nil` which is the bug this PR fixes.

PR #3460, opened 2024-01-23
Use v3 branch for pushes and feature branch for PRs for code coverage

raphael:
…e coverage

### DIFF
--- a/.github/workflows/report-coverage.yml
+++ b/.github/workflows/report-coverage.yml
@@ -12,6 +12,18 @@ jobs:
     if: github.event.workflow_run.conclusion == 'success'
 
     steps:
+      - name: Set ref for PR
+        if: github.event_name == 'pull_request'
+        run: echo "REF=${{ github.head_ref }}" >> $GITHUB_ENV
+      - name: Set ref for Push
+        if: github.event_name == 'push'
+        run: echo "REF=${{ github.ref }}" >> $GITHUB_ENV
+      - name: Check out code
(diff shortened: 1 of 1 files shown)


.github/workflows/report-coverage.yml
```diff
@@ -12,6 +12,18 @@ jobs:
     if: github.event.workflow_run.conclusion == 'success'
 
     steps:
+      - name: Set ref for PR
+        if: github.event_name == 'pull_request'
+        run: echo "REF=${{ github.head_ref }}" >> $GITHUB_ENV
+      - name: Set ref for Push
+        if: github.event_name == 'push'
+        run: echo "REF=${{ github.ref }}" >> $GITHUB_ENV
```
MichaelUrman, line 20: Is there a reason to prefer this over specifying something like `ref: ${{ github.head_ref || github.ref }}` in the checkout action?

raphael:
My understanding is that `head_ref` would also exist in the case of a push but would point to the branch that was just deleted. At least that is what I saw when it was just using `head_ref`.

raphael:
The approach above didn't work... I updated it in #3463 - hard to test this stuff without merging first unfortunately.

MichaelUrman:
Ah, gotcha. I can never keep clear exactly which of those are available when. 👍 

And agreed testing them is a bear. It's almost like they expect you to create throw-away projects, or a mess in your history.

raphael:
It's still not working properly ugh...

Issue #3361, opened 2023-09-20
Meta("struct:pkg:path", "foo/bar") doesn't work.


Albert-Coding:
When using `Meta("struct:pkg:path", "foo/bar")`, code produced doesn't compile.

Example code:
```
package design

import . "goa.design/goa/v3/dsl"

var _ = API("admin_and_internal_grpc_api", func() {
	Server("admin_grpc_server", func() { Services("admin_grpc_service") })
	Server("internal_grpc_server", func() { Services("internal_grpc_service") })
})

var _ = Service("admin_grpc_service", func() {
	Method("CreateThing", func() {
		Payload(func() { Field(1, "event", Event) })
		Result(Boolean, "success")
		GRPC(func() {})
	})
})

var _ = Service("internal_grpc_service", func() {
	Method("CreateThing", func() {
		Payload(func() { Field(1, "event", Event) })
		Result(Boolean, "success")
		GRPC(func() {})
	})
})

var Event = Type("Event", func() {
	Meta("struct:pkg:path", "common/payloads")
	Field(1, "associations", Associations)
})

var Associations = Type("Associations", func() {
	Meta("struct:pkg:path", "common/payloads")
	Field(1, "resources", ArrayOf(Resource))
})

var Resource = Type("Resource", func() {
	Meta("struct:pkg:path", "common/data")
	Field(1, "name", String)
})

```

Produces code like:
```
package payloads

type Event struct {
	Associations *payloads.Associations
}
```

and 

```
package payloads

type Associations struct {
	Resources []*data.Resource
}
```

Albert-Coding:
Poke

Albert-Coding:
Poke

tchssk:
There are two issues:

1. Unneeded package name prefix for a qualified identifier (e.g. `payloads.Associations`) (resolved by #3469)
2. Missing imports (e.g. `data`)

The second issue is unclear whether it is a specification or a bug. [The Meta DSL's godoc](https://pkg.go.dev/goa.design/goa/v3@v3.14.6/dsl#Meta) says:

> Note: If that meta tag is used more that once in the same design, but with different values in the meta statement (ex. one type has Meta("struct:pkg:path", "types1") and another has Meta("struct:pkg:path", "types2")) then those two types cannot both contain a field of the same user type. For the same reason, you may not set a different custom package in a user type than the one set on a containing user type. 

This specification can be reproduced by changing design as follows.

```diff
[...]

PR #3469, opened 2024-01-28
Treat nested paths for struct:pkg:path Meta

tchssk:
https://github.com/goadesign/goa/issues/3361

3361 has two issues:

1. Unneeded package name prefix for a qualified identifier (e.g. `payloads.Associations`)
2. Missing imports (e.g. `data`)

This pull request only fixes the first issue.

### DIFF
--- a/codegen/scope.go
+++ b/codegen/scope.go
@@ -2,6 +2,7 @@ package codegen
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


raphael:
This is great, thank you. And I finally managed to fix the DeepSource action...

PR #3474, opened 2024-02-01
Validate Array and Map for struct:pkg:path Meta

tchssk:
https://github.com/goadesign/goa/issues/3361


tchssk:
Would you release a new version? @raphael 
(It might be a good idea to add an auto release action)

raphael:
yes I'll release a new version soon!

PR #3480, opened 2024-02-20
Add openapi:json:indent Meta

tchssk:
This feature makes it possible to reduce the differences in regenerated JSON.


raphael:
This is great, thank you!

Issue #3481, opened 2024-02-26
Goa panics when query params have unclosed square brackets


phroggyy:
## Description

When a user makes a request with mismatched square brackets, Goa panics due to out-of-range slice bounds.

## Reproduction

Given the following:
```go
var DynamicValueFilter = MapOf(String, MapOf(String, ArrayOf(String,
	func() { Example("value") }),
	func() { Example(map[string][]string{"field": {"value"}}) }),
	func() { Example(map[string]map[string][]string{"parent": {"child": {"value"}}}) })

Payload(func() {
	Attribute("custom_field", DynamicValueFilter, "Filter on an incident custom field. Custom field ID should be sent, followed by the operator and values. Accepted operator will depend on the custom field type.", func() {
		Example(map[string]map[string][]string{"01GBSQF3FHF7FWZQNWGHAVQ804": {"one_of": {"01GBSQF3FHF7FWZQNWGHAVQ804", "01ET65M7ZARSFZ6TFDFVQDN9AA"}}})
	})
})
```

Below is an example of a failing request:
```
GET /example?custom_field[0000004JFGGSTWX5A814HVMH4Y[one_of]=test
```
_(note how there is no closing `]` before `[one_of]`)_

If user passes an unmatched set of square brackets, as above, Goa will panic due to
```
runtime error: slice bounds out of range [:-1]
```

I believe this is due to the following code not handling the case of unmatched brackets: https://github.com/goadesign/goa/blob/cdc84cd229ccf936f91ca725bf5f5e1a7ee0193e/http/codegen/templates/partial/query_map_conversion.go.tpl#L7

raphael:
That makes sense, would you be able to put together a minimal but complete design that reproduces the issue?

raphael:
This is fixed in v3.15.1!

PR #3488, opened 2024-03-08
Return a validation error when query string is malformed

raphael:
for map parameters.


raphael:
Fix https://github.com/goadesign/goa/issues/3481

PR #3497, opened 2024-03-18
adds openapi deprecated support as a dsl.

### LINKED ISSUE #3449: meta for externalDocs and deprecated?
I can see the Operations object has these fields, but I've been playing around with the DSL to try and set these fields but have not been able to get it to work.

How does one set them using Meta? Would be nice to have a good example

xlanor:
added deprecated support as a dsl for both v2 and v3 on HTTP routes.

Fixes #3449

### DIFF
--- a/dsl/method.go
+++ b/dsl/method.go
@@ -37,16 +37,16 @@ func Method(name string, fn func()) {
 	s.Methods = append(s.Methods, ep)
 }
 
-// Deprecated must appear in a Method expression.
+// Deprecated marks HTTP routes as deprecated in the generated OpenAPI specifications.
 //
-// Method takes no arguments. If called, it assumes that
-// the route is to be marked as deprecated.
-//
+// Deprecated must appear in a Method HTTP expression.
+// 
+// Deprecated takes no argument.
 // Example:
 //
 //	   Method("add", func() {
 //	    HTTP(func() {
-//				GET("/")
+//			GET("/")
 //				Deprecated()
 //			})
 //	   })
--- a/dsl/method.go
+++ b/dsl/method.go
@@ -36,3 +36,25 @@ func Method(name string, fn func()) {
 	ep := &expr.MethodExpr{Name: name, Service: s, DSLFunc: fn}
 	s.Methods = append(s.Methods, ep)
 }
+
+// Deprecated must appear in a Method expression.
+//
+// Method takes no arguments. If called, it assumes that
+// the route is to be marked as deprecated.
+//
+// Example:
+//
+//	   Method("add", func() {
+//	    HTTP(func() {
+//				GET("/")
+//				Deprecated()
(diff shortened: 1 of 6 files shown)


dsl/method.go
```diff
@@ -36,3 +36,25 @@ func Method(name string, fn func()) {
 	ep := &expr.MethodExpr{Name: name, Service: s, DSLFunc: fn}
 	s.Methods = append(s.Methods, ep)
 }
+
+// Deprecated must appear in a Method expression.
+//
+// Method takes no arguments. If called, it assumes that
```
raphael: Maybe something like this is slightly clearer and follows a bit more the convention (explanation then when it can appear then arguments).
```go
// Deprecated marks HTTP routes as deprecated in the generated OpenAPI specifications.
//
// Deprecated must appear in a Method HTTP expression.
// 
// Deprecated takes no argument.
```

dsl/method.go
```diff
@@ -36,3 +36,25 @@ func Method(name string, fn func()) {
...
+// the route is to be marked as deprecated.
+//
+// Example:
+//
+//	   Method("add", func() {
+//	    HTTP(func() {
+//				GET("/")
```
raphael: Great example, might be good to fix the indentation (use  4 spaces)

http/codegen/openapi/v2/builder.go
```diff
@@ -639,7 +639,7 @@ func buildPathFromExpr(s *V2, root *expr.RootExpr, h *expr.HostExpr, route *expr
 			}
 			requirements[i] = requirement
 		}
-
+		_, is_deprecated := endpoint.MethodExpr.Meta.Last("openapi:deprecated")
```
raphael: nitpick :) [idiomatic Go](https://go.dev/doc/effective_go#names) does not use underscores in variable names so maybe just `deprecated`? That would also be consistent with the field name.

http/codegen/openapi/v3/builder.go
```diff
@@ -339,7 +338,9 @@ func buildOperation(key string, r *expr.RouteExpr, bodies *EndpointBodies, rand
 			break
 		}
 	}
-
+	
+	// An endpoint may be marked as deprecated. if the openapi:deprecated tag is present, we populate it to true
+	_, is_deprecated := e.Meta.Last("openapi:deprecated")
```
raphael: Same comment as above

raphael (commented):
This looks great! I left a few nitpicks that would be nice to fix before merging, thank you!

xlanor:
Really not sure what's going on here because I've switched indentation to spaces and it shows up funny in github.

image

xlanor:
Should be good to go now once CI passes.

raphael:
This is great, thank you!

PR #3502, opened 2024-04-09
Fix eval.IncompatibleDSL() to not show internal DSL

tchssk:
The `caller()` function is from v1, but had to be modified for v3.

https://github.com/goadesign/goa/blob/v1/dslengine/runner.go#L419-L429


raphael:
Very nice, thank you!

PR #3506, opened 2024-04-11
Save r.URL.Query() in a variable

### LINKED ISSUE #3505: Generated server code parses query parameters repeatedly
When goa generates server code for query parameters, it checks each parameter one after another with code that looks something like this:

```go
		advRaw := r.URL.Query().Get("adv")
		if advRaw != "" {
			adv = &advRaw
		}
		afRaw := r.URL.Query().Get("af")
		if afRaw != "" {
			af = &afRaw
		}
		apbndlRaw := r.URL.Query().Get("apbndl")
		if apbndlRaw != "" {
			apbndl = &apbndlRaw
		}
```

The problem with this code, is that `r.URL.Query()` is called repeatedly. Each call parses `r.URL.RawQuery` into a `url.Values` map. One values is pulled from the map, then the map is left to be garbage collected. It'd be much better to parse the query parameters once and store the result in a variable to be referenced for each query parameter to check.

In a test on one of our projects on an endpoint with many query parameters defined, this resulted in 24.11GB (41.4%) of heap allocations (using the alloc profile) coming from calls to `url.(*URL).Query` within the generated `server.DecodeXXXXRequest.func1`.

duckbrain:
If there are multiple query parameters, save the call to Query() in a variable.

Prevents r.URL.Query() from being called multiple times for a handling a single request. Prevents unneeded churn on generated code by only applying that change to code that generates mutliple query parameters.

Closes #3505

### DIFF
--- a/http/codegen/templates/partial/request_elements.go.tpl
+++ b/http/codegen/templates/partial/request_elements.go.tpl
@@ -42,15 +42,20 @@
 		{{- end }}
 {{- end }}
 
+{{- $qpVar := "r.URL.Query()" }}
+{{- if gt (len .QueryParams) 1 }}
+{{- $qpVar = "qp" }}
+qp := r.URL.Query()
+{{- end }}
 {{- range .QueryParams }}
 	{{- if and (or (eq .Type.Name "string") (eq .Type.Name "any")) .Required }}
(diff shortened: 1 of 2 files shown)


raphael:
This is great, thank you!

PR #3509, opened 2024-04-16
Allow disabling version string in code generated header

davideme:
When Updating goa version it's hard to detect the files that are only changing the version number and the ones that are changing code.
Screenshot 2024-04-16 at 17 45 42

I'm proposing adding a Meta Field "goa:version:disable" that would disable the version for this specific section.

```go
	Meta("goa:version:disable", "true")
```

Remarks:
 - I can change any naming, I'm more interested in the functionality.


cmd/goa/gen.go
davideme, line 1: **note:** to avoid adding parameter to the command line, I kept this only header without configuration. And will still keep the version numbering.

codegen/header.go
```diff
@@ -5,13 +5,17 @@ import (
...
+	if !disableVersion {
+		toolVersion = goa.Version()
+	}
 	return &SectionTemplate{
 		Name:   "source-header",
 		Source: headerT,
 		Data: map[string]any{
 			"Title":       title,
-			"ToolVersion": goa.Version(),
+			"ToolVersion": toolVersion,
```
raphael: Just to be consistent maybe this should be `disableVersion` and the test in the template should be `{{ if not .DisableVersion }}`?

codegen/sections_test.go
```diff
@@ -68,6 +68,16 @@ import (
 package testpackage
 
 `, goa.Version())
+		titleHeaderDisableVersion = `// Code generated by goa, DO NOT EDIT.
```
raphael: Maybe we call this `titleHeaderVersionDisabled` to make it a qualified name?

codegen/service/client.go
```diff
@@ -27,7 +27,9 @@ func ClientFile(_ string, service *expr.ServiceExpr) *codegen.File {
 			codegen.GoaImport(""),
 		}
 		imports = append(imports, svc.UserTypeImports...)
-		header := codegen.Header(service.Name+" client", svc.PkgName, imports)
+		value, ok := service.Meta.Last("goa:version:disable")
+		disableVersion := ok && value == "true"
+		header := codegen.Header(service.Name+" client", svc.PkgName, imports, disableVersion)
```
raphael: I'm wondering if we wouldn't be better of passing the metadata object to `codegen.Header` and moving the logic that computes `disableVersion` there:
* This would remove the need for having this computation spread everywhere
* This would allow for other metadata based tweaks in the future when generating file headers

raphael (commented):
Thank you, this looks great! I left a few suggestions.

davideme:
@raphael all suggestions applied. Happy to see it's not far off what was expected.

davideme:
Any update ?

raphael:
Sorry for the delay, changes look great! Just one last question: I see a number of places where `nil` is given as argument for the metadata. This means that these files will still have the version number in the header, is that what we want?

codegen/header.go
```diff
@@ -1,19 +1,23 @@
...
 
 import (
+	"goa.design/goa/v3/expr"
 	goa "goa.design/goa/v3/pkg"
 )
 
 // Header returns a Go source file header section template.
-func Header(title, pack string, imports []*ImportSpec) *SectionTemplate {
+func Header(title, pack string, imports []*ImportSpec, meta expr.MetaExpr) *SectionTemplate {
+	value, ok := meta.Last("goa:version:disable")
```
tchssk: `Meta("goa:version:generate", "false")` might be better.
It's a similar syntax to `Meta("openapi:generate", "false").`

davideme: Done

davideme:
@tchssk I update the Meta name.
@raphael I review all the calls to Header, 
- The ones that have a service associated are called with `.Meta`
- The ones that are in example files that don't have service associated are called with nil, happy to bring another improvement in a subsequent PR if you have any suggestions on how to pass it.
- The one in cmd/goa/gen.go could be provided by command line params but it felt like an additional feature not part of this. And could be confusing as it's not configurable via Meta tags. 
  - https://github.com/goadesign/goa/blob/9466a07e6f00ef7d348ada937fa4e3d539c1fc49/cmd/goa/gen.go#L142

raphael:
The changes look great, thank you for the contribution!

tchssk:
Could you please wait a moment?
I'm worried about backward compatibility.

davideme:
@tchssk let me know which kind of scenario you are thinking? Having the meta tag in a previous version of goa?

davideme:
Before merging, we are doing some additional testing in our own codebase of around 600 goa generated files.

tchssk:
Adding arguments to `codegen.Header()` breaks compatibility. Normally this function is only used internally in Goa, but I'm wondering if there is another way to solve the problem if possible.

* Adding a cli flag
* Implement as a plugin

raphael:
> Adding arguments to `codegen.Header()` breaks compatibility. Normally this function is only used internally in Goa, but I'm wondering if there is another way to solve the problem if possible.
> 
> * Adding a cli flag
> * Implement as a plugin

Good point! A third option could be to make the extra parameter optional (e.g. using [the functional options pattern](https://golang.cafe/blog/golang-functional-options-pattern.html)). We've avoided adding CLI flags in the past and I'm not sure this would be worth the trade-off.

tchssk:
I prefer plugins to keep the core simple, but optional parameters aren't bad either.

davideme:
After testing extensively in our code base,  using Meta tag seems impractical.  it will fill the design files with tag for each kind of files. 
I'll explore the plugin way.

tchssk:
I created a plugin. Here you are.
https://github.com/tchssk/goaplugins/tree/master/goaversionremover

PR #3513, opened 2024-04-24
Handle unsupported request content type

raphael:
Return 415 Unsupported Media Type when the request decoder does not support the request content type.

### DIFF
--- a/http/encoding.go
+++ b/http/encoding.go
@@ -11,6 +11,8 @@ import (
 	"mime"
 	"net/http"
 	"strings"
+
+	goa "goa.design/goa/v3/pkg"
 )
 
 const (
@@ -23,6 +25,13 @@ const (
 	// response Content-Type header when explicitly set in the DSL. The value
 	// may be used by encoders to set the header appropriately.
 	ContentTypeKey
+)
+
+const (
+	// ErrUnsupportedMediaTypeName is the name of the Goa service error returned
+	// by the built-in decoder when the content type of the request body is not
+	// supported.
+	ErrUnsupportedMediaTypeName = "ErrUnsupportedMediaType"
 )
 
 type (
@@ -80,7 +89,7 @@ func RequestDecoder(r *http.Request) Decoder {
 	case "text/html", "text/plain":
 		return newTextDecoder(r.Body, contentType)
 	default:
-		return json.NewDecoder(r.Body)
+		return newUnsupportedDecoder(contentType)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


http/encoding.go
```diff
@@ -25,6 +27,13 @@ const (
 	ContentTypeKey
 )
 
+const (
+	// ErrUnsupportedMediaTypeName is the name of the Goa service error returned
+	// by the built-in decoder when the content type of the request body is not
+	// supported.
+	ErrUnsupportedMediaTypeName = "ErrUnsupportedMediaType"
+)
+
```
tchssk: Should it be defined in `pkg/error.go`?

```go
const (
	UnsupportedMediaType = "unsupported_media_type"
)

func UnsupportedMediaTypeError(ct string) error {
	return PermanentError(UnsupportedMediaType, "unsupported media type %s", ct)
}
```

It seems better to define the error name in snake case like the existing definitions.

* `missing_payload`
* `decode_payload`
* `invalid_field_type`
* `missing_field`
* `invalid_enum_value`
* `invalid_format`
* `invalid_pattern`
* `invalid_range`
* `invalid_length`

http/error.go
```diff
@@ -81,6 +81,9 @@ func (resp *ErrorResponse) MarshalXML(e *xml.Encoder, _ xml.StartElement) error
 // error. This method is used by the generated server code when the error is not
 // described explicitly in the design.
 func (resp *ErrorResponse) StatusCode() int {
+	if resp.Name == ErrUnsupportedMediaTypeName {
+		return http.StatusUnsupportedMediaType
+	}
```
tchssk, line 86: It seems that 400 is returned instead of 415 because the error name is overwritten by `DecodePayloadError()`.

https://github.com/goadesign/goa/blob/feature/unsupported_media_type/http/codegen/templates/request_decoder.go.tpl#L25

 If the error is `ServiceError`, it must be returned as is.

```go
var gerr *goa.ServiceError
if errors.As(err, &gerr) {
	return nil, err
}
```

http/encoding.go
raphael: Thank you for the comment! Yes I agree that the name should be snake case for consistency. `pkg/error.go` contains service level errors while this is a HTTP transport level error so it might be better to keep this error definition in the `http` package.

http/error.go
raphael: Great catch! I have modified the generated code to follow the suggestion above.

http/encoding.go
tchssk: `MissingPayloadError()` and `DecodePayloadError()` are only used for HTTP, so aren't they HTTP transport level errors?

raphael: Makes sense, I moved the error definition to `pkg/error.go`.

tchssk (approved):
LGTM!

PR #3515, opened 2024-05-06
 Fix OpenAPI v2 to not merge service params into endpoint headers

tchssk:
v3 is correct so I fixed v2 accordingly.


raphael:
This is great, thank you!

tchssk:
It's ready to merge.

PR #3512, opened 2024-04-21
Add eval.TooManyArgError()

tchssk:

PR #3520, opened 2024-05-16
Add test for eval.TooManyArgError()

tchssk:
Related to https://github.com/goadesign/goa/pull/3512

### DIFF
--- a/dsl/result_type.go
+++ b/dsl/result_type.go
@@ -99,6 +99,7 @@ func ResultType(identifier string, args ...any) *expr.ResultTypeExpr {
 				}
 				if len(args) > 2 {
 					eval.TooManyArgError()
+					return nil
 				}
 			}
 		}
--- a/dsl/server.go
+++ b/dsl/server.go
@@ -57,15 +57,15 @@ import (
 //	    })
 //	})
 func Server(name string, fn ...func()) *expr.ServerExpr {
+	server := &expr.ServerExpr{Name: name}
 	if len(fn) > 1 {
 		eval.TooManyArgError()
-		return nil
+		return server
 	}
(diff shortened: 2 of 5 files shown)


dsl/result_type.go
```diff
@@ -99,6 +99,7 @@ func ResultType(identifier string, args ...any) *expr.ResultTypeExpr {
 				}
 				if len(args) > 2 {
 					eval.TooManyArgError()
+					return nil
```
raphael, line 102: The value returned by this function might be used to do validations later on so the idea is to not return nil as the validation code assumes there's always an expression (which could be invalid). So we shouldn't return here - or return a zero result expression.

dsl/server.go
```diff
@@ -59,6 +59,7 @@ import (
 func Server(name string, fn ...func()) *expr.ServerExpr {
 	if len(fn) > 1 {
 		eval.TooManyArgError()
+		return nil
```
raphael: same as above

raphael (commented):
This is great! I left a couple of comments that might be good to address, thank you!

dsl/result_type.go
tchssk: There are already two lines that return nil in ResultType. Are these wrong?

https://github.com/goadesign/goa/blob/78b71ce62a436f21023b685eb2d15c90ac1db7a4/dsl/result_type.go#L65-L69
https://github.com/goadesign/goa/blob/78b71ce62a436f21023b685eb2d15c90ac1db7a4/dsl/result_type.go#L109-L116

raphael: Good point, looks like `Type` can also return `nil` so this case is probably safe.

PR #3526, opened 2024-06-04
Use eval.TooManyArgError() more

tchssk:
Related to
* https://github.com/goadesign/goa/pull/3512
* https://github.com/goadesign/goa/pull/3520


raphael:
This is great, thank you!

PR #3536, opened 2024-06-17
Add test for eval.InvalidArgError()

tchssk:


raphael:
Nice catch, thank you!

Issue #3522, opened 2024-05-22
Allow implementing `io.WriterTo` for `SkipResponseBodyEncodeDecode` responses.


duckbrain:
When a server uses `SkipResponseBodyEncodeDecode`, the response must be of the type `io.ReadCloser`. In most cases I've come across, an `io.Writer` would be more ergonomic to use. I often find myself needing to use `io.Pipe` to create an `io.Reader` I can write to. It's made worse by the fact that it's abstracting the `http.ResponseWriter` and pushing it through a `bufio.Reader`, so there's lots of extra copying going on.

I propose to have the generated code check if the `io.ReadCloser` implements `io.WriterTo`, and if it does, call that instead.

Additionally, it'd be nice to have a helper in the goa package to implement the interface from an `io.WriterTo`, to make it easier to switch to, eg: If used as the response to a handler, the `Read` would never be called, and the pipe would never be created.

```go
func SkipResponseWriter(wt io.WriterTo) io.ReadCloser {
	return writerToReaderAdapter{wt, nil}
}

type writerToReaderAdapter struct {
	io.WriterTo
	pr *io.PipeReader
}

func (a *writerToReaderAdapter) initPipe() {
	if a.pr != nil {
		return
	}
	r, w := io.Pipe()
	go func() {
		_, err := wt.WriteTo(w)
		w.CloseWithError(err)
	}()
	a.pr = r
}

func (a *writerToReaderAdapter) Read(b []byte) (n int, err error) {
	a.initPipe()
	return a.pr.Read(b)
}

func (a *writerToReaderAdapter) Close() error {
	a.initPipe()
	return a.pr.Close()
}
```

---

As an alternate suggestion, you could have some way in the DSL that changes the expected type.

raphael:
That makes sense, a couple of thoughts: We'd have to make sure the documentation makes it clear that the generated code supports this alternative. Maybe the generated comment could indicate that this is the case. In the high level implementation above should we use a `sync.Once` to initialize the pipe in a way that is goroutine safe? We also might want to update the [upload_download](https://github.com/goadesign/examples/tree/master/upload_download) example to illustrate the use of `SkipResponseWriter`.

duckbrain:
I went to update the `upload_download` example, then I realized it already will use this new code path, since it's returning an `*os.File` for the `io.ReadCloser` implementation. :smile:

I could contrive an alternative example that reads a file line by line and streams JSON for each line with line length, etc.

raphael:
[...]

PR #3537, opened 2024-06-18
SkipResponseWriter

duckbrain:
Implementation for #3522

Since the tests, don't include the comment, here's the result:

```go
// Service is the admin service interface.
type Service interface {
	// Assets implements assets.

	// If body implements [io.WriterTo], that implementation will be used instead.
	// Consider [goa.design/goa/v3/pkg.SkipResponseWriter] to adapt existing
	// implementations.
	Assets(context.Context, *AuthenticatedRequest) (res *CustomJSONResponse, body io.ReadCloser, err error)
}
```

Note: Tried this out in one of my projects and added `WriterToFunc` to the exported API as well, since I found myself wanting it for most of the handlers I was rewriting. Allows you to use a closure and avoids needing to manually keep track of bytes written, while still preserving the `io.WriterTo` interface. Useful for times when you're passing the writer to something else that doesn't report how many bytes are written like `json.Encoder` or `xml.Encoder`.


http/codegen/templates/server_handler_init.go.tpl
```diff
@@ -84,6 +84,22 @@ func {{ .HandlerInit }}(
...
+			if err != nil {
+				if n == 0 {
+					if err := encodeError(ctx, w, err); err != nil {
+						errhandler(ctx, w, err)
+					}
+				} else {
+					if f, ok := w.(http.Flusher); ok {
+						f.Flush()
+					}
+					panic(http.ErrAbortHandler) // too late to write an error
```
raphael, line 98: I wonder if it might not be better to call `errhandler` here instead of panicking. Typical implementations notify the observability stack of the problem (log, error span, event etc.).

pkg/skip_response_writer.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type writerToReaderAdapter struct {
+	io.WriterTo
+	prOnce sync.Once
+	pr     *io.PipeReader
+}
+
+func (a *writerToReaderAdapter) initPipe() {
+	if a.pr != nil {
```
raphael: Is this test necessary?

raphael (commented):
This is really good! I left a couple of comments.

pkg/skip_response_writer.go
duckbrain: Nope, that's a holdover from before switching to sync.Once.

http/codegen/templates/server_handler_init.go.tpl
duckbrain: I was a bit torn there. Decided to keep the semantics with the non-io.WriterTo version, which also panics. Should probably change both places if it's going to change, but this also feels like a nice bit of code to pull out into a helper for the generated code to call. 

```go
		if wt, ok := o.Body.(io.WriterTo); ok {
			n, err := wt.WriteTo(w)
			if err != nil {
				if n == 0 {
					if err := encodeError(ctx, w, err); err != nil {
						errhandler(ctx, w, err)
					}
				} else {
					if f, ok := w.(http.Flusher); ok {
						f.Flush()
					}
					panic(http.ErrAbortHandler) // too late to write an error
				}
			}
			return
		}
		// handle immediate read error like a returned error
		buf := bufio.NewReader(o.Body)
		if _, err := buf.Peek(1); err != nil && err != io.EOF {
			if err := encodeError(ctx, w, err); err != nil {
				errhandler(ctx, w, err)
			}
			return
		}
		if err := encodeResponse(ctx, w, o.Result); err != nil {
			errhandler(ctx, w, err)
			return
		}
		if _, err := io.Copy(w, buf); err != nil {
			if f, ok := w.(http.Flusher); ok {
				f.Flush()
			}
			panic(http.ErrAbortHandler) // too late to write an error
		}
```

raphael: Ah good point, hmm might as well stay consistent in both cases and stick with panic then - thanks for bringing this up!

raphael:
Thank you for the great contribution! I was wondering if you'd be up to updating the [upload_download](https://github.com/goadesign/examples/tree/master/upload_download) example to take advantage of this new feature after we release it?

PR #3542, opened 2024-06-27
Use errors.As()

tchssk:


eval/context.go
```diff
@@ -134,7 +135,11 @@ func (c *DSLContext) Record(err *Error) {
 	if c.Errors == nil {
 		c.Errors = MultiError{err}
 	} else {
-		c.Errors = append(c.Errors.(MultiError), err)
```
raphael, line 137: I wonder if in this case this is necessary? The type of the `Errors` field is controlled by Goa and guaranteed to be `MultiError`, see https://github.com/goadesign/goa/blob/v3/eval/context.go#L16

eval/eval.go
```diff
@@ -71,14 +72,20 @@ func Execute(fn func(), def Expression) bool {
 	}
 	var startCount int
 	if Context.Errors != nil {
-		startCount = len(Context.Errors.(MultiError))
+		var e MultiError
```
raphael: Same here, not sure this is needed.

eval/eval.go
```diff
@@ -71,14 +72,20 @@ func Execute(fn func(), def Expression) bool {
...
+		if errors.As(Context.Errors, &e) {
+			startCount = len(e)
+		}
 	}
 	Context.Stack = append(Context.Stack, def)
 	fn()
 	Context.Stack = Context.Stack[:len(Context.Stack)-1]
 	var endCount int
 	if Context.Errors != nil {
-		endCount = len(Context.Errors.(MultiError))
```
raphael, line 81: Also probably not needed

pkg/error.go
```diff
@@ -285,8 +285,8 @@ func newError(name string, timeout, temporary, fault bool, format string, v ...a
 }
 
 func asError(err error) *ServiceError {
-	e, ok := err.(*ServiceError)
-	if !ok {
+	var e *ServiceError
+	if !errors.As(err, &e) {
```
raphael, line 289: Should this use `Is`?

raphael (commented):
Thank you this is great! There is one case where we may not need to use `As`.

pkg/error.go
tchssk: This should be `As`. The Go Blog says

https://go.dev/blog/go1.13-errorse

> In the simplest case, the errors.Is function behaves like a comparison to a sentinel error, and the errors.As function behaves like a type assertion

I added test for `asError()`. It fails if you replace `As` with `Is`.

https://github.com/goadesign/goa/pull/3542/commits/a2d4dc8190655ea688fa274c6977e5f44a5ac5cd

eval/context.go
tchssk: Yes, but it can be a different error type in the Go type system. For clarity, how about changing `DSLContext.Errors` to `MultiError`?

pkg/error.go
raphael: Makes sense, thanks for the clarification!

eval/context.go
raphael: Yes making that change makes sense

tchssk: f2d8d28c6419dfb471296a2030304a0f05c39d8c

raphael:
Thank you!

PR #3544, opened 2024-07-02
Use fmt.Errorf("%w") for error wrapping

tchssk:


raphael:
This is great, thank you!

tchssk:
The v3.17.0 tag has been created, but not yet released. Are you planning to do it later?

raphael:
There's an issue with it (the generated example code won't compile), I'm fixing it and will release 3.17.1

PR #3548, opened 2024-07-06
Use errorlint from golangci-lint

tchssk:
Related to
* https://github.com/goadesign/goa/pull/3541
* https://github.com/goadesign/goa/pull/3542
* https://github.com/goadesign/goa/pull/3544


raphael:
Thank you!

Issue #3560, opened 2024-07-15
 OpenAPI schema for `Any` DSL type assumes nested serialization into a string


kjedruczyk:
In the GOA DSL, the `Any` type is described as:

```go
// Any is the type for an arbitrary JSON value (any in Go).
Any = expr.Any
```

This is quite useful interface to capture schemaless JSON values as `any` type in GO, while also easily allowing us to control how JSON responses would be formed using standard go JSON serialization tags. In other words, I'd like to be able to catch/emit an arbitrary JSON value in an attribute declared like this:

```go
Attribute("any", Any)
```

The trouble is that GOA openapi generator decides `Any` type should be encoded as serialized binary string. This is arguably inconsistent with the idea that `Any` represents JSON value (it now is just a string that the client would have to explicitly parse):

```yaml
any:
    type: string
    format: binary
```

Note, that the `go` code doesn't seem to do or expect any kind of nested binary string encoding/decoding. The expectation to handle the "Any JSON" encoding/decoding seems to be implicitly passed onto the client in this scenario.

It is also inconsistent with handling of `Any` type in different contexts. Consider `MapOf(String,Any)`, which will generate following openapi type (as I would expect):

```yaml
mapofstringany:
    type: object
    additionalProperties: true
```

In this case, an arbitrary `JSON` object will have `Any`-typed attributes passed to `go` layer directly as `any` go type without collapsing these attributes into string/binary encoding.

Can we make GOA, by explicit DSL support if needed, emit `AnyValue` or equivalent `oneOf` schema for `Any` DSL type? (I'd be happy to submit a PR if we agree on an approach).

kjedruczyk:
Here's an example design to try and illustrate the issue more precisely:

```go
package design

import (
	. "goa.design/goa/v3/dsl"
)

var _ = API("any-test", func() {
	Server("srv", func() {
		Services("any-test")
	})
})

var TestAny = Type("TestAny", func() {
	Attribute("AnyAttribute", Any)
	Attribute("MapOfAny", MapOf(String,Any))
})

var _ = Service("any-test", func() {
	Method("test", func() {
		Payload(TestAny)
		Result(TestAny)
		HTTP(func() {
			GET("/")
		})
	})
})
```

The above, will give us following schema in the OpenAPI3.yaml (I removed examples for clarity):

```yaml
components:
    schemas:
        TestRequestBody:
            type: object
            properties:
                AnyAttribute:
                    type: string
                    format: binary
                MapOfAny:
[...]

PR #3561, opened 2024-07-18
Handle Any type as a typeless schema in OpenAPI

tchssk:
https://github.com/goadesign/goa/issues/3560#issuecomment-2234991792


raphael:
This is great thank you! Just a note that this is a potential breaking change. It is the more correct approach so seems worth merging but let's keep an eye out for feedback.

PR #3580, opened 2024-08-25
Generate validation code for required attributes in inline struct http bodies

tchssk:
It looks the validation code is missing.
I tried to fix the code but it might not be the best way.


raphael:
This change makes sense, good catch!

PR #3585, opened 2024-09-05
DO NOT REVIEW: Replace w.Set() Header with w.Add Header to be able to set cookies in…

shashank-harness:
… response


raphael:
@shashank-harness are you still planning on working on this or should we close it?

PR #2856, opened 2021-05-25
Add http.Replace() and use it for file servers

tchssk:
This is a fix of #2839.

I'm glad if you release v3.4.3 after merging this. Thank you.

PR #3588, opened 2024-09-10
Refactor file server implementation

raphael:
So that it doesn't require introducing a `Replace` function. The new implementation is more robust and properly handles cases where the file path is deep and differs from the mount path.


tchssk:
This code cannot handle cases where the filename is different between request and file system.
My project contains that cases and v3.19.0 breaks the code.

```go
Files("/index.html", "/www/data/another.html")
```

* Related to https://github.com/goadesign/goa/pull/2856

tchssk:
@raphael

raphael:
Whoops, yes that makes sense, I think we can modify the generated `appendFS` struct to handle this case - let me take a stab at it.

raphael:
Would you be able to test https://github.com/goadesign/goa/pull/3593 to see if its fixes the issue? Thank you!

PR #3593, opened 2024-09-18
Properly handle serving renamed files

raphael:
This fixes a regression introduced by #3588 that broke the generation of file servers when the served filename differs from the request path filename, .e.g:
```go
Files("/index.html", "/www/data/another.html")
```


tchssk:
LGTM!
Would you release v3.19.1 after merging? Tnank you.

tchssk (approved):

PR #3599, opened 2024-10-04
Use doc.IsPredeclared() and token.IsKeyword() for codegen.fixReservedGo()

tchssk:


raphael (commented):
Thank you, looks great! Do we also need to consider the std lib package names separately?

tchssk:
Oops! I added a condition.

raphael:
Perfect, thank you!

Issue #3605, opened 2024-10-25
ConvertTo using struct with Http* name capitalises word and code does not compile.


markosfount:
As the title says, using `ConvertTo` function with an struct input with a name of `Http*` is generating the wrong code as it capitalises `Http` to `HTTP`. For example:

Define a model as follows:
```
package design

import (
	. "goa.design/goa/v3/dsl" //nolint:revive
	"goa_poc/pkg"
)

var HttpConfig = Type("HttpConfig", func() {
	ConvertTo(pkg.HttpConfig{})
	Attribute("host", String, "The host")
	Attribute("port", Int, "The port number")
})
```
having an external struct like so (assuming this is in a different package/project thus naming cannot be changed):
```
package pkg

type HttpConfig struct {
	Host string
	Port int
}
```
will result in the following `convert.go` after running `generate`:
```
package poc

import (
	pkg "goa_poc/pkg"
)

// ConvertToHttpConfig creates an instance of HttpConfig initialized from t.
func (t *HTTPConfig) ConvertToHttpConfig() *pkg.HttpConfig {
	v := &pkg.HTTPConfig{}
	if t.Host != nil {
		v.Host = *t.Host
	}
	if t.Port != nil {
		v.Port = *t.Port
	}
	return v
}
```
Whereas the return value is the correct one, the `v` value is defined with a nonexisting capitalised struct, which does not compile.
`Service` code and generate file are omitted. Full example here: https://github.com/markosfount/goa-poc

raphael:
Thank you for the report! @tchssk implemented [a fix](https://github.com/goadesign/goa/pull/3607) that is now merged. Are you able to validate the fix? You can try the latest version with:

```
go get goa.design/goa/v3@405ca1c1e52acecef7ab7963f94c58a2a22640a8
```

markosfount:
> Thank you for the report! @tchssk implemented [a fix](https://github.com/goadesign/goa/pull/3607) that is now merged. Are you able to validate the fix? You can try the latest version with:
> 
> ```
> go get goa.design/goa/v3@405ca1c1e52acecef7ab7963f94c58a2a22640a8
> ```

Thanks @raphael, it appears the generate is working now!

PR #3607, opened 2024-10-28
Fix conversion methods to use correct external type

tchssk:
#3605 

It seems that `Goify()` is also applied to external types and their names change.
I implemented using `struct:type:name`, but there may be a more better way.


raphael:
This approach looks good to me, it's consistent with how we've used the `struct:type:name` to coerce the generated type in other places. Also it's low risk since there should be no other generated code that manipulates the external type. Thank you for the fix!

Issue #3614, opened 2024-11-21
Interceptors Proposal


raphael:
# Interceptors in Goa

## Overview
This proposal introduces typed interceptors to Goa's design DSL. Interceptors provide a type-safe mechanism for injecting cross-cutting concerns into method execution. They support both server-side and client-side interception, with clean interfaces for reading and modifying payloads and results, including support for streaming operations.

## Requirements
* Interceptors must be defined in the design
* Interceptors must be fully typed with generated Info types
* Interceptors can read and modify specific payload fields
* Interceptors can read and modify specific result fields 
* Interceptors support both unary and streaming operations
* Interceptors can be applied at both server and client side
* Multiple interceptors can be chained in a defined order

### DSL Example

The DSL allows developers to define interceptors with explicit read and write permissions for payloads and results. Here's an example showing different types of interceptors:

```go
var JWTAuth = Interceptor("JWTAuth", func() {
    Description("Server-side interceptor that validates JWT token and tenant ID")
    
    ReadPayload(func() {
        Attribute("auth", String, "JWT auth token")
        Attribute("tenantID", String, "Tenant ID to validate against")
    })
})

var RequestAudit = Interceptor("RequestAudit", func() {
    Description("Server-side interceptor that provides request/response audit logging")
    
    ReadResult(func() {
        Attribute("status", Int, "Response status code")
    })
    
    WriteResult(func() {
        Attribute("processedAt", String, "Timestamp when processed")
        Attribute("duration", Int, "Processing duration in ms")
    })
})
```

For streaming operations, the DSL provides additional constructs to handle streaming payloads and results:

```go
var TraceBidirectionalStream = Interceptor("TraceBidirectionalStream", func() {
    Description("Interceptor that adds trace context to stream payload")
    
    ReadStreamingPayload(func() {
        Attribute("traceID", String)
        Attribute("spanID", String)
    })
    WriteStreamingPayload(func() {
        Attribute("traceID", String)
        Attribute("spanID", String)
    })
    ReadStreamingResult(func() {
        Attribute("traceID", String)
        Attribute("spanID", String)
    })
    WriteStreamingResult(func() {
        Attribute("traceID", String)
        Attribute("spanID", String)
    })
})
```

Interceptors can be applied at both service and method levels:

```go
var _ = Service("interceptors", func() {
    // Service-wide interceptors
    ServerInterceptor(JWTAuth)
    ServerInterceptor(SetDeadline)
    ClientInterceptor(EncodeTenant)

[...]

PR #3616, opened 2024-11-26
Initial interceptors implementation

raphael:
This PR implements the interceptors proposal described in https://github.com/goadesign/goa/issues/3614


pkg/interceptor.go
```diff
@@ -0,0 +1,22 @@
...
+		// Name of method handling request
+		Method string
+		// Endpoint of request, can be used for retrying
+		Endpoint Endpoint
+		// Payload of request
+		RawPayload any
+	}
+
+	// NextFunc is a function that will continue the request processing chain.
+	NextFunc func(ctx context.Context) (any, error)
```
tchssk: Can't we simply use `goa.Endpoint` instead?

codegen/service/interceptors.go
tchssk, line 1: Should we also generate example interceptor implementation like example service implementation?

raphael: Yes, the latest commits add example generation.

PR #3628, opened 2025-01-02
chore: update speakeasy sponsor graphic

chailandau:
We (Speakeasy) have updated our sponsorship graphic.

Can this be propagated to goa.design if it is not already automatically doing so?


raphael:
Done, thank you!

PR #3633, opened 2025-01-13
Allow setting protoc command via Meta

### LINKED ISSUE #3623: Allow specifying protoc command
**Goal:** To reduce onboarding and setup I want to be able to run `go generate ./...` in a newly cloned project and have that work without any other dependencies other than the correct Go version. I also want to keep versions pinned (using go.mod if possible). For goa I can accomplish that by including a go generate comment like the following. This also ensures my project's goa matches the code generator when maintaining many goa-based projects. 

```go
//go:generate go run goa.design/goa/v3/cmd/goa gen ...
```

**Problem**: If goa needs to launch `protoc` to generate gRPC, the developer must have `protoc` installed and the two plugins that goa uses must be in protoc's path. I've been able to get around this by creating a wrapper script that can download the correct version of `protoc` from GitHub into a cache directory and `go install` the plugins into that same directory, then execute goa with its `PATH` prefixed with that directory. Special care must be taken to ensure this works across Windows/Mac/Linux.

**Solution**: Add a new flag to specify the command to run for protoc. I created https://github.com/duckbrain/goprotoc This is a `go run`able drop-in replacement for `protoc` that will also use `go run` for plugins, You could use  `-protoc "go generate github.com/duckbrain/goprotoc"` (and [soon](https://tip.golang.org/doc/modules/managing-dependencies#tools) `"go tool goprotoc"`) to get a zero dependency go generate with goa.

duckbrain:
Fixes #3623

Adds a meta value `protoc:cmd` to override the command (and prefix args).

If the meta value is provided for both the service and the API, the API will be used an API-wide default and the Service definition will be used as a Service-level override. This seems to make more sense than appending them together like we do with includes.

I wasn't sure if it made more sense to have `Meta("protoc:cmd")` (no values) do nothing or restore the default `protoc` if used as an override in the Service definition, so I made it error. Also considered not allowing it on the service definition to avoid the situation. 

---

Tested this with a private project as well.


grpc/codegen/proto.go
```diff
@@ -100,7 +112,9 @@ func pkgName(svc *expr.GRPCServiceExpr, svcName string) string {
 	return codegen.SnakeCase(svcName)
 }
 
-func protoc(path string, includes []string) error {
+var defaultProtocCmd = []string{"protoc"}
```
raphael: Maybe we add a constant to the DSL that represents the default command so it's possible to do e.g.
```go
Meta("protoc:cmd", DefaultProtoc)
```
to reset to the default behavior in specific services.

The trick used in multiple places to avoid circular dependencies is to define the constant in the `expr` package (so the code above can reference it) then define it as a synonym in the `dsl` package so it's possible to use it in the DSL (and not have the code above nor the `expr` package include the `dsl` package which should not be imported by any code other than designs). See the [`http.go`](https://github.com/goadesign/goa/blob/v3/dsl/http.go#L12) DSL package file as an example.

dsl/meta.go
```diff
@@ -124,6 +124,19 @@ import (
 //	    })
 //	})
 //
+// - "protoc:cmd" provides an alternate command to execute for protoc with
```
raphael, line 133: Great comment, one thing that might help users is to say explicitly that the arguments following "protoc:cmd" are used to build the command line and should expect the same parameters as `protoc`.

grpc/codegen/proto_test.go
```diff
@@ -49,7 +49,7 @@ func TestProtoFiles(t *testing.T) {
 			}
```
raphael, line 52: Might be good to add tests for non default commands.

raphael (commented):
This looks great! I left a few suggestions - nothing major though.

raphael:
Looks great! One tiny last modification: it would probably be good to have a header comment on the "DefaultProtoc" constant so that it is clearly documented in the dsl package online docs. Thank you!

duckbrain:
Whoops, good catch.

Thanks for the great feedback, especially pointing me to the `Meta` option. I started out brainstorming how to handle multiple arguments on a CLI flag.

raphael:

PR #3640, opened 2025-01-29
OpenAPIv3: count different validation rules as different types

isaacseymour:
This replicates an issue we're seeing in the OpenAPIv3 generator, where if two request bodies are structurally similar (i.e. same primitive types), but not identical (in this example both have a single string attribute with the same name, but with different enum validations), the request body for those two methods points at the same type schema.

There's a (previously-failing) test, where:
- two types which are 'similar' (both have a string-type attr called `my_attr`), but not identical (they are different enums)
- two different services with the same method, each referencing different types

The resulting OpenAPIv3 schema generates a single
`MethodEnumRequestBody` type, and uses it for the `requestBody` type of both methods.

This is incorrect, since the two methods use a different enum!

The fix here is to update the `hashAttribute` method to hash the `.Validation` of the attribute, if it is set :tada:

### DIFF
--- a/http/codegen/openapi/v3/testdata/dsls/types.go
+++ b/http/codegen/openapi/v3/testdata/dsls/types.go
@@ -15,6 +15,49 @@ func StringBodyDSL(svcName, metName string) func() {
 	}
 }
 
+func StringEnumBodyDSL() func() {
+	return func() {
+		var T1 = Type("T1", func() {
+			Attribute("my_attr", String, func() {
+				Enum("a", "b", "c")
+			})
+		})
+
+		var T2 = Type("T2", func() {
+			Attribute("my_attr", String, func() {
+				Enum("d", "e")
+			})
+		})
+
+		var _ = Service("svc_enum_1", func() {
+			Method("method_enum", func() {
+				Payload(func() {
+					Reference(T1)
+					Attribute("my_attr")
+					Required("my_attr")
+				})
+				HTTP(func() {
+					POST("/")
+				})
+			})
+		})
+
+		var _ = Service("svc_enum_2", func() {
+			Method("method_enum", func() {
+				Payload(func() {
+					Reference(T2)
+					Attribute("my_attr")
+					Required("my_attr")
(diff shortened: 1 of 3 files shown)


http/codegen/openapi/v3/testdata/dsls/types.go
```diff
@@ -15,6 +15,49 @@ func StringBodyDSL(svcName, metName string) func() {
...
+		var T2 = Type("T2", func() {
+			Attribute("my_attr", String, func() {
+				Enum("d", "e")
+			})
+		})
+
+		var _ = Service("svc_enum_1", func() {
+			Method("method_enum", func() {
+				Payload(func() {
+					Reference(T1)
```
raphael, line 35: Is the `Reference` required to repro the bug? would it also happen by simply doing `Payload(T1)`?

http/codegen/openapi/v3/types.go
```diff
@@ -440,6 +440,87 @@ func hashAttribute(att *expr.AttributeExpr, h hash.Hash64, seen map[string]*uint
...
+		return 0
+	}
+	var parts []uint64
+	if val.Format != "" {
+		parts = append(parts, hashString(string(val.Format), h))
+	}
+	if val.Pattern != "" {
+		parts = append(parts, hashString(val.Pattern, h))
+	}
+	if val.MinLength != nil {
```
raphael, line 466: This algo may require a bit of tweaking, as it stands the following two validations would result in the same hash:
```go
Pattern("1")
```
and 
```
MinLength(1)
```
Obviously there are many other combinations that would result in the same hash being generated. Maybe we need to use something like https://github.com/gohugoio/hashstructure,

Might also be good to add the example above (and maybe others) to tests.

raphael (commented):
Thank you for doing this, makes a lot of sense. The algo might need a bit of tweaking before we can merge it.

raphael:
Superseded by https://github.com/goadesign/goa/pull/3642

PR #3641, opened 2025-02-01
Streaming Interceptors

douglaswth:
- Initial implementation of Streaming Interceptors generation based on #3614 
- Add `SendWithContext` and `RecvWithContext` methods to the stream interfaces
- Fix several generation bugs


codegen/service/templates/client_interceptor_wrappers.go.tpl
```diff
@@ -4,19 +4,125 @@
...
+		return i.{{ $interceptor.Name }}(ctx, info, endpoint)
+	{{- end }}
+	}
 }
 {{ end }}
 {{- end }}
+{{- range .WrappedClientStreams }}
+
+{{ comment (printf "wrapped%s is a client interceptor wrapper for the %s stream." .Interface .Interface) }}
+type wrapped{{ .Interface }} struct {
```
raphael: Might be nice to move the type definition to the top of the file

pkg/interceptor.go
```diff
@@ -9,8 +9,10 @@ type (
 		Service string
```
raphael: We should update the struct header comment to reflect the changes

pkg/interceptor.go
```diff
@@ -9,8 +9,10 @@ type (
...
 		// Name of method handling request
 		Method string
-		// Endpoint of request, can be used for retrying
-		Endpoint Endpoint
+		// Send is true if the request is a streaming Send
```
raphael: What's the rationale for these two fields? I could see `Send` helping differentiate between the initial request and the streaming request in the case of websocket but not sure I understand the point of `Recv`?

raphael (commented):
Really great PR! I left very few minor comments. Overall it all makes sense and follows the existing patterns very well, good job!

pkg/interceptor.go
douglaswth: Let's say you are building an OTel tracing interceptor for a bidirectional stream that needs to send per message trace and span IDs in both directions:
- The interceptor will need to grab the tracing data from the `context.Context` and add it to the message when intercepting `Send`
- The interceptor will need to grab the tracing data from the message and add it to the `context.Context` when intercepting `Recv`
Having both of these fields allows the interceptor to know whether it is `Send`, `Recv`, or the initial request.

raphael: Got it, makes sense. I wonder if we could have only one field then?

douglaswth: I suppose we could replace it with an enum.

douglaswth: Maybe something like: `goa.InterceptorMethod`, `goa.InterceptorStreamingSend`, `goa.InterceptorStreamingRecv`.

raphael: Yes, `goa.InterceptorUnary` instead of `Method` maybe?

Issue #2850, opened 2021-05-18
"additionalProperties: false" not specified in OpenAPI specification


jamestelfer:
Goa supports OpenAPI 3.03, whose [specification states](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#properties) that:

> additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. Consistent with JSON Schema, additionalProperties defaults to true.

Since `additionalProperties` is not specified in the OpenAPI specifications, this means the default applies (`true`) rather than the expected `false`. For example, the generated types do not support generating additional properties on serialization and will ignore them when deserializing.

The proposal here would be to generate `additionalProperties: false` by default in the generated spec, as suggested by @raphael in the discussion linked below.

_Original discussion started by @jamestelfer in https://github.com/goadesign/goa/discussions/2833_

PR #3644, opened 2025-02-03
Add openapi:additionalProperties Meta

tchssk:
https://github.com/goadesign/goa/issues/2850

* Currently only boolean values ​​are supported.
* This is a feature that only affects OpenAPI, but we would like to consider real server validation functionality in the future.


raphael:
My understanding is that `additionalProperties` should only be true for maps in Go, Go objects do not make it possible to dynamically add fields to them. Maybe we should instead always set it to `false` except for maps (which is already taken into consideration)? The [specification](https://spec.openapis.org/oas/v3.0.3#schema-object) says:

>Consistent with JSON Schema, additionalProperties defaults to true.

So it does seem like we should always generate it and set it to `false` except for maps.

tchssk:
Sorry, I didn't explain it enough. This function was supposed to set `additionalProperties: false` on an object.

In this case, `bar` is sent even though it is not defined in the OpenAPI schema, but that is OK because `additionalProperties` is considered `true` by default.

### OpenAPI schema

```yaml
Payload:
    title: Payload
    type: object
    properties:
        foo:
            type: string
```

### Incoming request

```json
{"foo":"1","bar":"2"}
```

If `additionalProperties: false` is specified, this will be considered an invalid request and will cause OpenAPI validation errors in many libraries.

All of this is about OpenAPI and has nothing to do with the actual server implementation. In terms of Go `encoding/json`, if the unmarshal target contains undefined fields, they are simply ignored.

raphael:
Ah great, that makes sense! Thank you for the explanation.

PR #3655, opened 2025-02-18
Fix a few streaming interceptor bugs

douglaswth:
- The wrap function for server interceptors was working as if it were a client interceptor: expecting the stream to be the result of calling the endpoint or the next unary interceptor in the chain. It now properly wraps the stream in the raw payload (which is actually a special `EndpointStruct` with the stream and optionally the unary payload).
- The unary `Payload` methods of server interceptor info structs for a streaming method assumed that the raw payload would be the payload type rather than the special `EndpointStruct` type. Now, the `Payload` methods properly expect the raw payload to either be the `EndpointStruct` or the payload type in the case when the server interceptor info struct is also used for a client interceptor.
- With the change of `goa.InterceptorInfo` from a struct to an interface, the generation of implementation of the interface methods on the generated interceptor info structs became necessary. The code for those methods was included in a section of the template that was conditional on the interceptor having accessors. Now, the conditional has been moved to a more appropriate place so the interface methods are implemented.

### DIFF
--- a/codegen/service/interceptors.go
+++ b/codegen/service/interceptors.go
@@ -122,6 +122,7 @@ func interceptorFile(svc *Data, server bool) *codegen.File {
 			Data:   interceptors,
 			FuncMap: map[string]any{
 				"hasPrivateImplementationTypes": hasPrivateImplementationTypes,
+				"hasEndpointStruct":             hasEndpointStruct(server),
 			},
 		})
 	}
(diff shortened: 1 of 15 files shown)


douglaswth:
I've found one more bug to fix which I'll push here.

raphael:
This is great, thank you!

PR #3656, opened 2025-02-21
Fix client gRPC endpoint generation to appease Go 1.24 test

douglaswth:
- With a `go.mod` which sets the Go version to `1.24.0`, `go test` complains with `non-constant format string in call to goa.design/goa/v3/pkg.Fault` when it sees generated client gRPC endpoint methods. This fixes the generation to include a format string of `"%s"` before the single argument.


raphael:
Nice, thank you!

PR #3684, opened 2025-03-27
Fix warning QF1004 of staticcheck

mamo3gr:
## Problem

A linter [Staticcheck](https://staticcheck.dev/), which could be included in golangci-lint, warns QF1004 in the generated code:  
https://staticcheck.dev/docs/checks/#QF1004

>Use strings.ReplaceAll instead of strings.Replace with n == -1

## Solution

Do what the linter says.  
There seems not to be any drawback so all replacement are applied in an automatic manner (including test data).


raphael:
This is great, thank you!

Issue #3696, opened 2025-04-11
Proposal: Native Support for Server-Sent-Events (SSE)


raphael:
# Background

Server-Side-Events is a HTTP based server to client streaming protocol that is gaining more popularity thanks to its simplicity, HTTP compatibility and adoption by newer protocols like [MCP](https://modelcontextprotocol.io/introduction).  

Goa makes it possible to implement an HTTP endpoint which implements [Server-Sent-Events](https://html.spec.whatwg.org/multipage/server-sent-events.html) with a Goa service today by defining a design that complies with the standard and bypassing the generation of decoder via [SkipResponseBodyEncodeDecode](https://pkg.go.dev/goa.design/goa/v3/dsl#SkipResponseBodyEncodeDecode):

```go
Method("stream", func() {
	Description("Stream events using Server-Sent Events")
	HTTP(func() {
		GET("/events/stream")
		SkipResponseBodyEncodeDecode()
	})
})
```

Dummy implementation:

```go
type sseReader struct{}

// Implement Goa generated endpoint method
func (s *Service) Stream(ctx context.Context) (resp io.ReadCloser, err error) {
    resp = &sseReader{}
    return
}

// Go http stdlib implementation will call WriteTo on the ResponseWriter
func (r *sseReader) WriteTo(w io.Writer) (n int64, err error) {
    rw := w.(http.ResponseWriter)
    rw.Header().Set("Content-Type", "text/event-stream")
    rw.Header().Set("Cache-Control", "no-cache")
    w.Write([]byte("data: foo"))
    return 9, nil
}

// Dummy implementation to satisfy io.ReadCloser interface
// WriteTo will get used
func (r *sseReader) Read(_ []byte) (n int64, err error) {
    return 0, nil
}

func (r *sseReader) Close() error {
    return nil
}
```

While this works it requires intimate knowledge with the SSE protocol, the Go http stdlib package and Goa's `SkipResponseBodyEncodeDecode` behavior. This proposal adds native support for SSE to Goa, making it trivial to add SSE endpoints in Goa services.

# DSL

This proposal introduces a new DSL function `ServerSentEvents` which tells Goa that a streaming endpoint should be implemented using SSE instead of the default WebSocket implementation:

```go
var Event = Type("Event", func() {
	Description("A notification message sent via SSE")
	Attribute("message", String, "Message body")
	Attribute("timestamp", Int, "Unix timestamp")
	Required("message", "timestamp")
})

Method("stream", func() {
[...]

PR #3697, opened 2025-04-12
SSE Support

raphael:
This PR implements the proposal described in #3696

Done:
- [x] DSL and expressions
- [x] Response rendering
- [x] Handling of Last-Request-Id server side
- [x] Client side


http/codegen/server.go
```diff
@@ -266,10 +270,10 @@ func viewedServerBody(sbd []*TypeData, view string) *TypeData {
 }
 
 func addLeadingSlash(s string) string {
-	if strings.HasPrefix(s, "/") {
-		return s
+	if s == "" || s[0] != '/' {
+		return "/" + s
 	}
-	return "/" + s
+	return s
```
tchssk, line 276: Is this better?

dsl/sse.go
```diff
@@ -0,0 +1,302 @@
...
+	if len(args) > 2 {
+		eval.TooManyArgError()
+		return
+	}
+	if len(args) == 2 {
+		if _, ok := args[1].(func()); !ok {
+			eval.InvalidArgError("function", args[1])
+			return
+		}
+	}
```
tchssk: Is this if block a duplicate with L104?

http/codegen/server.go
raphael: Probably doesn't change much but slightly more efficient (we don't need to check for an entire string prefix, we can just test the first character).

dsl/sse.go
raphael: Yes, good catch!

PR #3711, opened 2025-05-19
Support Go 1.23+ only

tchssk:
* There is no need to specify a toolchain.
* There is no need to test for Go 1.20 and 1.21.
    * They should be removed from required status checks.


raphael:
Great, thank you! I removed Go 1.21 and Go 1.22 from the required checks.

PR #3713, opened 2025-05-24
add support for path params in middle of the path

dvjn:
Added support for path params in the middles of the paths as well

Earlier `/user/@{username}` didn't register `username` as a path param.
Similarly `/{month}-{day}-{year}` from chi's [readme examples](https://github.com/go-chi/chi?tab=readme-ov-file#examples) would only detect `{month}` as a param and nothing else.

After this change, all these path params should be detected and should work .

### DIFF
--- a/expr/http.go
+++ b/expr/http.go
@@ -37,7 +37,7 @@ type (
 
 // HTTPWildcardRegex is the regular expression used to capture path
 // parameters.
-var HTTPWildcardRegex = regexp.MustCompile(`/{\*?([a-zA-Z0-9_]+)}`)
+var HTTPWildcardRegex = regexp.MustCompile(`{\*?([a-zA-Z0-9_]+)}`)
 
 // ExtractHTTPWildcards returns the names of the wildcards that appear in
 // a HTTP path.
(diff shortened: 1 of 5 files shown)


dvjn:
I have tested these changes locally in a working project and am waiting for a maintainers review before adding specific test cases for this feature.

Please let me know if anything else is needed for this PR. Thanks.

raphael:
Thanks for this PR! This is a great enhancement that brings Goa's path parameter detection in line with Chi router. The examples you mentioned (`/user/@{username}` and `/{month}-{day}-{year}`) are useful patterns. The core change seems simple, since it still requires valid parameter names `[a-zA-Z0-9_]+`, it won't accidentally match anything weird. And the OpenAPI generation should work fine since the surrounding characters (like `@` or `-`) stay in place.

To make sure everything works end-to-end, could you add a few tests? I'm thinking `http/mux_test.go` would be a good place to add new mux tests. We should also add a validation test to make sure routes like `/{year}-{month}-{day}` properly validate that all three parameters exist in the method payload - you can extend the existing `TestHTTPRouteValidation` in `expr/http_endpoint_test.go` for this. Maybe add an OpenAPI test case with one of the new patterns to verify the generated spec looks right - adding a new DSL function to `http/codegen/testdata/openapi_dsls.go` should automatically generate the test cases.

That should cover it! The change itself is solid - just want to make sure the integration points work as expected.
Thanks for the contribution! 👍

raphael:
Thanks again for this, I still think it's a useful enhancement. Closing it for now since it went stale before the tests we discussed were added, and it now conflicts with the OpenAPI builders on `v3` as well as with the v3.31 generator rework (#3971). If you want to pick it back up after v3.31 ships, a refreshed PR with the mux, route validation and OpenAPI tests would be very welcome. 👍

PR #3727, opened 2025-06-18
Check for nil interceptors before wrapping

nitinmohan87:
Wrap the endpoints with interceptors only when provided


raphael:
Makes sense, thank you!

PR #3737, opened 2025-07-08
Wrap json.Encoder returned by RequestEncoder to assign GetBody on the http.Request 

### LINKED ISSUE #3725: Generated client cannot handle http2 GOAWAY
Using the generated goa client, I've been getting this error somewhat regularly while under load.

```
http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error
```

In `http.NewRequestFromContext`, there's [special logic](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/net/http/request.go;l=928) to assign `GetBody` when the passed in body is a `*bytes.Buffer`. This avoids that issue. Goa passes `nil` and later assigns a request body via `goa.design/goa/v3/http.RequestEncoder`, which does not assign `GetBody`. It should be updated to return an `Encoder` that assigns `r.GetBody`.

duckbrain:
Closes #3725

Prior to this change, the following error could occur with the generated client code when connecting over http2

```
http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error
```

This PR does as the message suggests and defines Request.GetBody


raphael:
This looks great, thank you. The only thing might be to add comments to explain the code, something like:

  1. In the RequestEncoder function (around lines 181-184):

  Before:
```go
  enc := new(jsonEncoder)
  r.Body = enc
  r.GetBody = enc.GetBody
  return enc
```

  After:

```go
  enc := new(jsonEncoder)
  r.Body = enc
  // GetBody enables request retry on HTTP/2 connections when the server
  // sends GOAWAY during graceful shutdown. Without GetBody, the HTTP transport
  // cannot retry because the request body has already been consumed.
  r.GetBody = enc.GetBody
  return enc
```

  2. For the jsonEncoder struct (around lines 185-188):

```go
  Before:
  type jsonEncoder struct {
  	b []byte
  	r bytes.Reader
  }
```

  After:

```go
  // jsonEncoder implements io.ReadCloser and provides GetBody functionality
  // to support HTTP/2 request retries during server graceful shutdown (GOAWAY).
  type jsonEncoder struct {
  	b []byte       // encoded JSON bytes
  	r bytes.Reader // reader for the encoded bytes
  }
```

  3. For the GetBody method (around lines 211-215):

  Before:

```go
  func (je *jsonEncoder) GetBody() (io.ReadCloser, error) {
  	if len(je.b) == 0 {
  		return nil, errEncodeNotCalled
  	}
  	return io.NopCloser(bytes.NewReader(je.b)), nil
  }
```

  After:

```go
  // GetBody returns a new reader of the encoded bytes, enabling request retries.
  // This is required for HTTP/2 connections to handle server GOAWAY during graceful shutdown.
  func (je *jsonEncoder) GetBody() (io.ReadCloser, error) {
  	if len(je.b) == 0 {
  		return nil, errEncodeNotCalled
  	}
  	return io.NopCloser(bytes.NewReader(je.b)), nil
  }
```

duckbrain:
Sorry, for the delay. I've now got the comments added and satisfied the CI checks.

raphael:
Looks great, thank you!

PR #3741, opened 2025-07-18
Codegen fixes

nitinmohan87:
* Fixes user type imports which was broken after #3721 
* Adds Unwrap interface to unwrap interceptor objects which is then used in error handling
* Nil checks for `errhandler` func in HTTP since it is optional


raphael:
This is great, thank you!

PR #3746, opened 2025-07-21
Fix panic when parsing malformed query parameters with nested brackets

rorydbain:
## Summary
- Fixed panic in generated query parameter parsing code when malformed brackets are provided
- Added validation to check that brackets are properly paired and ordered before attempting string slicing
- Updated test expectations to match the new error handling

## Problem
When query parameters with malformed brackets are sent (e.g., `custom_field[id]]][operator]` or `custom_field[[nestedbrackets]]`), the generated code would panic with `runtime error: slice bounds out of range [2:0]`.

This occurred because the code only checked if the closing bracket was missing (`closeIdx == -1`) but didn't validate:
1. If the opening bracket exists (`openIdx == -1`)
2. If the closing bracket comes after the opening bracket (`closeIdx <= openIdx`)

## Solution
Updated the template to check all three conditions:
```go
if openIdx == -1 || closeIdx == -1 || closeIdx <= openIdx {
    err = goa.MergeErrors(err, goa.DecodePayloadError("invalid query string: malformed brackets"))
}
```

### DIFF
--- a/http/codegen/templates/partial/query_map_conversion.go.tpl
+++ b/http/codegen/templates/partial/query_map_conversion.go.tpl
@@ -5,8 +5,8 @@
 	{
 		openIdx := strings.IndexRune(keyRaw, '[')
 		closeIdx := strings.IndexRune(keyRaw, ']')
-		if closeIdx == -1 {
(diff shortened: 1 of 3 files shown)


raphael:
That makes sense, thank you for the great contribution!

PR #3750, opened 2025-08-01
Apply modernize

tchssk:
https://pkg.go.dev/golang.org/x/tools/gopls@v0.20.0/internal/analysis/modernize


raphael:
This is great, thank you!

PR #3794, opened 2025-09-05
grpc/cli: fix usage format with % in examples

raphael:
Avoid raw fmt.Fprintf with %[1]s and raw backtick blocks when example text includes percent signs; use Fprint + Fprintf to prevent format parsing errors in generated CLI.


raphael:
superseded by #3795

PR #3826, opened 2025-10-20
Add `allowEmptyValue` overwrite

danidomi:
right now it's not possible to add `AllowEmptyValues` if configured the following which end up with errors when using swagger

```go
Method("fruits", func() {
		Description("Lists of fruits")
		Payload(func() {
			Field(1, "fruits", ArrayOf(String, func() {
				AllowEmptyValue(true)
				Enum(
					"orange",
					"banana",
					"kiwi",
				)
			}), func() {
				Description("awe some fruits")
				MaxLength(10)
			})
```

swagger-ui error when calling 

```
{
  "name": "invalid_enum_value",
  "id": "X0Le_P7U",
  "message": "value of fruits[*] must be one of \"orange\", \"banana\", \"kiwi\" but got value \"\"",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```


raphael:
Makes sense, I am wondering if instead of adding extra DSL, we could make it work by simply adding the empty value to the enum? In general a high level goal is to keep the DSL surface as small as possible to make it easier for newcomers (and experienced users) to learn / remember what's possible.

raphael:
Thanks for the report and the PR. I'm going to close this one to keep the DSL surface small: you can get the behavior you want today by adding the empty string to the enum, e.g. `Enum("", "orange", "banana", "kiwi")`. That accepts the empty value at runtime and documents it in the OpenAPI spec, whereas the `allowEmptyValue` flag alone would only change the spec while the generated validation still rejected `""`.

Issue #3828, opened 2025-10-27
Sending `null` values in arrays that are required causes a panic


leonore:
If you declare something like this:
```go
var Item = Type("Item", func() {
	Attribute("id", String, "ID of the item", func() {
		Example("item-1234")
	})
	Attribute("name", String, "Name of the item", func() {
		Example("Example Item")
	})
	Attribute("description", String, "Description of the item", func() {
		Example("This is an example item.")
	})
	
	Required("id", "name", "description")
})

var Payload = Type("Payload", func() {
	Attribute("items", ArrayOf(Item))
	Required("items")
})
```

And then spend a payload with an array with a null item:
```
{"items": [null]}
```

We've noticed you get a nil pointer dereference from the generated `unmarshalPayloadRequestBodyToPayload` function, as goa assumes each array item is an object.

The expected behaviour would be
1. Goa should not panic, it should return an array with nils (i.e. it's up to the API handler to handle this)
2. There should be a validation rule that makes it possible to specify that elements in an array cannot be nil

raphael:
Thanks for opening the issue!

* #3830 prevents the panic
* #3834 introduces a new DSL `ArrayOfRequired` which makes it possible to define arrays that must not include nil elements

PR #3834, opened 2025-10-30
Enhance array validation: add support for non-nullable Array elements

raphael:
- Introduced `ArrayOfRequired` function to create array types that disallow null elements.
- Updated validation logic to check for non-nullable elements in arrays during validation.
- Added new test cases to cover scenarios involving required arrays with non-nullable elements.
- Updated golden files to reflect changes in validation behavior.

This change improves the robustness of array validation in the code generation process.

**DSL design considerations** 
In general we try to avoid introducing new DSL functions to limit the language surface. In this case the considered alternatives ended up being worse:
- Introducing a `NonNilItem` or such validation introduces new DSL anyway and is less convenient to use (would have to be in the Array sub-DSL).
- reusing the `Required` keyword with a special marker introduces a new DSL (the marker) or with no argument makes it inconsistent and confusing to implement (also less convenient to use)

All in all `ArrayOfRequired` does not seem to add too much cognitive load and allows for a more natural DSL.

### DIFF
--- a/codegen/templates/validation/array.go.tpl
+++ b/codegen/templates/validation/array.go.tpl
@@ -1,3 +1,10 @@
 for _, e := range {{ .target }} {
+{{- if .nonNullableElems }}
+	if e == nil {
+		err = goa.MergeErrors(err, goa.MissingFieldError("{{ .context }}", "[*]"))
+	}
+{{- end }}
+{{- if .validation }}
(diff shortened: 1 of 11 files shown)


raphael:
See https://github.com/goadesign/goa/issues/3828

PR #3841, opened 2025-11-10
Performance: optimise file writes

isaacseymour:
tl;dr: make codegen much faster through parallel writing of files, and writing each file's contents once.

---

We have a _very large_ API design, with over 100 services, which generates almost 3k files with `goa generate`. As our API has grown, running `goa gen` has got really slow.

There's three changes in this PR:
1. Emit `[TIMING]` logs when `--debug` is passed - this makes finding bottlenecks _much_ easier
2. Use a worker pool to write files: parallelising this makes it significantly faster (~3x)
3. Only write files once they are finalized - i.e. tidying up imports with an in-memory copy of the file, before persisting it at the end.

This cut the time on that stage for us from ~52s to ~10s 🎉

### DIFF
--- a/codegen/generator/generate.go
+++ b/codegen/generator/generate.go
@@ -1,9 +1,13 @@
 package generator
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
+	"sync"
+	"time"
 
 	"goa.design/goa/v3/codegen"
 	"goa.design/goa/v3/eval"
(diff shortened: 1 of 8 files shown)


codegen/generator/generate.go
```diff
@@ -65,58 +80,167 @@ func Generate(dir, cmd string) (outputs []string, err1 error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if debug {
+			fmt.Fprintf(os.Stderr, "[TIMING]     [generate] Stage 4: Run pre-generation plugins took %v\n", time.Since(start))
+		}
 	}
 
 	// 5. Generate initial set of files produced by goa code generators.
+	// NOTE: Parallelization causes infinite recursion in AsObject() for circular type references
```
isaacseymour, line 116: I'm going to try to find time to work on this problem in the next couple weeks ideally. This stage is now the slowest (~20s), but parallelising is fiddly due to circular references.

expr/random.go
```diff
@@ -69,10 +70,13 @@ func NewRandom(seed string) *ExampleGenerator {
 type ExampleGenerator struct {
 	Randomizer
 	seen map[string]*any
+	mu   sync.RWMutex
```
isaacseymour, line 73: While this isn't essential right now, I think it's sensible to add locks around map operations to avoid panics in future!

raphael:
This is awesome, thank you for the great work! Just one thing that might be nice before we can merge: would it be possible to add a few tests for the new concurrent part of the code?

isaacseymour:
I've added some tests that hit the "generate lots of files" code, but they don't _directly_ test the concurrency (although I have run them with `-race` successfully!)

Was there anything else you were thinking of @raphael ?

raphael:
Nope all good, - merged it in this PR (had to rebase): https://github.com/goadesign/goa/pull/3852
It will be part of v3.23.0 - thank you!

PR #3852, opened 2025-11-26
Performance: optimise file writes

raphael:
tl;dr: make codegen much faster through parallel writing of files, and writing each file's contents once.

We have a very large API design, with over 100 services, which generates almost 3k files with goa generate. As our API has grown, running goa gen has got really slow.

There's three changes in this PR:

Emit [TIMING] logs when --debug is passed - this makes finding bottlenecks much easier
Use a worker pool to write files: parallelising this makes it significantly faster (~3x)
Only write files once they are finalized - i.e. tidying up imports with an in-memory copy of the file, before persisting it at the end.
This cut the time on that stage for us from ~52s to ~10s 🎉

Thank you @isaacseymour !

### DIFF
--- a/cmd/goa/gen.go
+++ b/cmd/goa/gen.go
@@ -13,6 +13,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
 	"goa.design/goa/v3/codegen"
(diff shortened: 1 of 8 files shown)


raphael:
Supersedes #3841

PR #3857, opened 2025-12-13
grpc/codegen: allow overriding proto json_name

ilCollez:
## Summary
- extend the DSL with `Meta("proto:tag:json", ...)` so designers can override the `json_name` emitted in generated protobuf definitions
- plumb the metadata through `grpc/codegen/protobuf.go`, emitting `[json_name = "..."]` on both regular fields and oneof members when present
- update the DSL documentation to mention the new metadata key and add unit tests that cover the option for structs and oneofs

## Motivation
Some downstream consumers rely on protobuf JSON descriptors whose field names must diverge from the default lower_snake_case. Prior to this change there was no way to control the descriptor in Goa-generated protos without patching the generated files. The new metadata brings parity with the existing struct-tag overrides and keeps designs self-contained.

## Example
DSL:
```go
var Item = Type("Item", func() {
    Field(1, "ID", String, func() {
        Meta("proto:tag:json", "identifier")
    })
})
```

Generated `.proto` fragment
```proto
message Item {
    string id = 1 [json_name = "identifier"];
}
```

Generated Go struct (`.pb.go`)
```go
type Item struct {
    Id string `protobuf:"bytes,1,opt,name=id,proto3" json:"identifier,omitempty"`
}
```

### DIFF
--- a/dsl/meta.go
+++ b/dsl/meta.go
@@ -130,6 +130,15 @@ const DefaultProtoc = expr.DefaultProtoc
 //	    })
 //	})
 //
+// - "proto:tag:json" sets the JSON name emitted in the generated protobuf
+//   field option. Applicable to attributes only.
+//
+//	var MyType = Type("MyType", func() {
+//	    Field(1, "id", String, func() {
+//	        Meta("proto:tag:json", "identifier")
+//	    })
+//	})
+//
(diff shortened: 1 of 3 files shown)


raphael:
Looks great, thank you!

PR #3868, opened 2025-12-18
grpc/codegen: use google.protobuf.Value instead of Any for Goa's Any …

ilCollez:
## Summary

Changes the gRPC transport mapping for Goa's `Any` type from `google.protobuf.Any` (with JSON marshaling) to `google.protobuf.Value` (direct conversion via `structpb.NewValue()`).

## Motivation

The previous implementation was semantically incorrect and inefficient:
- `google.protobuf.Any` is designed for wrapping typed messages with type URLs
- Required JSON marshaling/unmarshaling for dynamic values
- `google.protobuf.Value` is purpose-built for JSON-like dynamic values (null, number, string, bool, struct, list)

## Changes

**Code Generation:**
- protobuf.go: Map `Any` to `google.protobuf.Value` / `*structpb.Value`
- protobuf_transform.go: Use `structpb.NewValue()` for Go→Proto, `AsInterface()` for Proto→Go
- service_data.go: Import `google/protobuf/struct.proto`
- client_types.go, server_types.go: Auto-add `fmt` and `structpb` imports

**Behavior:**
- Go `nil` → `structpb.NewNullValue()` (protobuf null)
- Conversion errors trigger panic (indicates unsupported type bug)

**Documentation:**
- types.go: Updated `Any` type comment
- FAQ.md: Rewritten Any type section

## Benefits

- **Semantic correctness**: Using the right protobuf type for dynamic values
- **Better performance**: No JSON marshaling overhead
- **Simpler code**: Direct conversion instead of wrapper logic
- **Clear error handling**: Panics on programming errors

## Testing

- Updated unit tests in `grpc/codegen/*_test.go`
- Updated golden files to match new generated code
- All tests pass: `make`, `make test`, `make integration-test`

### DIFF
--- a/grpc/codegen/protobuf_transform.go
+++ b/grpc/codegen/protobuf_transform.go
@@ -679,32 +679,22 @@ func convertType(src, tgt *expr.AttributeExpr, srcPtr bool, tgtPtr bool, srcVar 
 	return convertPrimitiveFromProto(src, tgt, srcPtr, tgtPtr, srcVar, ta)
 }
 
-const convertGoAnyToProtobufAnyFunc = `func() *anypb.Any {
+const convertGoAnyToProtobufValueFunc = `func() *structpb.Value {
+	// Convert Go any to protobuf Value directly
 	if %s == nil {
-		return nil
+		return structpb.NewNullValue()
 	}
-	// Convert Go any to protobuf Any using JSON marshaling
-	if jsonData, err := json.Marshal(%s); err == nil {
-		if data, err := anypb.New(&structpb.Value{
-			Kind: &structpb.Value_StringValue{StringValue: string(jsonData)},
-		}); err == nil {
-			return data
-		}
+	value, err := structpb.NewValue(%s)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert value to structpb.Value: %%v", err))
 	}
-	return nil
+	return value
 }()`
 
-const convertProtobufAnyToGoAnyFunc = `func() any {
+const convertProtobufValueToGoAnyFunc = `func() any {
+	// Convert protobuf Value to Go any directly
 	if %s != nil {
-		var value structpb.Value
-		if err := %s.UnmarshalTo(&value); err == nil {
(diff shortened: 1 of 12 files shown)


grpc/codegen/protobuf_transform.go
```diff
@@ -679,32 +679,22 @@ func convertType(src, tgt *expr.AttributeExpr, srcPtr bool, tgtPtr bool, srcVar
...
-	// Convert Go any to protobuf Any using JSON marshaling
-	if jsonData, err := json.Marshal(%s); err == nil {
-		if data, err := anypb.New(&structpb.Value{
-			Kind: &structpb.Value_StringValue{StringValue: string(jsonData)},
-		}); err == nil {
-			return data
-		}
+	value, err := structpb.NewValue(%s)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert value to structpb.Value: %%v", err))
```
raphael, line 689: We probably shouldn't panic in runtime code unless it's a bug which in this case I'm not sure we can guarantee given that the value is dynamic?

raphael (commented):
This looks great, thank you for the contribution. I left a comment that would be good to address before we can merge.

grpc/codegen/protobuf_transform.go
ilCollez: This panic would only trigger if a programmer using the library tries to convert something that’s not JSON-serializable. Since the data comes from the client and is already JSON-deserializable, it shouldn’t happen with normal client data. It’s meant to catch programming errors. If this panics raises, it means they’re using the library wrong, not that there’s bad runtime data.

My only reason for this approach is that I wanted to make these errors explicit rather than silently creating undefined behavior that would need debugging later. But I get the concern about panics in library code, I'll happy to switch to returning an error instead if you prefer.

Since you’re the maintainer, it’s your call. What do you think?

raphael: Yeah that makes sense, the data was serialized so is guaranteed to be unserializable if the code is used as intended. Thanks for the clarification.

PR #3871, opened 2026-01-11
feat: added oneof:type:field and oneof:value:field support to OneOf union

akhilmhdh:
## Context

This PR implements to customize the keys for the  new oneOf union change for goa. Previously the keys were hardcoded as `type` and `value`. 

Example
```json
{ "type": "sql-database", "value":  {  } }
```

But the problem with this is - some APIs may not be following this same exact format. Example
```json
{ "kind": "", "payload": {} }
```

This PR will allow customization of these keys and also if not provided will keep the default keys as `type` and `value`.

Example
```go
OneOf("Shape", func() {
    Meta("oneof:type:field", "kinda")
    Meta("oneof:value:field", "payload")
    Attribute("Circle", Int)
})
```

### DIFF
--- a/codegen/service/service_data.go
+++ b/codegen/service/service_data.go
@@ -430,6 +430,10 @@ type (
 		// Loc defines the file and Go package of the union type if overridden via
 		// Meta. When nil the type is generated in the default service file.
 		Loc *codegen.Location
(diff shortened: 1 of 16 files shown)


raphael:
This is great, thank you!

PR #3879, opened 2026-01-30
Fixed http client code generator

abocim:
This PR fixes a bug we have hit within http client code generator.

In case two or more Methods using `Payload(ArrayOf(...))` are defined, generator generates _HTTP request body builder_ function for the first of defined Method only _(client/types.go file)_.

Bug can be reproduced using new added unit test.

**Goa design code:**
```go
       var PayloadA = Type("PayloadA", func() {
               Attribute("a", String, func() {
                       Pattern("patterna")
               })
       })
       var PayloadB = Type("PayloadB", func() {
               Attribute("a", String, func() {
                       Pattern("patterna")
               })  
               Attribute("b", String, func() {
                       Pattern("patternb")
               })  
               Required("a", "b")
       })  
       Service("ServiceMultipleMethods", func() {
               Method("MethodA", func() {
                       Payload(ArrayOf(PayloadA))
                       HTTP(func() {
                               POST("/")
                       })
               })
               Method("MethodB", func() {
                       Payload(ArrayOf(PayloadB))
                       HTTP(func() {
                               PUT("/")
                       })
               })
       })
```

**Bug is located in `http/codegen/client_types.go: func clientType`:**
In this function there is incorrectly used `data.Name` attribute for already seen payload types. Attribute is always set to `"array"` value regardless of real payload type.

**Fix:**
Using of `data.Ref` attribute instead, the problem is solved. Since its value corresponds to the real type of Payload `"[]*PayloadARequestBody"` resp. `"[]*PayloadBRequestBody"`. The code for the both `func NewPayloadARequestBody()` and `func NewPayloadBRequestBody()` is properly generated.

### DIFF
--- a/http/codegen/client_body_types_test.go
+++ b/http/codegen/client_body_types_test.go
@@ -2,12 +2,12 @@ package codegen
 
 import (
 	"bytes"
-	"goa.design/goa/v3/codegen/testutil"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
 	"goa.design/goa/v3/codegen"
+	"goa.design/goa/v3/codegen/testutil"
 	"goa.design/goa/v3/http/codegen/testdata"
 )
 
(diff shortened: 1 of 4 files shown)


raphael:
Nice catch! thanks for the nice PR.

abocim:
I'm sorry for failed checks before. I updated the source a little.
Now, the both unit tests and integreation tests are passing ok.

raphael:
No worries, thanks for the fix!

PR #3902, opened 2026-02-23
fix: reduce generator churn by removing goa version from file headers

disintegrator:
Remove the Goa version string from per-file generated headers (Go and proto) so that updating the goa dependency no longer causes every generated file to change. The version is now recorded once in gen/goa.json instead.

Example of this churn in the wild: https://github.com/speakeasy-api/gram/pull/1669/changes
Load the diff of various generated files in that PR and note that only the header file changed.

A nice bonus of this approach is that teams can add `gen/goa.json` to `.gitignore` and this will make certain dependabot PRs immediately mergeable if there are no changes relevant to their Goa usage in a given update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


codegen/generator/generate.go
```diff
@@ -156,13 +156,18 @@ func Generate(dir, cmd string, debug bool) (outputs []string, err1 error) {
...
+	}
+
+	// 9. Write the files (in parallel).
 	written := make(map[string]struct{})
 	{
 		start := time.Now()
 		numWorkers := runtime.NumCPU()
 		if debug {
-			fmt.Fprintf(os.Stderr, "[TIMING]     [generate] Stage 8: Starting parallel file writing with %d workers\n", numWorkers)
+			fmt.Fprintf(os.Stderr, "[TIMING]     [generate] Stage 9: Starting parallel file writing with %d workers\n", numWorkers)
```
tgmendes, line 170: Very nitpicky and probably irrelevant, but we skip from stage 7 to stage 9 in debug. Unsure if this won't be confusing in the future or not but worth pointing out.

Seems like we treat the new append of version file as a stage in our comments but doesn't reflect in the lines that we print out in debug.

tgmendes (approved):
Like 👍

codegen/generator/generate.go
disintegrator: This is aligned with how it was prior to this change it seems

raphael:
Makes sense, thank you!

PR #3905, opened 2026-02-25
Prefix interceptor var with service to avoid collision

nitinmohan87:
When multiple service designs contain the Interceptor DSL, the grpc CLI code generated contains an error with re-declaring same variable names.
```
Error: gen/grpc/cli/<service>/cli.go:48:2: inter redeclared in this block
Error: 	gen/grpc/cli/<service>/cli.go:47:2: other declaration of inter
```
Incorrect ParseEndpoint code with duplicate `inter` vars.
```
// ParseEndpoint returns the endpoint and payload as specified on the command
// line.
func ParseEndpoint(
	cc *grpc.ClientConn,
	inter <service_A>.ClientInterceptors,
	inter <service_B>.ClientInterceptors,
	opts ...grpc.CallOption,
) (goa.Endpoint, any, error) {
```

This fix will prefix the `inter` var with the service name to prevent collision.

### DIFF
--- a/codegen/cli/cli.go
+++ b/codegen/cli/cli.go
@@ -167,7 +167,7 @@ func BuildCommandData(data *service.Data) *CommandData {
 	var interceptors *InterceptorData
 	if len(data.ClientInterceptors) > 0 {
 		interceptors = &InterceptorData{
-			VarName: "inter",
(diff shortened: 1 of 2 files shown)


raphael:
This is great, thank you!

PR #3910, opened 2026-03-10
Support cookie-backed API key security inference

### LINKED ISSUE #3909: Support native cookie-backed APIKeySecurity for HTTP/OpenAPI
### Summary

Goa 3.25.3 does not appear to support expressing an HTTP `APIKeySecurity` scheme that is transported via request cookies in a way that is both runtime-correct and OpenAPI-correct.

We need a native Goa way to model authenticated browser routes that use a session cookie (for example `__Host-ak_session`) as the auth credential, instead of falling back to post-generation OpenAPI rewriting.

### What we expected

Given a design like this:

```go
var AuthSessionToken = JWTSecurity("auto_k_session_bearer", func() {
    Description("App bearer")
})

var BrowserSessionCookie = APIKeySecurity("browser_session_cookie", func() {
    Description("Browser session cookie")
})

var BrowserAuthenticatedRequest = Type("BrowserAuthenticatedRequest", func() {
    APIKey("browser_session_cookie", "browser_session", String, func() {
        Description("Opaque browser session cookie")
    })
    Token("auth", String, func() {
        Description("Bearer token")
    })
})

Method("profile", func() {
    Security(AuthSessionToken)
    Security(BrowserSessionCookie)
    Payload(BrowserAuthenticatedRequest)
    Result(User)
    HTTP(func() {
        GET("/auth/profile")
        Header("auth:Authorization")
        Cookie("browser_session:__Host-ak_session")
        Response(200)
    })
})
```

We expected Goa to:

1. Generate an OpenAPI security scheme for `browser_session_cookie` with `type: apiKey`, `in: cookie`, `name: __Host-ak_session`.
2. Generate the operation `security` array with cookie auth as an alternative requirement.
3. Not duplicate the auth cookie as a normal request parameter when it is already modeled as a security scheme.
4. Keep generated server/client code valid.

### What actually happens

In 3.25.3, Goa appears to resolve `APIKeySecurity` locations via `expr.findKey`, and `findKey` only checks query params, headers, and body. It does **not** check request cookies.

Relevant source:

- `expr/helpers.go`:

```go
func findKey(exp eval.Expression, keyAtt string) (string, string) {
    switch e := exp.(type) {
    case *HTTPEndpointExpr:
        if n, exists := e.Params.FindKey(keyAtt); exists {
            return n, "query"
        } else if n, exists := e.Headers.FindKey(keyAtt); exists {
            return n, "header"
        } else if e.Body == nil {
            return "", "header"
        }
        ...
        return "", "header"
    }
    return "", ""
}
```

Because cookies are ignored here, the generated OpenAPI/API key scheme ends up header-bound instead of cookie-bound.

### Practical consequences

In our case this caused two problems:

1. The generated OpenAPI modeled the browser session scheme as `header: Authorization` instead of `cookie: __Host-ak_session`.
2. When the same payload attribute was also mapped as a request cookie, generated server code ended up with duplicate locals/parameters for the same field and failed to compile.

Example shape from generated code:

```go
func NewBeginGithubLinkBrowserAuthenticatedOAuthLinkStartRequest(body *BeginGithubLinkRequestBody, auth *string, browserSession *string, browserSession *string) *restauth.BrowserAuthenticatedOAuthLinkStartRequest
```

and similarly duplicate local declarations in generated HTTP decode code.

### Current workaround

We had to keep the DSL/browser payloads using normal cookie request fields and then post-process the generated OpenAPI artifacts to:

1. add a cookie-based security scheme,
2. add cookie auth as an alternative security requirement on browser-authenticated operations,
3. remove the duplicate cookie request parameter,
4. normalize the 401 description away from bearer-specific wording.

That gets us a truthful served OpenAPI contract, but it is obviously not ideal.

### Request

Could Goa support a native way to express cookie-backed `APIKeySecurity` for HTTP endpoints?

Concretely, it would help if:

1. `APIKeySecurity` location inference considered HTTP cookies.
2. Cookie-backed security fields could be mapped without generating a duplicate cookie parameter.
3. Generated OpenAPI and generated transport code remained aligned for this case.

Even a documented/explicit DSL pattern for “API key security in cookie transport” would help a lot if there is already a supported way that we missed.

If useful, I can provide a smaller standalone reproducer, but the root cause looks pretty directly tied to `expr.findKey` ignoring `e.Cookies`.

CaliLuke:
## Summary
- make HTTP security key inference consider request cookies in addition to params, headers, and body
- add a regression test covering cookie-backed API key inference, OpenAPI v2/v3 output, and generated HTTP transport code

## Testing
- go test ./expr ./http/codegen/...

Fixes #3909

### DIFF
--- a/http/codegen/openapi/internal/security.go
+++ b/http/codegen/openapi/internal/security.go
@@ -1,4 +1,4 @@
-package openapi
+package internal
 
 import (
 	"strings"
--- a/http/codegen/openapi/v2/builder.go
+++ b/http/codegen/openapi/v2/builder.go
@@ -12,6 +12,7 @@ import (
 	"goa.design/goa/v3/codegen"
 	"goa.design/goa/v3/expr"
 	"goa.design/goa/v3/http/codegen/openapi"
+	openapiinternal "goa.design/goa/v3/http/codegen/openapi/internal"
 )
 
 // NewV2 returns the OpenAPI v2 specification for the given API.
@@ -35,7 +36,7 @@ func NewV2(root *expr.RootExpr, h *expr.HostExpr) (*V2, error) {
 	if hasAbsoluteRoutes(root) {
 		basePath = ""
 	}
-	params := paramsFromExpr(root.API.HTTP.Params, basePath)
+	params := paramsFromExpr(nil, root.API.HTTP.Params, basePath)
 	var paramMap map[string]*Parameter
 	if len(params) > 0 {
 		paramMap = make(map[string]*Parameter, len(params))
@@ -269,11 +270,7 @@ func summaryFromMeta(name string, meta expr.MetaExpr) string {
 	return name
 }
 
-func paramsFromExpr(params *expr.MappedAttributeExpr, path string) []*Parameter {
-	return paramsFromExprForEndpoint(nil, params, path)
-}
-
-func paramsFromExprForEndpoint(endpoint *expr.HTTPEndpointExpr, params *expr.MappedAttributeExpr, path string) []*Parameter {
+func paramsFromExpr(endpoint *expr.HTTPEndpointExpr, params *expr.MappedAttributeExpr, path string) []*Parameter {
 	if params == nil {
 		return nil
 	}
@@ -287,7 +284,7 @@ func paramsFromExprForEndpoint(endpoint *expr.HTTPEndpointExpr, params *expr.Map
 			in = "path"
 			required = true
 		}
-		if endpoint != nil && in != "path" && openapi.IsSecurityParameter(endpoint, in, pn) {
+		if endpoint != nil && in != "path" && openapiinternal.IsSecurityParameter(endpoint, in, pn) {
 			return nil
 		}
 		param := paramFor(at, pn, in, required)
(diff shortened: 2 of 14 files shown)


http/codegen/openapi/internal/security.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"strings"
+
+	"goa.design/goa/v3/expr"
+)
+
+// IsSecurityParameter returns true if the given HTTP transport element is used
+// by one of the endpoint security schemes and should therefore not be emitted
+// again as a regular OpenAPI parameter.
+func IsSecurityParameter(endpoint *expr.HTTPEndpointExpr, in, name string) bool {
```
raphael, line 12: Might be better to keep this private, this is probably not something that the openapi package should expose.

http/codegen/openapi/v2/builder.go
```diff
@@ -270,6 +270,10 @@ func summaryFromMeta(name string, meta expr.MetaExpr) string {
 }
 
 func paramsFromExpr(params *expr.MappedAttributeExpr, path string) []*Parameter {
+	return paramsFromExprForEndpoint(nil, params, path)
```
raphael: Can we avoid this wrapping? I don't know why LLMs insist in making these :)

raphael (commented):
Look great! thank you for the PR. I left a couple of minor comments that would be good to address before we can merge this in.

http/codegen/openapi/v2/builder.go
CaliLuke: they have their quirks! I also made a small amendment to AGENTS.md in the same commit to attempt avoiding the same type of behavior in the future.

raphael: Awesome, thank you!

PR #3921, opened 2026-05-01
service output directory

egandro:
This is a minor `goa example` enhancement.

`--service-output` allows the services to be created e.g. in `pkg/services`

This separates the files e.g. that we create in `cmd`.

example

```bash
goa example --service-output ./pkg/services
```

### DIFF
--- a/cmd/goa/gen.go
+++ b/cmd/goa/gen.go
@@ -30,6 +30,9 @@ type Generator struct {
 	// Output is the absolute path to the output directory.
 	Output string
 
+	// ServiceOutput is the absolute path to the example service output directory.
+	ServiceOutput string
+
 	// DesignVersion is the major component of the Goa version used by the design DSL.
 	// DesignVersion is either 2 or 3.
 	DesignVersion int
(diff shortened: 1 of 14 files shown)


egandro:
I keep this as draft, because I would love to get some thumbs up or down.

I didn't like the service.go files flying at the top of the project

raphael:
Hello, that makes sense, there are a couple of considerations that come to mind:

1. We've been trying to keep `goa` as flag-less as possible to keep its usage super simple, tweaks to how the generators work are typically done via the [Meta](https://pkg.go.dev/goa.design/goa/v3/dsl#Meta) DSL.

2. This is changing the path to the service Go package which has the potential to break other generated code that depend on it (e.g. transport, plugins). So we'd probably need some extensive tests for this that cover not just where the files are generated but also ensure that import statements are properly generated for code that depend on them.

egandro:
1) Ok that makes sense. I was missing some config object. Adding more paths to all of these functions is bad style.

If we go that route and do a refactoring - I suggest a better architecture - just in case we need 2-3 more flags e.g. add some AI agent call 🙈 I know this is stupid - sorry I opened Pandoras box here.

2) Probably that config object will make (more sense) in downstream layers, as they can relay now on paths.

I don't know if we should go down this route but the missing path for the service files bugs me for a while. I fooled around in Makefiles with mv hacks. This became worse now. I just ask for some consent to give the output and code generators a bit more flexibility (probably I just don't know enough about the templates).

raphael:
Thanks again for kicking this off and for the discussion. I'm going to close this one: we settled on not adding flags to `goa example`, and the v3.31 generator rework (#3971) changes how output paths are planned, so this diff no longer applies. If the service layout still bothers you once v3.31 is out, a `Meta`-driven proposal would be the right way to pick it back up. 👍

PR #3925, opened 2026-05-13
Add first-class Bearer security

### LINKED ISSUE #3923: Add first-class Bearer security
Goa can already generate support for HTTP `Authorization: Bearer <token>`, but today that support is only available through `JWTSecurity`. This makes it awkward to model APIs that accept opaque Bearer tokens. These APIs are not JWT-specific, but their Goa designs still need to use JWT terminology in the DSL and the generated auth hooks. As a result, the implementation ends up exposing JWT-specific names even when the public OpenAPI contract is just standard HTTP Bearer authentication.

It would be useful to have a `BearerSecurity` DSL for modeling generic Bearer authentication without JWT-specific terminology. Ideally, this would also include matching Bearer token payload helpers.

For OpenAPI v3, `BearerSecurity` should generate:

- `type: http`
- `scheme: bearer`
- `bearerFormat` only when the design explicitly sets one

`BearerSecurity` and `JWTSecurity` could share the same default HTTP wire format:

`Authorization: Bearer <token>`

The main difference would be semantic and reflected in the generated names. `BearerSecurity` would generate Bearer-oriented hooks for generic or opaque Bearer tokens, while `JWTSecurity` would continue to generate JWT-oriented hooks for designs that explicitly model JWT Bearer tokens.

sevein:
Add `BearerSecurity` and `BearerToken` helpers for APIs that use the standard `Authorization: Bearer` wire format without JWT-specific DSL and generated names. Wire the new scheme through expr validation, HTTP/gRPC transport inference, service auth codegen, runtime security types, and OpenAPI output.

`BearerSecurity` models generic bearer tokens while JWTSecurity keeps JWT-oriented generated names. Both use the same default HTTP Bearer header and both render as HTTP bearer schemes in OpenAPI v3.

Keep Goa scopes available to generated auth code, but do not emit scopes on OpenAPI v3 HTTP bearer security requirements. OpenAPI only permits non-empty requirement scope arrays for OAuth2/OpenID Connect schemes; HTTP bearer requirements must use an empty array.

Fixes https://github.com/goadesign/goa/issues/3923.


raphael:
Looks great! thank you.

PR #3930, opened 2026-05-25
Honor NoSecurity in OpenAPI

sevein:
Preserve explicit NoSecurity declarations after method finalization so the OpenAPI generators can distinguish them from methods that simply have no local security requirement.

For example, an API can define `Security(JWTAuth)` globally while a health or status method calls `NoSecurity()`. That method must emit an operation-level security: [] entry so generated OpenAPI v2 and v3 documents do not make clients inherit the global JWT requirement for that public endpoint.

### DIFF
--- a/codegen/service/service_data.go
+++ b/codegen/service/service_data.go
@@ -1345,7 +1345,7 @@ func (d *ServicesData) buildMethodData(m *expr.MethodExpr, scope *codegen.NameSc
 		}
 	}
 
-	for _, req := range m.Requirements {
+	for _, req := range expr.EffectiveSecurityRequirements(m.Requirements) {
 		var rs SchemesData
 		for _, s := range req.Schemes {
 			sch := BuildSchemeData(s, m)
--- a/expr/grpc_endpoint.go
+++ b/expr/grpc_endpoint.go
@@ -289,9 +289,10 @@ func (e *GRPCEndpointExpr) Finalize() {
 
 		// Initialize any security attributes in request metadata unless it is
 		// specified explicitly in the request message via the DSL.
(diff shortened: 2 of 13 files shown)


raphael:
Thanks for chasing this down, this is definitely the behavior I’d expect from `NoSecurity()`. A public health check inheriting the API-level auth requirement in OpenAPI is the kind of tiny spec bug that creates a very confusing afternoon.

One thought: I think we already have a little breadcrumb for this in the expr layer. `NoSecurity()` records a `SecurityExpr` with a `SchemeExpr{Kind: expr.NoKind}`, then `MethodExpr.Finalize()` consumes that marker and clears `Requirements`. That seems like the moment where we lose the distinction between “explicitly no security” and “no local security requirement”.

Could we keep that existing marker as the canonical source of truth, instead of adding a second public `MethodExpr.NoSecurity` flag plus the mirrored `Operation.NoSecurity` flag? For example, maybe finalization can preserve/derive the no-security fact from the `NoKind` requirement, or expose a tiny query that answers that question without storing it twice.

Small related nit: injecting `security` through the extensions merge path works, but it feels a little surprising since `security` is a real OpenAPI field rather than an extension. If we can make the empty operation-level security requirement part of the normal operation shape, I think the intent will be easier to follow later.

sevein:
Thanks, Raphael! I updated it to keep the existing `NoKind` requirement as the source of truth instead of adding separate `NoSecurity` flags, with small helpers in `expr` for checking/effective requirements. I also moved `security: []` back into the normal OpenAPI `Operation.Security` field by using a named `SecurityRequirements` type that can distinguish nil from empty, and added v2/v3 regression tests for both API-level and service-level inherited security.

raphael:
Perfect, thank you!

PR #3933, opened 2026-06-01
codegen: memoize GetMetaTypeImports to avoid redundant type-graph walks

rorydbain:
## Summary

`GetMetaTypeImports` walks an attribute's entire type graph via
`safelyGetMetaTypeImports` to collect `struct:field:type` imports. The service
and transport generators resolve imports for the same method payloads and
results repeatedly, so on large designs this walk dominates generation time —
and on designs that use no `struct:field:type` metadata at all, it's pure
overhead that still walks every type.

This memoizes `GetMetaTypeImports` by attribute identity.

## Why this is safe

`safelyGetMetaTypeImports(att, nil)` returns the set of `struct:field:type`
imports reachable from `att` — a transitive-closure property that's independent
of traversal entry point (the `seen` map only prevents infinite recursion; it
never drops a reachable node). So a memoized result always equals a fresh
computation.

The cache is keyed on the `*expr.AttributeExpr` pointer rather than the type
name/ID, so entries can't collide across different designs generated in the same
process (e.g. the test suite). `sync.Map` keeps it safe if a caller ever resolves
imports concurrently; today all callers run in the serial generator phase.

## Output is unchanged

- `go test ./...` and the integration tests pass — the codegen golden-file tests
  cover the service, http, grpc and jsonrpc generators.
  A new test (`TestGetMetaTypeImportsAcrossDesigns`) pins the keying: two designs
  with a same-named type each resolve their own imports — it fails if the cache
  is keyed by type name instead of attribute identity.
- Regenerating `goadesign/examples` (`basic`) produces byte-identical output
  patched vs unpatched.
- A large production design (347 design files, 3671 generated files)
  regenerates byte-identical, with end-to-end `goa gen` dropping from ~39s to
  ~23s — the serial generator phase drops from 28.8s to 13.5s.

A focused micro-benchmark of the import resolution a generator pass performs
measured -63% time / -66% allocations. I left the benchmark out since the repo
has no existing benchmarks; happy to add it if you'd like one in-tree.

### DIFF
--- a/codegen/import.go
+++ b/codegen/import.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strconv"
+	"sync"
 
 	"goa.design/goa/v3/expr"
 	goa "goa.design/goa/v3/pkg"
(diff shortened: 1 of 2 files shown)


raphael:
Looks like we had similar ideas, is this covered by  https://github.com/goadesign/goa/pull/3932?

rorydbain:
> Looks like we had similar ideas, is this covered by #3932?

Yes that does! Apologies this was all Claude, hence it still being in draft, but that looks like it solves our problem - nice!

rorydbain:
> Looks like we had similar ideas, is this covered by #3932?

When are you thinking of cutting a new release for that @raphael?

raphael:
I'll release this today

PR #3934, opened 2026-06-03
 Fix race in http.ErrorEncoder

duckbrain:
Moves the nil `formatter` handler outside of the closure, so the returned encoder is safe concurrently when `nil` is passed in.


raphael:
Good catch, thank you!

PR #3947, opened 2026-07-20
fix: websocket client codegen

abocim:
Use named struct fields for result types in WebSocket client streaming

Update websocket_recv.go.tpl and test data to initialize viewed result types with named fields (`Projected:` and `View:`) instead of positional arguments.

This fixes go vet validation failures in generated WebSocket code.


raphael:
Awesome, thank you!

PR #3973, opened 2026-08-27
Fix go 1.27 panic in v1

dmlyons:
Go 1.27 now panics `panic: jsontext: invalid nil io.Reader` when NewEncoder is passed nil because v1 json is now backed by v2 json.

This commit changed the behavior:
https://github.com/golang/go/commit/0e17905793cb5e0acc323a0cdf3733199d93976a

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -1,6 +1,7 @@
 package goa
 
 import (
+	"bytes"
 	"encoding/gob"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


raphael:
Makes sense thank you! You are probably aware but Goa v1 is quite old, you might want to consider upgrading to v3.

PR #3980, opened 2026-09-01
fix(preview): reuse the generated collection type for suffixed identifiers

### LINKED ISSUE #3977: [Preview Report] `CollectionOf` creates a duplicate result type, so calling it twice in one service fails name planning
Runnable reproduction: https://github.com/ikawaha/goa.preview/tree/main/02-collection-name-collision

## Summary

Calling `CollectionOf` twice for the same result type inside one service fails generation on
`v3.31.0-preview.1` and `v3.31.0-preview.3`:

```
declare user type "ReservationCollection": generated package "repro/gen/reservation"
cannot declare exact type "ReservationCollection": already declared by exact type
```

The design declares no duplicate. `CollectionOf` has a cache that should reuse the existing
collection type, but the lookup compares a canonicalized identifier against raw stored identifiers,
so it never matches when the identifier has a suffix such as `+json`. Every call therefore creates
a new result type, and the new name planner reports the two as a collision.

## Environment

- Go: 1.27.0, darwin/arm64
- Transport: HTTP only
- Both client and server are regenerated by the command below (full `goa gen`)
- `protoc` versions: not applicable, no gRPC

## Reproduction

`design/design.go`:

```go
package design

var RT = ResultType("application/vnd.reservation+json", func() {
	Attributes(func() {
		Attribute("id", Int)
		Required("id")
	})
})

var _ = Service("reservation", func() {
	Method("search1", func() {
		Result(CollectionOf(RT))
		HTTP(func() {
			GET("/a")
		})
	})
	Method("search2", func() {
		Result(CollectionOf(RT))
		HTTP(func() {
			GET("/b")
		})
	})
})
```

Command:

```sh
```

## Complete error

```
declare user type "ReservationCollection": generated package "repro/gen/reservation" cannot declare exact type "ReservationCollection": already declared by exact type
exit status 1
```

## Versions

| Goa | Result |
|:--|:--|
| v3.31.0-preview.1 | failed |
| v3.31.0-preview.3 | failed |
| v3.30.0, v3.29.2, v3.28.0, v3.27.0 | generates successfully |

| Design | preview | released versions |
|:--|:--|:--|
| two methods in one service, `CollectionOf` called twice | **failed** | OK |
| same, but the `CollectionOf` value is stored in a variable and shared | OK | OK |
| two services, `CollectionOf` called once in each | OK | OK |
| identifier without a `+json` suffix, `CollectionOf` called twice | OK | OK |

The last row isolates the cause: the only difference is the media type suffix.

## Cause

`dsl/result_type.go` looks the collection up by canonical identifier:

```go
canonical := expr.CanonicalIdentifier(id)   // "application/vnd.reservation; type=collection"
if mt := expr.GeneratedResultType(canonical); mt != nil {
	// Already have a type for this collection, reuse it.
	return mt
}
```

`expr/result_types_root.go` compares that key against raw stored identifiers:

```go
func GeneratedResultType(id string) *ResultTypeExpr {
	for _, rt := range *GeneratedResultTypes {
		if rt.Identifier == id {   // stored raw: "application/vnd.reservation+json; type=collection"
			return rt
		}
	}
	return nil
}
```

`CanonicalIdentifier` strips the `+json` suffix while the stored identifiers keep it, so the
comparison can never match. Instrumented output from the reproduction:

```
lookup key   : "application/vnd.reservation; type=collection"
stored entry : "application/vnd.reservation+json; type=collection"  (canonical: "application/vnd.reservation; type=collection")
-> no match, creating a new type
```

The two resulting expressions are distinct values that both want the Go name
`ReservationCollection`:

```
new     : ResultType ptr=0x...31d0 name="ReservationCollection" identifier="application/vnd.reservation+json; type=collection" origin=self
existing: ResultType ptr=0x...3130 name="ReservationCollection" identifier="application/vnd.reservation+json; type=collection" origin=self
```

`GeneratedPackage.DeclareUserType` (`codegen/generated_types.go`) deduplicates by
`userType.Origin()`, so the two distinct expressions are not merged and the second `DeclareName`
collides.

This code is identical in v3.30, where the duplicate was absorbed later during generation. The
preview assigns names up front and reports the collision, which surfaces the pre-existing cache
bug.

This also does not match the documented behavior for collisions. `codegen/ARCHITECTURE.md` states
that "real name collisions receive stable numeric suffixes instead of suffixes determined by
discovery order", and describes failing generation only for unions: "Goa never adds a numeric suffix
to resolve a collision. Separately authored unions that ask for the same family name now fail
generation". This design has no union and declares no duplicate; the duplicate is created inside
`CollectionOf`.

## Suggested fix

Compare canonical identifiers on both sides. There are two callers: `dsl/result_type.go` passes a
canonicalized value and `expr/dup.go:152` passes a raw one, so canonicalizing inside the lookup
keeps both correct.

```diff
 func GeneratedResultType(id string) *ResultTypeExpr {
+	canonical := CanonicalIdentifier(id)
 	for _, rt := range *GeneratedResultTypes {
-		if rt.Identifier == id {
+		if CanonicalIdentifier(rt.Identifier) == canonical {
 			return rt
 		}
 	}
 	return nil
 }
```

## Generated output with the fix

For the design above, `gen/reservation/service.go` becomes byte-identical to the file generated by
v3.30.0: a single shared declaration.

```go
type ReservationCollection []*Reservation
```

The remaining differences against v3.30.0 in other files are the preview's intentional changes
(response validation, joined close errors), not name changes.

Also verified with the fix:

- the Goa test suite shows no new failures (the same set of unrelated pre-existing failures)
- two services of a real HTTP project that previously failed here now generate

## Workaround

Call `CollectionOf` once and share the value:

```go
var RTC = CollectionOf(RT)

Method("search1", func() { Result(RTC) /* ... */ })
Method("search2", func() { Result(RTC) /* ... */ })
```

ikawaha:
close #3977 

`CollectionOf` looks its cache up with a canonical identifier while generated result types are recorded with their authored identifier. A result type whose identifier carries a suffix such as `+json` therefore never matched, and every call created another collection type. Two declarations then asked for one generated Go name and generation stopped:

    declare user type "ItemCollection": generated package "gen/item" cannot
    declare exact type "ItemCollection": already declared by exact type

The design declares no duplicate: the second type is created inside `CollectionOf`.

* Compare canonical identifiers on both sides of the lookup. The two callers pass a canonical and a raw identifier respectively, so the comparison belongs in the lookup itself.
* Add an `expr` test for the lookup and a `dsl` test that calls `CollectionOf` twice for one result type.

For the design in #3977, `gen/reservation/service.go` is byte-identical to the file generated by v3.30.0.

### DIFF
--- a/dsl/collection_of_test.go
+++ b/dsl/collection_of_test.go
@@ -1,0 +1,41 @@
+package dsl
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"goa.design/goa/v3/expr"
+)
+
+// TestCollectionOfReusesGeneratedType verifies that CollectionOf reuses the
+// generated collection type when it is called more than once for the same
+// result type. The lookup is made with a canonical identifier, so a result type
+// whose identifier carries a suffix such as "+json" must still match. Creating a
+// second type makes two declarations ask for one generated Go name.
+func TestCollectionOfReusesGeneratedType(t *testing.T) {
+	var first, second *expr.ResultTypeExpr
+	expr.RunDSL(t, func() {
+		rt := ResultType("application/vnd.item+json", func() {
+			Attributes(func() {
+				Attribute("id", expr.Int)
+				Required("id")
+			})
+		})
+		Service("svc", func() {
+			Method("first", func() {
+				first = CollectionOf(rt)
+				Result(first)
+				HTTP(func() { GET("/first") })
+			})
+			Method("second", func() {
+				second = CollectionOf(rt)
+				Result(second)
+				HTTP(func() { GET("/second") })
+			})
+		})
+	})
+
+	assert.Same(t, first, second)
+	assert.Len(t, *expr.GeneratedResultTypes, 1)
+}
--- a/expr/result_types_root.go
+++ b/expr/result_types_root.go
@@ -15,10 +15,13 @@ type (
 var GeneratedResultTypes = new(ResultTypesRoot)
 
 // GeneratedResultType returns the generated result type expression with the given
-// id, nil if there isn't one.
+// id, nil if there isn't one. Identifiers are compared canonically so that a
+// lookup by canonical identifier finds a type recorded with its authored
+// suffix, such as "+json".
 func GeneratedResultType(id string) *ResultTypeExpr {
+	canonical := CanonicalIdentifier(id)
 	for _, rt := range *GeneratedResultTypes {
-		if rt.Identifier == id {
+		if CanonicalIdentifier(rt.Identifier) == canonical {
 			return rt
 		}
 	}
--- a/expr/result_types_root_test.go
+++ b/expr/result_types_root_test.go
@@ -1,0 +1,27 @@
+package expr
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGeneratedResultTypeMatchesCanonicalIdentifier verifies that a lookup by
+// canonical identifier finds a generated result type recorded with its authored
+// media type suffix. CollectionOf looks its cache up with a canonical
+// identifier, so a suffixed identifier must not create a second type.
+func TestGeneratedResultTypeMatchesCanonicalIdentifier(t *testing.T) {
+	ResetDSL(t)
+
+	suffixed := "application/vnd.item+json; type=collection"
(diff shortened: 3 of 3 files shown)


raphael:
Thanks @ikawaha, putting the canonical comparison in the lookup itself is the right place for it. Merged into the preview branch, it will be in the next v3.31 preview. 👍

PR #3979, opened 2026-09-01
fix(preview): fix required array element scope for viewed results

ikawaha:
close #3976 

Generating the empty slice fallback for a required array field spelled the element type with the outer transform context. When the enclosing type is a result type that context still selects a view, while nested types are always planned without one, so resolving the element type found no planned declaration and generation panicked:

    panic: resolve HTTP type "Elem": plan Go named type "Elem":
           HTTP type "Elem" has no planned declaration

* Use the field context, which every other call in the same block already uses.
* Add server type tests for a required array of a user type, a required array of arrays, and a result type with explicit views.

PR #3981, opened 2026-09-01
fix(preview): stop planning conversions for unrendered responses

### LINKED ISSUE #3978: [Preview Report] Two untagged success responses: a planned conversion is never rendered
Runnable reproduction: https://github.com/ikawaha/goa.preview/tree/main/03-multiple-success-responses

## Summary

A method whose HTTP mapping declares two success responses without `Tag` stops generation on
`v3.31.0-preview.1` and `v3.31.0-preview.3`:

```
HTTP marshal conversion for m server response 202   default was planned but not rendered
```

Goa handles only the first untagged response, but the loop that collects wire conversions plans one
for every response, so the conversion planned for the second response is never rendered.

## Environment

- Released versions checked: `v3.30.0`, `v3.29.2`, `v3.28.0`, `v3.27.0` (all generate successfully)
- Go: 1.25.0 and 1.27.0, darwin/arm64 (both reproduce)
- Transport: HTTP only, so `protoc` versions do not apply
- The reproduction runs a complete `goa gen`, so client and server are regenerated together

## Reproduction

```go
package design

var RT = ResultType("application/vnd.repro+json", func() {
	Attributes(func() {
		Attribute("name", String)
		Required("name")
	})
})

var _ = Service("svc", func() {
	Method("m", func() {
		Result(RT)
		HTTP(func() {
			POST("/")
			Response(StatusOK)
			Response(StatusAccepted)
		})
	})
})
```

```sh
```

| Design | preview | released |
|:--|:--|:--|
| two untagged success responses, result type | **failed** | OK |
| two untagged success responses, plain type | **failed** | OK |
| two success responses selected by `Tag` | OK | OK |
| single success response | OK | OK |

The result type does not matter, so views are not involved.

## Cause

`http/codegen/plan.go:1798` keeps only the first untagged response when it declares the client
result constructors:

```go
noTagSeen := false
for _, response := range endpoint.Responses {
	if response.Tag[0] == "" {
		if noTagSeen {
			continue
		}
		noTagSeen = true
	}
	...
}
```

The loop that collects the wire conversions, `http/codegen/service_data.go:1610`, has no such skip:

```go
for _, response := range endpoint.Responses {
	body := bodies.response(response)
	...
}
```

Instrumenting both sides shows the asymmetry:

```
[COLLECT] method=m status=200 tag=[] view="default"
[COLLECT] method=m status=202 tag=[] view="default"   <- planned
[LOOKUP ] method=m status=200 tag=[] view="default" found=true
                                                      <- 202 is never looked up
```

The keys match on both sides. The problem is that a conversion is planned for a response the
generator does not render.

ikawaha:
close #3978

Generation renders only the first response without a tag, but the loop that collects wire conversions planned one for every response. The conversion planned for a second untagged response was never rendered, and generation stopped:

    HTTP marshal conversion for m server response 202   default was planned but not rendered

The client result constructors already keep only the first untagged response (`http/codegen/plan.go:1798`); the collecting loop in `http/codegen/service_data.go` did not, so the two sides disagreed about which responses exist.

* Apply the same rule when collecting the conversions.
* Add a server encode test for two untagged success responses.

Responses selected with `Tag` are unaffected: every tagged response is still collected, and a design that distinguishes `StatusCreated` and `StatusAccepted` by tag generates the same conditional encoder as before. The generated server encoder for two untagged responses is byte-identical to the one produced by v3.30.0, which also renders only the first of them.


raphael:
Thanks @ikawaha, aligning the collection loop with the constructor rule is exactly right. I merged the preview branch into your branch to resolve a trivial overlap in the test DSL file with #3979, everything else is untouched. Merged into the preview branch, it will be in the next v3.31 preview. 👍

PR #3983, opened 2026-09-01
fix(preview): settle an explicit body attribute before a selected view

### LINKED ISSUE #3982: [Preview Report] A response body set to one result attribute panics with a nil pointer dereference
Runnable reproduction: https://github.com/ikawaha/goa.preview/tree/main/04-explicit-response-body

## Summary

`goa gen` panics on `v3.31.0-preview.1` and `v3.31.0-preview.3` when a response sets its body to one
attribute of a result type. One response is enough, and the same design with a plain `Type` result
generates.

Two switches decide the same thing and order their cases differently, so the planned response
transform is recorded under one view and looked up under another. The miss yields a nil pointer that
is dereferenced on the next line.

## Environment

- Released versions checked: `v3.30.0`, `v3.29.2`, `v3.28.0`, `v3.27.0` (all generate successfully)
- Go: 1.25.0 and 1.27.0, darwin/arm64 (both reproduce)
- Transport: HTTP only, so `protoc` versions do not apply
- The reproduction runs a complete `goa gen`, so client and server are regenerated together

## Reproduction

```go
package design

var RT = ResultType("application/vnd.repro+json", func() {
	Attributes(func() {
		Attribute("name", String)
		Required("name")
	})
})

var _ = Service("svc", func() {
	Method("m", func() {
		Result(RT)
		HTTP(func() {
			POST("/")
			Response(StatusOK, func() {
				Body("name")
			})
		})
	})
})
```

```sh
```

## Complete error

```
panic: runtime error: invalid memory address or nil pointer dereference

goroutine 1 [running]:
goa.design/goa/v3/http/codegen.(*ServicesData).buildResponseResultInit(...)
	http/codegen/service_data.go:3216
goa.design/goa/v3/http/codegen.(*ServicesData).buildResponses(...)
	http/codegen/service_data.go:3044
goa.design/goa/v3/http/codegen.(*ServicesData).buildResultData(...)
	http/codegen/service_data.go:2894
goa.design/goa/v3/http/codegen.(*ServicesData).analyze(...)
	http/codegen/service_data.go:1263
goa.design/goa/v3/http/codegen.(*Plan).link(...)
	http/codegen/plan.go:1968
```

| Design | preview | released |
|:--|:--|:--|
| response `Body(attribute)`, result type | **panic** | OK |
| response `Body(attribute)` holding a user type, result type | **panic** | OK |
| response `Body(attribute)`, result type with explicit views | OK | OK |
| response `Body(attribute)`, plain type | OK | OK |
| no response `Body`, result type | OK | OK |

## Cause

The collecting side, `collectPlannedTransforms` in `http/codegen/service_data.go`, settles an
explicit body attribute before a selected view:

```go
selected := clientResponseViewNameExpr(endpoint, resultType)
switch {
case origin != "":            // Body("name")
	empty := ""
	clientViews = append(clientViews, &empty)
case selected != "":
	clientViews = append(clientViews, &selected)
case explicitBody:            // Body(func() { ... })
	empty := ""
	clientViews = append(clientViews, &empty)
default:
	...
}
```

The consuming side, `buildResponses`, checks the selected view first:

```go
switch {
case clientView != "":
	...
	clientBodyView = &clientView
case origin != "" || explicitBody:
	...
	clientBodyView = &vname
default:
	...
}
```

`Body("name")` sets both `origin` and a selected client view, so the two sides disagree.
`responses` is `map[viewedConstructorKey]*plannedResponseTransforms` (`http/codegen/plan.go:718`),
so the miss at `service_data.go:3215` yields a nil pointer that is dereferenced on the next line.
Instrumenting both sides:

```
[COLLECT] method=m status=200 tag=[] view=""
[LOOKUP ] method=m status=200 tag=[] view="default" found=false
```

The rest of the matrix follows: with explicit views the selected view is empty, so the order does
not matter; a plain type has no views; and without an explicit body `origin` is empty, so both sides
take the same branch.

## Suggested fix

Move only the `origin` case ahead of the selected view on the consuming side, so both switches
settle an explicit body attribute first:

```diff
 					switch {
+					case origin != "":
+						clientBodyData = sds.buildResponseBodyType(respBody, result, e, false, &vname, sd, nil, resultOwner, bodyOwner)
+						clientBodyView = &vname
 					case clientView != "":
 						clientRespBody = effectiveClientResponseBodyForView(respBody, clientView, e)
 						clientBodyData = sds.buildResponseBodyType(clientRespBody, result, e, false, &clientView, sd, nil, resultOwner, bodyOwner)
 						clientBodyView = &clientView
-					case origin != "" || explicitBody:
+					case explicitBody:
 						clientBodyData = sds.buildResponseBodyType(respBody, result, e, false, &vname, sd, nil, resultOwner, bodyOwner)
 						clientBodyView = &vname
```

Moving `explicitBody` as well breaks `TestBodyTypeInit/result-explicit-body-object`: an inline
`Body(func() { ... })` sets `explicitBody` without `origin`, and collection settles that case after
the selected view. Only the `origin` case belongs in front.

Verified with this change:

- all five designs in the matrix generate
- the generated client decoder is the same as the one produced by v3.30.0
- the Goa test suite shows no new failures

## Why existing tests did not catch it

`TestBodyTypeInit` already covers explicit response bodies, but every one of those designs declares
its views (`ExplicitBodyPrimitiveResultMultipleViewsDSL` and friends). With explicit views the
selected client view is empty, which is exactly the case where the order of the two switches does
not matter. A result type that declares no view takes the failing path.

ikawaha:
close #3982

Two switches decide which view a response body belongs to, and they ordered their cases differently. Collection settles `Body("name")` first, while the consuming side checked the selected view first. A result type that declares no view sets both, so the transform was recorded under an empty view and looked up under the default one. The lookup returned a nil record that was dereferenced:

    panic: runtime error: invalid memory address or nil pointer dereference
           http/codegen/service_data.go:3216

Instrumenting both sides:

    [COLLECT] method=m status=200 tag=[] view=""
    [LOOKUP ] method=m status=200 tag=[] view="default" found=false

* Move only the `origin` case ahead of the selected view. An inline `Body(func() { ... })` sets `explicitBody` without `origin`, and collection settles that case after the selected view, so it stays where it is.
* Add a client body type test for a result type that declares no view.

The existing explicit body cases in `TestBodyTypeInit` all declare their views, which is the path where the order of the two switches does not matter, so none of them covered this. The generated client decoder for the design in #3982 is the same as the one produced by v3.30.0.

### DIFF
--- a/http/codegen/client_body_types_test.go
+++ b/http/codegen/client_body_types_test.go
@@ -46,6 +46,7 @@ func TestBodyTypeInit(t *testing.T) {
 		{"result-body-user-required", testdata.ResultBodyUserRequiredDSL, 3},
 		{"result-body-inline-object", testdata.ResultBodyInlineObjectDSL, 2},
 		{"result-explicit-body-primitive", testdata.ExplicitBodyPrimitiveResultMultipleViewsDSL, 1},
+		{"result-explicit-body-primitive-default-view", testdata.ExplicitBodyPrimitiveResultDefaultViewDSL, 1},
 		{"result-explicit-body-user-type", testdata.ExplicitBodyUserResultMultipleViewsDSL, 3},
 		{"result-explicit-body-object", testdata.ExplicitBodyUserResultObjectDSL, 3},
 		{"result-explicit-body-object-views", testdata.ExplicitBodyUserResultObjectMultipleViewDSL, 3},
--- a/http/codegen/service_data.go
+++ b/http/codegen/service_data.go
@@ -3000,11 +3000,14 @@ func (sds *ServicesData) buildResponses(e *expr.HTTPEndpointExpr, result *expr.A
 						}
 					}
 					switch {
+					case origin != "":
+						clientBodyData = sds.buildResponseBodyType(respBody, result, e, false, &vname, sd, nil, resultOwner, bodyOwner)
+						clientBodyView = &vname
 					case clientView != "":
 						clientRespBody = effectiveClientResponseBodyForView(respBody, clientView, e)
 						clientBodyData = sds.buildResponseBodyType(clientRespBody, result, e, false, &clientView, sd, nil, resultOwner, bodyOwner)
 						clientBodyView = &clientView
-					case origin != "" || explicitBody:
+					case explicitBody:
 						clientBodyData = sds.buildResponseBodyType(respBody, result, e, false, &vname, sd, nil, resultOwner, bodyOwner)
 						clientBodyView = &vname
 					default:
--- a/http/codegen/testdata/golden/client_body_type_init_result-explicit-body-primitive-default-view.go.golden
+++ b/http/codegen/testdata/golden/client_body_type_init_result-explicit-body-primitive-default-view.go.golden
@@ -1,0 +1,13 @@
+// NewMethodExplicitBodyPrimitiveResultDefaultViewResulttypedefaultviewOK
+// builds a "ServiceExplicitBodyPrimitiveResultDefaultView" service
+// "MethodExplicitBodyPrimitiveResultDefaultView" endpoint result from a HTTP
+// "OK" response.
+func NewMethodExplicitBodyPrimitiveResultDefaultViewResulttypedefaultviewOK(body string, b *string) *serviceexplicitbodyprimitiveresultdefaultviewviews.ResulttypedefaultviewView {
+	v := body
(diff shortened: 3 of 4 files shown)


raphael:
Thanks @ikawaha, nice catch on the switch ordering and a clean fix. Merged into the preview branch, it will be in the next v3.31 preview. 👍

PR #3994, opened 2026-09-18
add support for path params in middle of the path

dvjn:
Refreshed version of #3713, opened against `master` after v3.31.1, with the tests requested during review.

Adds support for path parameters that occupy part of a path segment.

Before this change, `/user/@{username}` did not register `username` as a path parameter. The chi [README example](https://github.com/go-chi/chi?tab=readme-ov-file#examples) `/{month}-{day}-{year}` detected only `{month}`. Both patterns now work end to end.

### Changes

`HTTPWildcardRegex` no longer requires a leading `/`. Each call site that relied on the match containing the slash has been updated. Two call sites were added after the original PR:

- `routePatternWithoutParameterNames` in `expr/root.go`
- File server path handling in the OpenAPI v3 builder

The `Sprintf` call in the v2 builder now slices `w[1:]` instead of `w[2:]`.

`http/mux.go` needs no change. Its `wildPath` regex matches only catch-all parameters such as `{*name}`. Chi handles mid-segment parameters natively.

### Tests

`http/mux_test.go` adds four `TestVars` cases. The cases route real requests through these patterns:

- `/user/@{username}`
- `/{month}-{day}-{year}`
- `/files/{name}.json`
- A pattern containing both styles

Two `TestMuxRegexp` cases verify that mid-segment parameters are not treated as catch-all wildcards.

`expr/http_endpoint_test.go` extends `TestHTTPRouteValidation` with:

- A valid design that uses both patterns
- A case verifying that `/{year}-{month}-{day}` reports every missing payload attribute
- A case verifying duplicate detection within one segment

The OpenAPI tests add `PathWithMidSegmentWildcardDSL` and goldens for v2 and v3. Both emit `/user/@{username}/{month}-{day}-{year}` with all four parameters declared as `in: path`.

`TestPaths` covers path constructor generation in `service_data.go`. An incorrect URL at this point would reach generated clients. The generated code is:

```go
fmt.Sprintf("/user/@%v/%v-%v-%v", username, month, day, year)
```

`go test ./...` passes across the repository, and `golangci-lint run` is clean. No existing goldens changed, so paths that already worked are unaffected.


raphael:
Thanks for picking this back up and adding the tests! I found one small OpenAPI v2 edge case: an API base path like `/users/@{username}` gets repeated in the operation path because the replacement adds an extra `}`. I'll add a small fix and regression tests for that here. 👍

PR #3998, opened 2026-09-22
Fix HTTP CLI imports for MapParams

egandro:
HTTP commands for a payload field mapped with `MapParams`, such as `Attribute("foo", MapOf(String, String))` and `MapParams("foo")`, could not compile. The generated payload builder used `json.Unmarshal`, `fmt`, and any required validation helpers without importing their packages.

The HTTP generator now includes the mapped query value when collecting CLI flags, so the existing conversion and validation rules select the required imports. Custom type imports continue using the existing alias-resolution rules. There are no DSL, API, or wire-format changes and no data migration; regenerate the client and command packages to pick up the fix.

The regression tests compile and execute generated commands. They cover optional and required maps, explicit empty maps, defaults, malformed JSON, Unicode length validation, inline whole-map payloads, and the resulting HTTP query string. A separate generated-client test verifies that custom type aliases can still share a preferred name with a standard-library import.

Validation: `GOWORK=off make lint` and `GOWORK=off make test` pass. The new tests reproduce the missing-import and alias-conflict failures against the original PR code.


raphael:
Thanks for tracking this down! I reproduced the missing `json` and `fmt` imports with `MapParams("foo")`.

Custom type imports are already collected elsewhere, so the new test passes without this change. Adding them here makes their aliases fixed, which can also reject valid designs when an alias conflicts with another import.

The missing piece is that `MapParams` fields are left out when Goa determines which imports the CLI needs. I’ll update this PR to include them there and add a test that compiles and exercises the generated client.

egandro:
Thank you @raphael - It took me a while to make the PR. I monkey patched the gen folder 🙈 so always glad to help.

